PR #170, opened 2024-07-29
refactor(transaction): make the only query field in declare tx public

meship-starkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/170)

### DIFF
--- a/crates/blockifier/src/transaction/transactions.rs
+++ b/crates/blockifier/src/transaction/transactions.rs
@@ -136,7 +136,7 @@ pub struct DeclareTransaction {
 }
 
 impl DeclareTransaction {
-    fn create(
+    pub fn create(
         declare_tx: starknet_api::transaction::DeclareTransaction,
         tx_hash: TransactionHash,
         class_info: ClassInfo,
@@ -286,7 +286,7 @@ pub struct DeployAccountTransaction {
     pub tx_hash: TransactionHash,
     pub contract_address: ContractAddress,
     // Indicates the presence of the only_query bit in the version.
-    pub only_query: bool,
+    only_query: bool,
 }
 
 impl DeployAccountTransaction {
@@ -386,7 +386,7 @@ pub struct InvokeTransaction {
     pub tx: starknet_api::transaction::InvokeTransaction,
     pub tx_hash: TransactionHash,
     // Indicates the presence of the only_query bit in the version.
-    pub only_query: bool,
+    only_query: bool,
 }
 
 impl InvokeTransaction {


meship-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/170)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @noaov1)

*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/170#-O2ynZ0d3E16iZpOYGHB:-O2ynZ0eBsd52unwrpy2:bkvrl3k) (no related file):*
@tdelabro

___

tdelabro (approved):

noaov1 (commented):
Doing this allow to create an incorrect declare transaction (we want to enforce calling `verify_contract_class_version`).

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/170#-:-O31g4Zz1DtrJhHwweYN:b6v5xc6)* status: all files reviewed, 1 unresolved discussion (waiting on @noaov1)

noaov1 (commented):
You can let the `create` method be public. We may need to consider the necessity of the method `new_for_query` (and maybe define `new` only for tests). 

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/170#-:-O32RY4W1xZ_9_yCEYxl:b-pp38kh)* status: all files reviewed, 1 unresolved discussion (waiting on @noaov1)

meship-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/170)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @noaov1)

___
*[`crates/blockifier/src/transaction/transactions.rs` line 136 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/170#-O32Wx4O7iJinBomkoOO:-O32Wx4O7iJinBomkoOP:b2nya23) ([raw file](https://github.com/starkware-libs/sequencer/blob/d8c8b8daee747b1bb088dbc5b128a3717bc41923/crates/blockifier/src/transaction/transactions.rs#L136)):*
> ```XML
>     pub tx_hash: TransactionHash,
>     // Indicates the presence of the only_query bit in the version.
>     only_query: bool,
> ```

Changing only query to a pub trait will allow users to create a Declare tx struct without using create.  we want to enforce calling `verify_contract_class_version` so it is problematic.

meship-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/170)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @noaov1)

___
*[`crates/blockifier/src/transaction/transactions.rs` line 136 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/170#-O32Wx4O7iJinBomkoOO:-O330Dms1KGQAsExxO8T:b-fje9vh) ([raw file](https://github.com/starkware-libs/sequencer/blob/d8c8b8daee747b1bb088dbc5b128a3717bc41923/crates/blockifier/src/transaction/transactions.rs#L136)):*
<details><summary><i>Previously, meship-starkware (Meshi Peled) wrote…</i></summary><blockquote>

Changing only query to a pub trait will allow users to create a Declare tx struct without using create.  we want to enforce calling `verify_contract_class_version` so it is problematic.
</blockquote></details>

@tdelabro WDYT? Do you have another suggestion?

tdelabro:
@meship-starkware 
https://github.com/starkware-libs/blockifier/issues/1725#issuecomment-2035502658

You then need to be coherent and make it private for the other transactions types too

meship-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/170)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @noaov1)

___
*[`crates/blockifier/src/transaction/transactions.rs` line 292 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/170#-O36asMR7_Hdwz7yj34j:-O36asMSEfnq2bhiyMSg:b4tco3m) ([raw file](https://github.com/starkware-libs/sequencer/blob/d8c8b8daee747b1bb088dbc5b128a3717bc41923/crates/blockifier/src/transaction/transactions.rs#L292)):*
> ```XML
>     // Indicates the presence of the only_query bit in the version.
>     pub only_query: bool,
> }
> ```

This change means that we will have to create a Deploy Account tx using new and new for query.

tdelabro (approved):

meship-starkware (commented):
After some discussion, we decided that making it private is not a good solution because it is less readable, it is not clear why the only query is private there, and it is a breaking change. A solution that @noaov1 has suggested is making the only query public in declare but making the class info private. This is also a breaking change, but it makes more sense because the declare struct needs to be private because we need to check the class info in the constructor.  

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/170#-:-O5DCwT95vnBzf-5RiGO:bppd7cb)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @noaov1)

tdelabro (commented):
> making it private is not a good solution because it is less readable

What about adding a getter? `tx.is_query()` is not less readable that`tx.only_query`

> making the only query public in declare but making the class info private

@noaov1 has a good point that if we force users to use our `create` method in order to check for the `class_info` data, maybe we shouldn't allow users to change it's content afterward.
Note that the create method don't just use the `class_info`
```rust
    fn create(
        declare_tx: starknet_api::transaction::DeclareTransaction,
        tx_hash: TransactionHash,
        class_info: ClassInfo,
        only_query: bool,
    ) -> TransactionExecutionResult<Self> {
        let declare_version = declare_tx.version();
        verify_contract_class_version(&class_info.contract_class(), declare_version)?;
        Ok(Self { tx: declare_tx, tx_hash, class_info, only_query })
    }
```
It uses data from both `declare_tx` and `class_info` to call `verify_contract_class_version`, so those two fields should be private and only accessible through getters

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/170#-:-O2yqlIM3SXl_K97FDyU:b-7boco1)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @noaov1)

meship-starkware (commented):
It will be solved in a future PR. 

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/170#-:-O7hXNzp71Udyki6fVkC:b8xqgcz)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @noaov1)

PR #205, opened 2024-07-30
test(mempool): impl tests for calculate gas price method

MohammadNassar1:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)

### DIFF
--- a/crates/batcher/src/fee_market.rs
+++ b/crates/batcher/src/fee_market.rs
@@ -1,5 +1,9 @@
 use std::cmp::max;
 
+#[cfg(test)]
+#[path = "fee_market_test.rs"]
+pub mod fee_market_test;
+
 //  This constant is used to calculate the base gas price for the next block according to EIP-1559
 // and serves as a sensitivity parameter that limits the maximum rate of change of the gas price
 // between consecutive blocks.
--- a/crates/batcher/src/fee_market_test.rs
+++ b/crates/batcher/src/fee_market_test.rs
@@ -1,0 +1,62 @@
+use crate::fee_market::{
+    calculate_next_base_gas_price,
+    GAS_PRICE_MAX_CHANGE_DENOMINATOR,
+    MAX_BLOCK_SIZE,
+    MIN_GAS_PRICE,
+};
+
+#[test]
+fn test_price_calculation_snapshot() {
+    // Setup: using realistic arbitrary values.
+    const INIT_PRICE: u64 = 1_000_000;
+    const GAS_TARGET: u64 = MAX_BLOCK_SIZE / 2;
+    const HIGH_CONGESTION_GAS_USED: u64 = MAX_BLOCK_SIZE * 3 / 4;
+    const LOW_CONGESTION_GAS_USED: u64 = MAX_BLOCK_SIZE / 4;
+    const STABLE_CONGESTION_GAS_USED: u64 = GAS_TARGET;
+
+    // Fixed expected output values.
+    let increased_price = 1000000 + 10416; // 1000000 + (1000000 * 1 / 4 * MAX_BLOCK_SIZE) / (0.5 * MAX_BLOCK_SIZE * 48);
+    let decreased_price = 1000000 - 10416; // 1000000 - (1000000 * 1 / 4 * MAX_BLOCK_SIZE) / (0.5 * MAX_BLOCK_SIZE * 48);
+
+    // Assert.
+    assert_eq!(
+        calculate_next_base_gas_price(INIT_PRICE, HIGH_CONGESTION_GAS_USED, GAS_TARGET),
+        increased_price
+    );
+    assert_eq!(
+        calculate_next_base_gas_price(INIT_PRICE, LOW_CONGESTION_GAS_USED, GAS_TARGET),
+        decreased_price
+    );
+    assert_eq!(
+        calculate_next_base_gas_price(INIT_PRICE, STABLE_CONGESTION_GAS_USED, GAS_TARGET),
+        INIT_PRICE
+    );
+}
+
+#[test]
+// This test ensures that the gas price calculation does not overflow with extreme values,
+fn test_gas_price_with_extreme_values() {
+    let price = MIN_GAS_PRICE;
+    let gas_target = MAX_BLOCK_SIZE / 2;
+    let gas_used = 0;
+    assert_eq!(calculate_next_base_gas_price(price, gas_used, gas_target), MIN_GAS_PRICE);
+
+    let price = MIN_GAS_PRICE;
+    let gas_target = MAX_BLOCK_SIZE / 2;
+    let gas_used = MAX_BLOCK_SIZE;
+    assert!(calculate_next_base_gas_price(price, gas_used, gas_target) > MIN_GAS_PRICE);
+
+    let price = u64::MAX;
+    let gas_target = MAX_BLOCK_SIZE / 2;
+    let gas_used = 0;
+    calculate_next_base_gas_price(price, gas_used, gas_target); // Should not panic.
+
+    // To avoid overflow when updating the price, the value is set below a certain threshold so that
+    // the new price does not exceed u64::MAX.
+    let max_u128 = u128::from(u64::MAX);
+    let price_u128 =
+        max_u128 * GAS_PRICE_MAX_CHANGE_DENOMINATOR / (GAS_PRICE_MAX_CHANGE_DENOMINATOR + 1);
+    let gas_target = MAX_BLOCK_SIZE / 2;
+    let gas_used = MAX_BLOCK_SIZE;
+    calculate_next_base_gas_price(u64::try_from(price_u128).unwrap(), gas_used, gas_target); // Should not panic.
+}


elintul (dismissed):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: all files reviewed, 1 unresolved discussion (waiting on @ayeletstarkware, @giladchase, and @MohammadNassar1)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 49 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O3bEKPS1sK-oSJwy54G:-O3bEKPS1sK-oSJwy54H:b-4d9j6v) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L49)):*
> ```XML
>         format!("{}", error.unwrap_err())
>     );
> }
> ```

Don't we want to check change rate?

giladchase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: all files reviewed, 4 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @MohammadNassar1)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 49 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O3bEKPS1sK-oSJwy54G:-O44HzxXA8brmhUi1A_n:bn0afat) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L49)):*

Don't we want to check change rate?
</blockquote></details>

+1, at the very least a test that checks that a drastic change in gas usage doesn't change the block size drastically, but preferably something more exact and not flaky.

It might help to make the sensitivity parameter`BASE_FEE_MAX_CHANGE_DENOMINATOR` configurable some how for testing.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 56 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O44Iw4n2pHG_6CxSDmW:-O44Iw4n2pHG_6CxSDmX:b-n3e8xm) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L56)):*
> ```XML
>     let mut prev_price = 100;
>     let current_gas_target = 100;
>     let gas_usages = [150, 160, 170, 180, 190, 200];
> ```

5 is too much, 2 are sufficient I think



_Code quote:_
```XML
    let gas_usages = [150, 160, 170, 180, 190, 200];
```
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 88 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O44HqKI-4A7aOGTgtp0:-O44HqKI-4A7aOGTgtp1:b45h) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L88)):*
> ```XML
>     for &gas_used in &gas_usages {
>         current_price = BlockInfo::calculate_gas_price(current_price, gas_used, current_gas_target);
>         assert!(current_price == prev_price);
> ```





_Suggestion:_
```XML
       assert_eq!(current_price == prev_price);
```
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 91 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O44JKjY9brbJENxlPgI:-O44JKjY9brbJENxlPgJ:bo3y3nw) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L91)):*
> ```XML
>         prev_price = current_price;
>     }
> }
> ```

Can gas usage be 0? i'm assuming the price can never be?
Any other edge cases to test for?

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 56 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O44Iw4n2pHG_6CxSDmW:-O57cj-Y1HDafSr2C9-S:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L56)):*

5 is too much, 2 are sufficient I think
</blockquote></details>

Done.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 88 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O44HqKI-4A7aOGTgtp0:-O57cmwTCaog0hoPfDx5:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L88)):*
> ```XML
>     for &gas_used in &gas_usages {
>         current_price = BlockInfo::calculate_gas_price(current_price, gas_used, current_gas_target);
>         assert!(current_price == prev_price);
> ```

Done.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 91 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O44JKjY9brbJENxlPgI:-O57cpzyA-PrT37sqc1j:b-1xditk) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L91)):*

Can gas usage be 0? i'm assuming the price can never be?
Any other edge cases to test for?
</blockquote></details>

I've added such a test.

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ayeletstarkware and @elintul)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 88 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O44HqKI-4A7aOGTgtp0:-O57ilQDCyzVm2RxkzZf:b76dtdl) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L88)):*

Done.
</blockquote></details>

plz use pretty_assertions for assert_eq (in general every time you add assert_eq check if pretty_assertion is used, if it isn't add it)
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 91 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O44JKjY9brbJENxlPgI:-O57izARDusgfEb0iGr3:b-f435lv) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L91)):*

I've added such a test.
</blockquote></details>

link plz

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 91 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O44JKjY9brbJENxlPgI:-O57pqD5BowoE3lKX3ni:b-onv42o) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L91)):*

link plz
</blockquote></details>

The below test:
`test_gas_price_with_extreme_values`

ayeletstarkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @elintul, @giladchase, and @MohammadNassar1)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O58aiDz1taUoP2G0TK2):*
impl



_Code quote:_
```
 impll
```
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 56 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O58dUvL4tHgO4LB5wwk:-O58dUvL4tHgO4LB5wwl:b4xypz0) ([raw file](https://github.com/starkware-libs/sequencer/blob/6cfb9b7683ad1369ac1868cf2da753ce50830d5e/crates/blockifier/src/blockifier/block_test.rs#L56)):*
> ```XML
>     let mut prev_price = 100;
>     let current_gas_target = 100;
>     let gas_usages = [150, 160];
> ```

why did you choose these numbers?



_Code quote:_
```XML
[150, 160];
```
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 61 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O58dYxfDjVb70Umd0aB:-O58dYxfDjVb70Umd0aC:bebrecp) ([raw file](https://github.com/starkware-libs/sequencer/blob/6cfb9b7683ad1369ac1868cf2da753ce50830d5e/crates/blockifier/src/blockifier/block_test.rs#L61)):*
> ```XML
>         current_price =
>             BlockInfo::calculate_next_base_gas_price(current_price, gas_used, current_gas_target);
>         assert!(current_price > prev_price);
> ```

Should we only check for price growth, or do we need to verify the exact number?

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O58aiDz1taUoP2G0TK2):*
<details><summary><i>Previously, ayeletstarkware (Ayelet Zilber) wrote…</i></summary><blockquote>

impl
</blockquote></details>

Done
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 49 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O3bEKPS1sK-oSJwy54G:-O591_j6-13qnxPJQnRe:b-11ke6j) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L49)):*

+1, at the very least a test that checks that a drastic change in gas usage doesn't change the block size drastically, but preferably something more exact and not flaky.

It might help to make the sensitivity parameter`BASE_FEE_MAX_CHANGE_DENOMINATOR` configurable some how for testing.
</blockquote></details>

I've added a bound check.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 88 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O44HqKI-4A7aOGTgtp0:-O592eBE-NO5llydVIlI:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L88)):*

plz use pretty_assertions for assert_eq (in general every time you add assert_eq check if pretty_assertion is used, if it isn't add it)
</blockquote></details>

Done
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 56 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O58dUvL4tHgO4LB5wwk:-O591j2fD0NZOwuIS1rV:b-bt0qe9) ([raw file](https://github.com/starkware-libs/sequencer/blob/6cfb9b7683ad1369ac1868cf2da753ce50830d5e/crates/blockifier/src/blockifier/block_test.rs#L56)):*
<details><summary><i>Previously, ayeletstarkware (Ayelet Zilber) wrote…</i></summary><blockquote>

why did you choose these numbers?
</blockquote></details>

Good question!
I simulated a traffic congestion scenario by setting the initial gas target at 100, and then ran two blocks. 
The congestion continued to increase, from 100 to 150 and then to 160. I tracked the increasing gas prices throughout the simulation.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 61 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O58dYxfDjVb70Umd0aB:-O592JbR3PG828GP2sb3:b-gyei5c) ([raw file](https://github.com/starkware-libs/sequencer/blob/6cfb9b7683ad1369ac1868cf2da753ce50830d5e/crates/blockifier/src/blockifier/block_test.rs#L61)):*
<details><summary><i>Previously, ayeletstarkware (Ayelet Zilber) wrote…</i></summary><blockquote>

Should we only check for price growth, or do we need to verify the exact number?
</blockquote></details>

I added a bound check that the price is bounded with + - 15%.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 125 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O593HrvEDVL3-DVAo_R:-O593HrvEDVL3-DVAo_S:b-o9hbr5) ([raw file](https://github.com/starkware-libs/sequencer/blob/642368cdc08bba2508019126eecfac3bc681c5f5/crates/blockifier/src/blockifier/block_test.rs#L125)):*
> ```XML
>     let price = u64::MAX;
>     let gas_target = 1;
>     BlockInfo::calculate_next_base_gas_price(price, u64::MAX, gas_target);
> ```

Currently, this check fails.
I'm on it.



_Code quote:_
```XML
 BlockInfo::calculate_next_base_gas_price(price, u64::MAX, gas_target);
```

giladchase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 11 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @MohammadNassar1)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 91 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O44JKjY9brbJENxlPgI:-O57rINCCUbvI5F67X4p:b6g5t67) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L91)):*

The below test:
`test_gas_price_with_extreme_values`
</blockquote></details>

Oh right, sorry
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 56 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O58dUvL4tHgO4LB5wwk:-O5CFbuU3wv4X2qoyZnn:b63mzi) ([raw file](https://github.com/starkware-libs/sequencer/blob/6cfb9b7683ad1369ac1868cf2da753ce50830d5e/crates/blockifier/src/blockifier/block_test.rs#L56)):*

Good question!
I simulated a traffic congestion scenario by setting the initial gas target at 100, and then ran two blocks. 
The congestion continued to increase, from 100 to 150 and then to 160. I tracked the increasing gas prices throughout the simulation.
</blockquote></details>

Hmm, is this a coincidence that the price and gas_usage are "close" in value? that is, in production are they commonly at the same ballpark, or is gas_usage typically much larger than price?
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 52 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CEBDpDSdhK1aqhcad:-O5CEBDpDSdhK1aqhcae:bv8fvn) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L52)):*
> ```XML
> }
> 
> fn check_price_change(prev_price: u64, new_price: u64, is_increase: bool) -> bool {
> ```

Diff is important, we are not assuming it increases, we are testing if it increases



_Suggestion:_
```XML
should_increase: bool
```
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 65 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CBkAY0Fcy5SsHerzS:-O5CBkAY0Fcy5SsHerzT:bvwyks4) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L65)):*
> ```XML
>         let min_allowed_price = prev_price_f64 * 0.85;
>         new_price_f64 < prev_price_f64 && new_price_f64 >= min_allowed_price
>     }
> ```

If we're already using float, can we test the ratio new/prev?

Style nit (not a must): consider doing the comparison like X < Y && Y <= Z, it's a bit easier to read than Y > X && Y <= Z (more similar to how you'd write this in math notation)



_Code quote:_
```XML
    let prev_price_u32: u32 = prev_price.try_into().unwrap();
    let prev_price_f64 = f64::from(prev_price_u32);
    let new_price_u32: u32 = new_price.try_into().unwrap();
    let new_price_f64 = f64::from(new_price_u32);

    // Bound the price change to 15% increase or 15% decrease.
    if is_increase {
        let max_allowed_price = prev_price_f64 * 1.15;
        new_price_f64 > prev_price_f64 && new_price_f64 <= max_allowed_price
    } else {
        let min_allowed_price = prev_price_f64 * 0.85;
        new_price_f64 < prev_price_f64 && new_price_f64 >= min_allowed_price
    }
```
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 75 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CDv1EBsJqTzIhGkl0:-O5CDv1EBsJqTzIhGkl1:bwji30l) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L75)):*
> ```XML
>     let gas_usages = [150, 160];
> 
>     for &gas_used in &gas_usages {
> ```

Are & necessary? all vals are Copy here



_Code quote:_
```XML
    for &gas_used in &gas_usages {
```
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 78 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CE6FVEOwBNueSgWgt:-O5CE6FVEOwBNueSgWgu:bhaa06l) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L78)):*
> ```XML
>         current_price =
>             BlockInfo::calculate_next_base_gas_price(current_price, gas_used, current_gas_target);
>         assert!(current_price > prev_price);
> ```

Isn't this already covered by the check inside the checker func?



_Code quote:_
```XML
        assert!(current_price > prev_price);
```
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 79 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CEQJGA4-YoVqrcmUH:-O5CEQJH4GTxsadN1K4o:bbzts6f) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L79)):*
> ```XML
>             BlockInfo::calculate_next_base_gas_price(current_price, gas_used, current_gas_target);
>         assert!(current_price > prev_price);
>         assert!(check_price_change(prev_price, current_price, true));
> ```

assert is boilerplate, move inside check_price_change.

Remember that unit tests should be as succinct as possible, _ruthlessly_ optimize them for char-size.



_Suggestion:_
```XML
    check_price_change(prev_price, current_price, true);
```
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 109 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CEs7T6kNZE61Z0Js2:-O5CEs7T6kNZE61Z0Js3:b5g88jj) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L109)):*
> ```XML
>     for &gas_used in &gas_usages {
>         current_price =
>             BlockInfo::calculate_next_base_gas_price(current_price, gas_used, current_gas_target);
> ```

boilerplate, push this into `check_price_change`
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 116 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5C8LeuBAiiCDb3MNn5:-O5C8LeuBAiiCDb3MNn6:b-vlpsqf) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L116)):*
> ```XML
> 
> #[test]
> fn test_gas_price_with_extreme_values() {
> ```

I think this test is missing an assertion



_Code quote:_
```XML
fn test_gas_price_with_extreme_values() {
```

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 11 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 56 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O58dUvL4tHgO4LB5wwk:-O5CsGlb8kf-fSM3Sx3b:b-zgsm2y) ([raw file](https://github.com/starkware-libs/sequencer/blob/6cfb9b7683ad1369ac1868cf2da753ce50830d5e/crates/blockifier/src/blockifier/block_test.rs#L56)):*

Hmm, is this a coincidence that the price and gas_usage are "close" in value? that is, in production are they commonly at the same ballpark, or is gas_usage typically much larger than price?
</blockquote></details>

Today I'll ask production for these values.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 52 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CEBDpDSdhK1aqhcad:-O5CpWHE-uCZ28rAPr77:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L52)):*

Diff is important, we are not assuming it increases, we are testing if it increases
</blockquote></details>

Done.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 65 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CBkAY0Fcy5SsHerzS:-O5CtCUk77viTHF5yDBH:bywn5bh) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L65)):*

If we're already using float, can we test the ratio new/prev?

Style nit (not a must): consider doing the comparison like X < Y && Y <= Z, it's a bit easier to read than Y > X && Y <= Z (more similar to how you'd write this in math notation)
</blockquote></details>

The style was updated.
What do you mean here:
"we test the ratio new/prev?"
Do you mean to calculate the exact price? Note that the tested function doesn't use float and therefore doesn't provide an exact calculation of the price.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 75 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CDv1EBsJqTzIhGkl0:-O5CqFTBCjdRBMIlhL7f:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L75)):*

Are & necessary? all vals are Copy here
</blockquote></details>

Done
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 78 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CE6FVEOwBNueSgWgt:-O5CqkFIAoCHgFIhowRp:binbgdj) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L78)):*

Isn't this already covered by the check inside the checker func?
</blockquote></details>

Correct, removed.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 79 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CEQJGA4-YoVqrcmUH:-O5Cr4ZxEUQFPjRAnY4J:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L79)):*

assert is boilerplate, move inside check_price_change.

Remember that unit tests should be as succinct as possible, _ruthlessly_ optimize them for char-size.
</blockquote></details>

Done.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 109 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CEs7T6kNZE61Z0Js2:-O5Ct-7n196PuLT0PeRx:bktf6c4) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L109)):*

boilerplate, push this into `check_price_change`
</blockquote></details>

Not sure I get your suggestion.
I still need to get the prev value.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 116 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5C8LeuBAiiCDb3MNn5:-O5Cs6kbCeCw-B570TCI:b-o4trc0) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L116)):*

I think this test is missing an assertion
</blockquote></details>

Which assertion?
This test checks that the calculation didn't panic.

giladchase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 9 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @MohammadNassar1)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 65 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CBkAY0Fcy5SsHerzS:-O5IVxdwBR_VUsYi2YbX:bxc705z) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L65)):*

The style was updated.
What do you mean here:
"we test the ratio new/prev?"
Do you mean to calculate the exact price? Note that the tested function doesn't use float and therefore doesn't provide an exact calculation of the price.
</blockquote></details>

```rust
let ratio = new_price_f64/prev_price_f64
    if should_increase {
        assert!(1.0 < ratio && ratio <= 1.15, 
    } else {
        assert!(0.85 <= ratio && ratio< 1.0, 
```
will this kind of thing work?
Oferring it because we are already using new/prev price as floats here
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 116 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5C8LeuBAiiCDb3MNn5:-O5IafFk8ivxrIjFtN15:bwkwdnr) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L116)):*

Which assertion?
This test checks that the calculation didn't panic.
</blockquote></details>

Oh this is a "should not panic" test, that needs to be documented in the test name or code comment (I prefer comment).

Typically unit tests contain assertions, so unless said otherwise, not having one looks like a bug.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 54 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5IXDPX3YUeffP5J7xl:-O5IXDPX3YUeffP5J7xm:b-nnxtwi) ([raw file](https://github.com/starkware-libs/sequencer/blob/2f6fc1c6a80e302d645a954ab9ce0594884691d8/crates/blockifier/src/blockifier/block_test.rs#L54)):*
> ```XML
> fn check_price_change(prev_price: u64, new_price: u64, should_increase: bool) {
>     let prev_price_u32: u32 = prev_price.try_into().unwrap();
>     let prev_price_f64 = f64::from(prev_price_u32);
> ```

Ditto for second one.

The u32 is just a technical constraint, so it doesn't have a good enough claim for its own variable if readability is OK without it.



_Suggestion:_
```XML
    let prev_price_f64 = f64::from(u32::try_from(prev_price).unwrap());
```
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 122 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5Ib3fR-yfx3-UzMoyg:-O5Ib3fR-yfx3-UzMoyh:b-vhamh0) ([raw file](https://github.com/starkware-libs/sequencer/blob/2f6fc1c6a80e302d645a954ab9ce0594884691d8/crates/blockifier/src/blockifier/block_test.rs#L122)):*
> ```XML
>     // Test with maximum price and minimum gas target
>     let price = u64::MAX;
>     let gas_target = 1;
> ```

Code is clear enough on these, documentation is only for things that aren't documented through code (always add comments for these cases), or are hard to spot (sometimes add comments).

Otherwise, if the code is clear, you can trust the user to understand from reading the code.



_Suggestion:_
```XML
    let price = u64::MAX;
    let gas_target = u64::MAX;
    BlockInfo::calculate_next_base_gas_price(price, 0, gas_target);

    let price = u64::MAX;
    let gas_target = 1;
```

ayeletstarkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 8 unresolved discussions (waiting on @elintul, @giladchase, and @MohammadNassar1)

___
*[`-- commits` line 2 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O58aiDz1taUoP2G0TK2):*

Done
</blockquote></details>

+ in PR's header

giladchase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 8 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @MohammadNassar1)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 109 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CEs7T6kNZE61Z0Js2:-O5IgyRr5UZEOUokhSD8:b-ucci0q) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L109)):*

Not sure I get your suggestion.
I still need to get the prev value.
</blockquote></details>

Ya i put this in a confusing spot, sorry.
I meant all the tests that call check_price_change, they all need to call calculate-next_base_gas_price in exactly the same way before check_price_change is called? if that's the case then push it into check_price_change.

In this function there isn't check_price_change ATM, so it doesn't fit here.
(Unless check_price_change will be extended to take  Ordering::{Less,Greater,Equal} instead of a boolean and allow it to be used here as well, which might be worth it if there will be a couple more tests that test for no change in the fee, but maybe not now.)

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 6 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 65 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CBkAY0Fcy5SsHerzS:-O5Oi6rl4znG34PCKPBh:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L65)):*

```rust
let ratio = new_price_f64/prev_price_f64
    if should_increase {
        assert!(1.0 < ratio && ratio <= 1.15, 
    } else {
        assert!(0.85 <= ratio && ratio< 1.0, 
```
will this kind of thing work?
Oferring it because we are already using new/prev price as floats here
</blockquote></details>

Done
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 109 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CEs7T6kNZE61Z0Js2:-O5Om-Gv4vOd4Fo0yjuO:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L109)):*

Ya i put this in a confusing spot, sorry.
I meant all the tests that call check_price_change, they all need to call calculate-next_base_gas_price in exactly the same way before check_price_change is called? if that's the case then push it into check_price_change.

In this function there isn't check_price_change ATM, so it doesn't fit here.
(Unless check_price_change will be extended to take  Ordering::{Less,Greater,Equal} instead of a boolean and allow it to be used here as well, which might be worth it if there will be a couple more tests that test for no change in the fee, but maybe not now.)
</blockquote></details>

Done.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 116 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5C8LeuBAiiCDb3MNn5:-O5OjvLmDyDnke9FqDht:b-f8sf1p) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L116)):*

Oh this is a "should not panic" test, that needs to be documented in the test name or code comment (I prefer comment).

Typically unit tests contain assertions, so unless said otherwise, not having one looks like a bug.
</blockquote></details>

Added a comment.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 54 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5IXDPX3YUeffP5J7xl:-O5OeuLH6zNfWT_cdK1l:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/2f6fc1c6a80e302d645a954ab9ce0594884691d8/crates/blockifier/src/blockifier/block_test.rs#L54)):*

Ditto for second one.

The u32 is just a technical constraint, so it doesn't have a good enough claim for its own variable if readability is OK without it.
</blockquote></details>

Done.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 122 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5Ib3fR-yfx3-UzMoyg:-O5OiI767ZV7RxnaQNlj:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/2f6fc1c6a80e302d645a954ab9ce0594884691d8/crates/blockifier/src/blockifier/block_test.rs#L122)):*

Code is clear enough on these, documentation is only for things that aren't documented through code (always add comments for these cases), or are hard to spot (sometimes add comments).

Otherwise, if the code is clear, you can trust the user to understand from reading the code.
</blockquote></details>

Done.

giladchase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 6 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @MohammadNassar1)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 73 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5RISv40VSQJO6gWCKk:-O5RISv40VSQJO6gWCKl:bgfklk9) ([raw file](https://github.com/starkware-libs/sequencer/blob/b118b804eb65a4b1a98badd856a3da10b184789e/crates/blockifier/src/blockifier/block_test.rs#L73)):*
> ```XML
>         assert!((1.0..=1.0 + max_change).contains(&ratio));
>     } else {
>         assert!((1.0 - max_change..=1.0).contains(&ratio));
> ```

Is `!should_increase` also used for equality? If not, then we should remove the `=` before the 1.0



_Code quote:_
```XML
    } else {
        assert!((1.0 - max_change..=1.0).contains(&ratio));
```

giladchase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @MohammadNassar1)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 109 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CEs7T6kNZE61Z0Js2:-O5RJApC8gq7uz8mBlRA:bw6q751) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L109)):*

Done.
</blockquote></details>

Excellent, the tests are very readable

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 109 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5CEs7T6kNZE61Z0Js2:-O5RnXnT5v4p7Sro4n91:b3fd8e9) ([raw file](https://github.com/starkware-libs/sequencer/blob/6fd82d2de88b70d5cf732b4c23f75b0bdb2dddc4/crates/blockifier/src/blockifier/block_test.rs#L109)):*

Excellent, the tests are very readable
</blockquote></details>

I see you are still blocking here. Do you think there are additional changes needed?
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 73 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5RISv40VSQJO6gWCKk:-O5Rmmhx-cCAdYYCy5EI:b-shsayg) ([raw file](https://github.com/starkware-libs/sequencer/blob/b118b804eb65a4b1a98badd856a3da10b184789e/crates/blockifier/src/blockifier/block_test.rs#L73)):*

Is `!should_increase` also used for equality? If not, then we should remove the `=` before the 1.0
</blockquote></details>

It should also used for equality.
For example, with a small delta value (positive or negative), the division could result in zero, leaving the new price unchanged (ratio =1).

giladchase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @MohammadNassar1)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 73 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5RISv40VSQJO6gWCKk:-O5Rqvp5Cyk7vvPYDDRH:b-r279gf) ([raw file](https://github.com/starkware-libs/sequencer/blob/b118b804eb65a4b1a98badd856a3da10b184789e/crates/blockifier/src/blockifier/block_test.rs#L73)):*

It should also used for equality.
For example, with a small delta value (positive or negative), the division could result in zero, leaving the new price unchanged (ratio =1).
</blockquote></details>

Good point.
In that case, this function is also usable in `test_stable_gas_usage`? that tests equality

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 73 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5RISv40VSQJO6gWCKk:-O5Rv9GF9buU-OcxV-a5:b-tpz3cy) ([raw file](https://github.com/starkware-libs/sequencer/blob/b118b804eb65a4b1a98badd856a3da10b184789e/crates/blockifier/src/blockifier/block_test.rs#L73)):*

Good point.
In that case, this function is also usable in `test_stable_gas_usage`? that tests equality
</blockquote></details>

Please note that in `test_stable_gas_usage`, we check that the new price is exactly equal to the old price, not just within a range. Therefore, the equality is more precise.

If I want to use this method, so I need to add additional case.

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware and @elintul)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 73 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5RISv40VSQJO6gWCKk:-O5SHWkH1jadaEtksnrF:bqevoec) ([raw file](https://github.com/starkware-libs/sequencer/blob/b118b804eb65a4b1a98badd856a3da10b184789e/crates/blockifier/src/blockifier/block_test.rs#L73)):*

Please note that in `test_stable_gas_usage`, we check that the new price is exactly equal to the old price, not just within a range. Therefore, the equality is more precise.

If I want to use this method, so I need to add additional case.
</blockquote></details>

right, let's leave it as it is then

ayeletstarkware (commented):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: all files reviewed, 2 unresolved discussions (waiting on @elintul, @giladchase, and @MohammadNassar1)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 56 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O58dUvL4tHgO4LB5wwk:-O5geFu98H0TlE2rJr8j:b-s34tn6) ([raw file](https://github.com/starkware-libs/sequencer/blob/6cfb9b7683ad1369ac1868cf2da753ce50830d5e/crates/blockifier/src/blockifier/block_test.rs#L56)):*

Today I'll ask production for these values.
</blockquote></details>

any updates?

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: all files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 56 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O58dUvL4tHgO4LB5wwk:-O5gnWjb1NpD5E1Bvkif:b-jleigj) ([raw file](https://github.com/starkware-libs/sequencer/blob/6cfb9b7683ad1369ac1868cf2da753ce50830d5e/crates/blockifier/src/blockifier/block_test.rs#L56)):*
<details><summary><i>Previously, ayeletstarkware (Ayelet Zilber) wrote…</i></summary><blockquote>

any updates?
</blockquote></details>

These values are the values from product.
For example product decided that `min price = 100000`, and here in the test I used `price=1000000 > min price`.

ayeletstarkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: all files reviewed, 1 unresolved discussion (waiting on @elintul and @giladchase)

ayeletstarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: all files reviewed, 3 unresolved discussions (waiting on @elintul, @giladchase, and @MohammadNassar1)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 58 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5hH9rYEDbp2cyKFCzo:-O5hH9rYEDbp2cyKFCzp:be2x1hg) ([raw file](https://github.com/starkware-libs/sequencer/blob/2d117b777e78bc4823281021080b74c297c701fe/crates/blockifier/src/blockifier/block_test.rs#L58)):*
> ```XML
> }
> 
> fn compute_and_check_price_range(
> ```

is this the best place to locate this function? perhaps above all tests?
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 94 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5hHc894xEFH_hH7wRg:-O5hHc894xEFH_hH7wRh:b-msso1z) ([raw file](https://github.com/starkware-libs/sequencer/blob/2d117b777e78bc4823281021080b74c297c701fe/crates/blockifier/src/blockifier/block_test.rs#L94)):*
> ```XML
>     let mut prev_price = 1000000;
>     let current_gas_target = MAX_BLOCK_SIZE / 2;
>     let gas_usages = [MAX_BLOCK_SIZE * 3 / 8, MAX_BLOCK_SIZE / 3];
> ```

can you explain these calculations?



_Code quote:_
```XML
[MAX_BLOCK_SIZE * 3 / 8, MAX_BLOCK_SIZE / 3
```

ayeletstarkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: all files reviewed, 4 unresolved discussions (waiting on @elintul, @giladchase, and @MohammadNassar1)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 83 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5hJ3GJ1qdPisQw0-yo:-O5hJ3GJ1qdPisQw0-yp:b-msso1z) ([raw file](https://github.com/starkware-libs/sequencer/blob/2d117b777e78bc4823281021080b74c297c701fe/crates/blockifier/src/blockifier/block_test.rs#L83)):*
> ```XML
>     let mut prev_price = 1000000;
>     let current_gas_target = MAX_BLOCK_SIZE / 2;
>     let gas_usages = [MAX_BLOCK_SIZE * 2 / 3, MAX_BLOCK_SIZE * 3 / 4];
> ```

can you explain these calculations?

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 0 of 1 files reviewed, 4 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 58 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5hH9rYEDbp2cyKFCzo:-O5hL2yr2P1DVugQPO9e:b-1ybkor) ([raw file](https://github.com/starkware-libs/sequencer/blob/2d117b777e78bc4823281021080b74c297c701fe/crates/blockifier/src/blockifier/block_test.rs#L58)):*
<details><summary><i>Previously, ayeletstarkware (Ayelet Zilber) wrote…</i></summary><blockquote>

is this the best place to locate this function? perhaps above all tests?
</blockquote></details>

Good point!
I'll move it once I move this test to be under the batcher crate.
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 83 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5hJ3GJ1qdPisQw0-yo:-O5hLNYp-gY22mBDiFrs:bmkfpm1) ([raw file](https://github.com/starkware-libs/sequencer/blob/2d117b777e78bc4823281021080b74c297c701fe/crates/blockifier/src/blockifier/block_test.rs#L83)):*
<details><summary><i>Previously, ayeletstarkware (Ayelet Zilber) wrote…</i></summary><blockquote>

can you explain these calculations?
</blockquote></details>

Simply I wanted to test a scenario where there is congestion in the network, so the gas used is above the target.
I've added a comment for that!
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 94 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5hHc894xEFH_hH7wRg:-O5hLbwy7UQunfjnI1bN:b-gp9u4r) ([raw file](https://github.com/starkware-libs/sequencer/blob/2d117b777e78bc4823281021080b74c297c701fe/crates/blockifier/src/blockifier/block_test.rs#L94)):*
<details><summary><i>Previously, ayeletstarkware (Ayelet Zilber) wrote…</i></summary><blockquote>

can you explain these calculations?
</blockquote></details>

The same like above :)

ayeletstarkware (approved):
> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: all files reviewed, 3 unresolved discussions (waiting on @elintul, @giladchase, and @MohammadNassar1)

___
*[`crates/blockifier/src/blockifier/block_test.rs` line 69 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5hPk1r5ajCpeJxl59i:-O5hPk1r5ajCpeJxl59j:b-9v19ah) ([raw file](https://github.com/starkware-libs/sequencer/blob/cf85d3ed66b6a4dd4255bd65db65559c705b51ba/crates/blockifier/src/blockifier/block_test.rs#L69)):*
> ```XML
>     let new_price_f64 = f64::from(u32::try_from(new_price).unwrap());
> 
>     let ratio = new_price_f64 / prev_price_f64;
> ```

price_change_ratio



_Code quote:_
```XML
ratio
```

ayeletstarkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: all files reviewed, 2 unresolved discussions (waiting on @elintul, @giladchase, and @MohammadNassar1)

giladchase (approved):
:lgtm:

> Reviewed 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205#-:-O5sjEWE72sy79XYpZXH:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on @elintul and @MohammadNassar1)

___
*[`crates/batcher/src/fee_market_test.rs` line 17 at r11](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5seTrz5ufzEWoJtrHL:-O5seTrz5ufzEWoJtrHM:bd23tcz) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f3e6d3557339740260f664cd29ff7cafcb1a0fe/crates/batcher/src/fee_market_test.rs#L17)):*
> ```XML
> 
>     let increased_price = 1000000 + 10416; // 1000000 + (1000000 * 1 / 4 * MAX_BLOCK_SIZE) / (0.5 * MAX_BLOCK_SIZE * 48);
>     let decreased_price = 1000000 - 10416; // 1000000 - (1000000 * 1 / 4 * MAX_BLOCK_SIZE) / (0.5 * MAX_BLOCK_SIZE * 48);
> ```

Great!
Before the `INIT` add something like "realistic arbitrary values for these arguments", and change the test name to be more direct about what is being tested: something like price_calculation_snapshot_test, which is the general testing term for a test that just assert a recalculated "complex" response.

I want the test name and comment before the setup part to discourage readers from reading through this test (sorry :P), so they understand we're only testing some arbitrary fixed value



_Suggestion:_
```XML
    const INIT_PRICE: u64 = 1_000_000;
    const GAS_TARGET: u64 = MAX_BLOCK_SIZE / 2;
    const HIGH_CONGESTION_GAS_USED: u64 = MAX_BLOCK_SIZE * 3 / 4;
    const LOW_CONGESTION_GAS_USED: u64 = MAX_BLOCK_SIZE / 4;
    const STABLE_CONGESTION_GAS_USED: u64 = GAS_TARGET;

    let increased_price = 1000000 + 10416; // 1000000 + (1000000 * 1 / 4 * MAX_BLOCK_SIZE) / (0.5 * MAX_BLOCK_SIZE * 48);
    let decreased_price = 1000000 - 10416; // 1000000 - (1000000 * 1 / 4 * MAX_BLOCK_SIZE) / (0.5 * MAX_BLOCK_SIZE * 48);   
```
___
*[`crates/batcher/src/fee_market_test.rs` line 59 at r11](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5se1pA3am8UFKeDmMM:-O5se1pA3am8UFKeDmMN:b-fqtsps) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f3e6d3557339740260f664cd29ff7cafcb1a0fe/crates/batcher/src/fee_market_test.rs#L59)):*
> ```XML
>     let gas_target = MAX_BLOCK_SIZE / 2;
>     let gas_used = MAX_BLOCK_SIZE;
>     calculate_next_base_gas_price(u64::try_from(price_u128).unwrap(), gas_used, gas_target);
> ```

add `// should not panic` near the `calculate_next_base_gas_price`s so it won't look like a bug



_Code quote:_
```XML
    let price = u64::MAX;
    let gas_target = MAX_BLOCK_SIZE / 2;
    let gas_used = 0;
    calculate_next_base_gas_price(price, gas_used, gas_target);

    // To avoid overflow when updating the price, the value is set below a certain threshold so that
    // the new price does not exceed u64::MAX.
    let max_u128 = u128::from(u64::MAX);
    let price_u128 =
        max_u128 * GAS_PRICE_MAX_CHANGE_DENOMINATOR / (GAS_PRICE_MAX_CHANGE_DENOMINATOR + 1);
    let gas_target = MAX_BLOCK_SIZE / 2;
    let gas_used = MAX_BLOCK_SIZE;
    calculate_next_base_gas_price(u64::try_from(price_u128).unwrap(), gas_used, gas_target);
```

MohammadNassar1 (commented):
> Dismissed @elintul from a discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 1 of 3 files reviewed, all discussions resolved (waiting on @elintul and @giladchase)

___
*[`crates/batcher/src/fee_market_test.rs` line 17 at r11](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5seTrz5ufzEWoJtrHL:-O6FlZVJ53MsuP1B2GbC:b-iy0omr) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f3e6d3557339740260f664cd29ff7cafcb1a0fe/crates/batcher/src/fee_market_test.rs#L17)):*

Great!
Before the `INIT` add something like "realistic arbitrary values for these arguments", and change the test name to be more direct about what is being tested: something like price_calculation_snapshot_test, which is the general testing term for a test that just assert a recalculated "complex" response.

I want the test name and comment before the setup part to discourage readers from reading through this test (sorry :P), so they understand we're only testing some arbitrary fixed value
</blockquote></details>

Done

test name:
`test_price_calculation_snapshot`.
___
*[`crates/batcher/src/fee_market_test.rs` line 59 at r11](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5se1pA3am8UFKeDmMM:-O6FlgLd86nYNX9lypIK:bt9q6kj) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f3e6d3557339740260f664cd29ff7cafcb1a0fe/crates/batcher/src/fee_market_test.rs#L59)):*

add `// should not panic` near the `calculate_next_base_gas_price`s so it won't look like a bug
</blockquote></details>

Added.
Please note that I have added it to the test documentation - do you want me to delete it?
___
*[`crates/blockifier/src/blockifier/block_test.rs` line 49 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O3bEKPS1sK-oSJwy54G:-O6FfptAA15GUVSqWBV2:b-9wksgs) ([raw file](https://github.com/starkware-libs/sequencer/blob/7486ecc9f2002a850eaa20ab5c5a864ee3ec297e/crates/blockifier/src/blockifier/block_test.rs#L49)):*

I've added a bound check.
</blockquote></details>

Agree with you both.
After a discussion with Lior and then with you. We decided to check the actual new price, a fixed number.

giladchase (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 1 of 3 files reviewed, all discussions resolved (waiting on @elintul)

___
*[`crates/batcher/src/fee_market_test.rs` line 59 at r11](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5se1pA3am8UFKeDmMM:-O6FtBx90lHX7WwIgdCO:b-53xoja) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f3e6d3557339740260f664cd29ff7cafcb1a0fe/crates/batcher/src/fee_market_test.rs#L59)):*

Added.
Please note that I have added it to the test documentation - do you want me to delete it?
</blockquote></details>

Yep delete, it were OK in the test level if there were no asserts anywhere in the test, but since the first two _do_ assert it's better to comment inline to make it not look like a bug

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: 2 of 3 files reviewed, all discussions resolved (waiting on @elintul and @giladchase)

___
*[`crates/batcher/src/fee_market_test.rs` line 59 at r11](https://reviewable.io/reviews/starkware-libs/sequencer/205#-O5se1pA3am8UFKeDmMM:-O6FufW66vdXqh4l8Wz-:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f3e6d3557339740260f664cd29ff7cafcb1a0fe/crates/batcher/src/fee_market_test.rs#L59)):*

Yep delete, it were OK in the test level if there were no asserts anywhere in the test, but since the first two _do_ assert it's better to comment inline to make it not look like a bug
</blockquote></details>

Done

giladchase (approved):
> Reviewed 1 of 1 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/205)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @elintul)

PR #222, opened 2024-07-31
style: avoid repetion in Transaction variant names

tdelabro:
fixes: https://github.com/starkware-libs/blockifier/issues/1706

This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/222)

### DIFF
--- a/crates/blockifier/src/blockifier/stateful_validator.rs
+++ b/crates/blockifier/src/blockifier/stateful_validator.rs
@@ -84,7 +84,7 @@ impl<S: StateReader> StatefulValidator<S> {
     }
 
     fn execute(&mut self, tx: AccountTransaction) -> StatefulValidatorResult<()> {
-        self.tx_executor.execute(&Transaction::AccountTransaction(tx))?;
+        self.tx_executor.execute(&Transaction::Account(tx))?;
         Ok(())
     }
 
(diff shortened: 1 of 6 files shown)


meship-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/222)* status: all files reviewed, 1 unresolved discussion (waiting on @tdelabro)

___
*[`crates/blockifier/src/transaction/transaction_execution.rs` line 66 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/222#-O4dMTf-6cAVF3Q5pQro:-O4dMTf-6cAVF3Q5pQrp:bz1rcx2) ([raw file](https://github.com/starkware-libs/sequencer/blob/0a4d9c24e9511f70671dfec66c9d027edbeebbbc/crates/blockifier/src/transaction/transaction_execution.rs#L66)):*
> ```XML
>                     false => DeclareTransaction::new(declare, tx_hash, non_optional_class_info),
>                 };
>                 Ok(Self::AccountTransaction(AccountTransaction::Declare(declare_tx?)))
> ```

I  think that Self::Account might be a bit confusing here. Account and AccountTransaction are different things. I fear that a reader unfamiliar with the code might think we are creating an account that will run the transaction.    



_Code quote:_
```XML
(Self::AccountTransaction(AccountTransaction::Declare(declare_tx?))
```

tdelabro:
Due to the way Enum syntax works there won't be any occurrence of `Account` just by itself, it will always be `Account::Transaction` which is self-explanatory and doesn't remove information in comparison to `Account::AccountTransaction`.

When using `Self::Transaction` you are always in a context where you are fully aware of what `Self` is.

So I don't think it is misleading at all. Especially because it is coherent with the rest of the naming convention used

meship-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/222)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @tdelabro)

PR #482, opened 2024-08-15
feat(mempool): add pending queue to tx queue

MohammadNassar1:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)

### DIFF
--- a/crates/mempool/src/transaction_queue.rs
+++ b/crates/mempool/src/transaction_queue.rs
@@ -5,12 +5,16 @@ use starknet_api::core::{ContractAddress, Nonce};
 
 use crate::mempool::TransactionReference;
 
+// A queue holding the transaction that with nonces that match account nonces.
 // Note: the derived comparison functionality considers the order guaranteed by the data structures
 // used.
 #[derive(Debug, Default, Eq, PartialEq)]
 pub struct TransactionQueue {
-    // Priority queue of transactions with associated priority.
+    gas_price_threshold: u128,
+    // Transactions with gas price above gas price threshold (sorted by tip).
     priority_queue: BTreeSet<PriorityTransaction>,
+    // Transactions with gas price below gas price threshold (sorted by price).
+    pending_queue: BTreeSet<PendingTransaction>,
     // Set of account addresses for efficient existence checks.
     address_to_tx: HashMap<ContractAddress, TransactionReference>,
 }
@@ -18,8 +22,6 @@ pub struct TransactionQueue {
 impl TransactionQueue {
     /// Adds a transaction to the mempool, ensuring unique keys.
     /// Panics: if given a duplicate tx.
-    // TODO(Mohammad): Add test for two transactions from the same address, expecting specific
-    // assert.
     pub fn insert(&mut self, tx_reference: TransactionReference) {
         assert_eq!(
             self.address_to_tx.insert(tx_reference.sender_address, tx_reference.clone()),
@@ -27,8 +29,15 @@ impl TransactionQueue {
             "Only a single transaction from the same contract class can be in the mempool at a \
              time."
         );
+
+        let new_tx_successfully_inserted =
+            if tx_reference.get_l2_gas_price() < self.gas_price_threshold {
+                self.pending_queue.insert(tx_reference.into())
+            } else {
+                self.priority_queue.insert(tx_reference.into())
+            };
         assert!(
-            self.priority_queue.insert(tx_reference.into()),
+            new_tx_successfully_inserted,
             "Keys should be unique; duplicates are checked prior."
         );
     }
@@ -57,10 +66,12 @@ impl TransactionQueue {
     /// Removes the transaction of the given account address from the queue.
     /// This is well-defined, since there is at most one transaction per address in the queue.
     pub fn remove(&mut self, address: ContractAddress) -> bool {
-        if let Some(tx) = self.address_to_tx.remove(&address) {
-            return self.priority_queue.remove(&tx.into());
-        }
-        false
+        let Some(tx_reference) = self.address_to_tx.remove(&address) else {
+            return false;
+        };
+
+        self.priority_queue.remove(&tx_reference.clone().into())
+            || self.pending_queue.remove(&tx_reference.into())
     }
 
     pub fn has_ready_txs(&self) -> bool {


elintul (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 7 unresolved discussions (waiting on @ayeletstarkware, @giladchase, and @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 15 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jo-3LDl08xdHYRXZ6:-O4jo-3LDl08xdHYRXZ7:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L15)):*
> ```XML
>     priority_queue: BTreeSet<PriorityTransaction>,
>     // queue of transaction with low gas price.
>     pending_queue: BTreeSet<PendingTransaction>,
> ```





_Suggestion:_
```XML
// A queue holding the transaction that with nonces that match account nonces. 
// Note: the derived comparison functionality considers the order guaranteed by the data structures
// used.
#[derive(Debug, Default, Eq, PartialEq)]
pub struct TransactionQueue {
    // Transactions with gas price above base price (sorted by tip).
    priority_queue: BTreeSet<PriorityTransaction>,
    // Transactions with gas price below base price (sorted by price).
    pending_queue: BTreeSet<PendingTransaction>,
```
___
*[`crates/mempool/src/transaction_queue.rs` line 16 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jnhrb42mb8Av7zzOg:-O4jnhrb42mb8Av7zzOh:b-rqdhaj) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L16)):*
> ```XML
>     // queue of transaction with low gas price.
>     pending_queue: BTreeSet<PendingTransaction>,
>     gas_price_threshold: u128,
> ```

Will the mempool need it as well?



_Code quote:_
```XML
gas_price_threshold: u128,
```
___
*[`crates/mempool/src/transaction_queue.rs` line 33 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jpD1PD2RuPvasWkVi:-O4jpD1PD2RuPvasWkVj:bonw1qn) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L33)):*
> ```XML
>              time."
>         );
>         if tx_reference.get_l2_gas_price() < Some(self.gas_price_threshold) {
> ```

This shouldn't be `Option` for `>= V3`; like `tip`.



_Code quote:_
```XML
tx_reference.get_l2_gas_price()
```
___
*[`crates/mempool/src/transaction_queue.rs` line 33 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jpHJk-AMt0vl8HbOb:-O4jpHJk-AMt0vl8HbOc:b60bke9) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L33)):*
> ```XML
>              time."
>         );
>         if tx_reference.get_l2_gas_price() < Some(self.gas_price_threshold) {
> ```

Is there a more ergonomic way that isn't `if-let`?



_Suggestion:_
```XML
        );
        
        if tx_reference.get_l2_gas_price() < Some(self.gas_price_threshold) {
```
___
*[`crates/mempool/src/transaction_queue.rs` line 35 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jp7_XBYlpgoplymYb:-O4jp7_XBYlpgoplymYc:b-dhzn4d) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L35)):*
> ```XML
>         if tx_reference.get_l2_gas_price() < Some(self.gas_price_threshold) {
>             assert!(
>                 self.pending_queue.insert(tx_reference.into()),
> ```

Define a variable.



_Code quote:_
```XML
tx_reference.into()
```
___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jpW6S1CcDzSi2QrnL:-O4jpW6S1CcDzSi2QrnM:b-hal59a) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L70)):*
> ```XML
>     /// This is well-defined, since there is at most one transaction per address in the queue.
>     pub fn remove(&mut self, address: ContractAddress) -> bool {
>         if let Some(tx) = self.address_to_tx.remove(&address) {
> ```

Throughout this file, please; `tx.clone()` makes me wonder if we're cloning full transactions...



_Suggestion:_
```XML
tx_reference
```
___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jpo9t4wLi7q0rp7A1:-O4jpo9t4wLi7q0rp7A2:b-2rk92j) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L70)):*
> ```XML
>     /// This is well-defined, since there is at most one transaction per address in the queue.
>     pub fn remove(&mut self, address: ContractAddress) -> bool {
>         if let Some(tx) = self.address_to_tx.remove(&address) {
> ```

How about changing the order, so that `false` is an early `return`?



_Code quote:_
```XML
if let Some(tx)
```

elintul (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 5 unresolved discussions (waiting on @ayeletstarkware, @giladchase, and @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 16 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jnhrb42mb8Av7zzOg:-O4k2Cr8-UckwZRkMb4X:b-q11bum) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L16)):*

Currently, I think it's only relevant for the transaction queue.
When the transaction eviction is ready, the mempool or ranking manager must know the gap price threshold.
</blockquote></details>

Why `u128`? I wonder if we should call it `base_gas_price`, or we might apply some function on it (`0.8x`)?

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 16 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jnhrb42mb8Av7zzOg:-O4k6x906GDb_uIQX1RI:b-5r6kmp) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L16)):*

Why `u128`? I wonder if we should call it `base_gas_price`, or we might apply some function on it (`0.8x`)?
</blockquote></details>

-> Why `u128`?
This is how it defined in [resource_bounds](https://gitkraken.dev/link/dnNjb2RlOi8vZWFtb2Rpby5naXRsZW5zL2xpbmsvci9mZDUzOGU3ZGVjMTBlNDUwZTIzZWU1MDAyMDA2YzYyYjdhNzA2ZGUwL2YvY3JhdGVzL3N0YXJrbmV0X2FwaS9zcmMvdHJhbnNhY3Rpb24ucnM%2FdXJsPWh0dHBzJTNBJTJGJTJGZ2l0aHViLmNvbSUyRnN0YXJrd2FyZS1saWJzJTJGc2VxdWVuY2VyLmdpdCZsaW5lcz04ODQtODg1?origin=gitlens).

the `0.8x` is for gateway, in tx queue, it must send to Batcher only txs above the gas_price.
___
*[`crates/mempool/src/transaction_queue.rs` line 33 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jpD1PD2RuPvasWkVi:-O4k7LxL2YbYlHSmgl2U:b-n9tkhd) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L33)):*

This shouldn't be `Option` for `>= V3`; like `tip`.
</blockquote></details>

changed
___
*[`crates/mempool/src/transaction_queue.rs` line 35 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jp7_XBYlpgoplymYb:-O4k7VQB0f3je4onGZTo:b-dzusb9) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L35)):*

Define a variable.
</blockquote></details>

Note that the tx_reference.into()in pending and in priority queue have different types.
So, do you suggest to define two variables?

elintul (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 3 unresolved discussions (waiting on @ayeletstarkware, @giladchase, and @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 16 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jnhrb42mb8Av7zzOg:-O4k9W7YEkb9yih_Pk5I:b-8j2o0y) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L16)):*

-> Why `u128`?
This is how it defined in [resource_bounds](https://gitkraken.dev/link/dnNjb2RlOi8vZWFtb2Rpby5naXRsZW5zL2xpbmsvci9mZDUzOGU3ZGVjMTBlNDUwZTIzZWU1MDAyMDA2YzYyYjdhNzA2ZGUwL2YvY3JhdGVzL3N0YXJrbmV0X2FwaS9zcmMvdHJhbnNhY3Rpb24ucnM%2FdXJsPWh0dHBzJTNBJTJGJTJGZ2l0aHViLmNvbSUyRnN0YXJrd2FyZS1saWJzJTJGc2VxdWVuY2VyLmdpdCZsaW5lcz04ODQtODg1?origin=gitlens).

the `0.8x` is for gateway, in tx queue, it must send to Batcher only txs above the gas_price.
</blockquote></details>

Link doesn't work.
___
*[`crates/mempool/src/transaction_queue.rs` line 35 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jp7_XBYlpgoplymYb:-O4k9RWP3gRZojPDAuak:b-9tg7zn) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L35)):*

Note that the tx_reference.into()in pending and in priority queue have different types.
So, do you suggest to define two variables?
</blockquote></details>

Right, nope.
___
*[`crates/mempool/src/transaction_queue.rs` line 36 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4k9Nwg7uR8Zi9r4KzA:-O4k9Nwg7uR8Zi9r4KzB:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/8ae7bcaf2ee44de92063717690ef29ce98631c96/crates/mempool/src/transaction_queue.rs#L36)):*
> ```XML
> 
>         if tx_reference.get_l2_gas_price() < self.gas_price_threshold {
>             // self.gas_price_threshold} {
> ```

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 16 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4jnhrb42mb8Av7zzOg:-O4kJpd83-JInleL5DQ3:btqvbi7) ([raw file](https://github.com/starkware-libs/sequencer/blob/b0df4d60837ee6380021bbd63bee7c3793009b9b/crates/mempool/src/transaction_queue.rs#L16)):*

Link doesn't work.
</blockquote></details>

https://github.com/starkware-libs/sequencer/blob/6af5e477195984520106ae706e10ed7521d746f6/crates/starknet_api/src/transaction.rs#L884
___
*[`crates/mempool/src/transaction_queue.rs` line 36 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4k9Nwg7uR8Zi9r4KzA:-O4lOj7T8q-y-K0Jb-QW:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/8ae7bcaf2ee44de92063717690ef29ce98631c96/crates/mempool/src/transaction_queue.rs#L36)):*
> ```XML
> 
>         if tx_reference.get_l2_gas_price() < self.gas_price_threshold {
>             // self.gas_price_threshold} {
> ```

Done.

elintul (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 1 unresolved discussion (waiting on @ayeletstarkware, @giladchase, and @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 34 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4n-0rqESW4cvI786Fp:-O4n-0rr6xe7BsU_xkWO:b-cwla23) ([raw file](https://github.com/starkware-libs/sequencer/blob/811a0ea69320141c7e220fcc166fe8db50b8a20d/crates/mempool/src/transaction_queue.rs#L34)):*
> ```XML
>              time."
>         );
> ```

Can the `assert` code duplication be avoided? Maybe taking the target queue first?

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 1 unresolved discussion (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 34 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4n-0rqESW4cvI786Fp:-O4nPljO1Qqtf6WBFwzT:bl6wvpl) ([raw file](https://github.com/starkware-libs/sequencer/blob/811a0ea69320141c7e220fcc166fe8db50b8a20d/crates/mempool/src/transaction_queue.rs#L34)):*

Can the `assert` code duplication be avoided? Maybe taking the target queue first?
</blockquote></details>

The queues have different types, so to take the target queue first, I should add enum for the queue types. 

to save some duplication, I used the same error message.

elintul (approved):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 1 unresolved discussion (waiting on @ayeletstarkware, @giladchase, and @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 34 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4n-0rqESW4cvI786Fp:-O4nekco0RPmmsKM5gob:b-s0ojhe) ([raw file](https://github.com/starkware-libs/sequencer/blob/811a0ea69320141c7e220fcc166fe8db50b8a20d/crates/mempool/src/transaction_queue.rs#L34)):*

The queues have different types, so to take the target queue first, I should add enum for the queue types. 

to save some duplication, I used the same error message.
</blockquote></details>

Maybe Gilad will have a suggestion?

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 1 unresolved discussion (waiting on @ayeletstarkware and @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 34 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4n-0rqESW4cvI786Fp:-O4pBKKK8dMr2uuoL6-1:bfq8fzf) ([raw file](https://github.com/starkware-libs/sequencer/blob/811a0ea69320141c7e220fcc166fe8db50b8a20d/crates/mempool/src/transaction_queue.rs#L34)):*

Maybe Gilad will have a suggestion?
</blockquote></details>

@giladchase what do you think?

elintul (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware, @giladchase, and @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 34 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4n-0rqESW4cvI786Fp:-O4tMNZr17behftKDeBs:bbsg2n9) ([raw file](https://github.com/starkware-libs/sequencer/blob/811a0ea69320141c7e220fcc166fe8db50b8a20d/crates/mempool/src/transaction_queue.rs#L34)):*

@giladchase what do you think?
</blockquote></details>

Please move to Slack.
___
*[`crates/mempool/src/transaction_queue.rs` line 75 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4tMTaI-N57R4KNsnfZ:-O4tMTaI-N57R4KNsnf_:b-qa1c1p) ([raw file](https://github.com/starkware-libs/sequencer/blob/f77733384796ac498f3e35bf9498d025fee97ed3/crates/mempool/src/transaction_queue.rs#L75)):*
> ```XML
>     }
> 
>     pub fn is_empty(&self) -> bool {
> ```

No change here in the end? Why?



_Code quote:_
```XML
is_empty
```

elintul (commented):
> Reviewed 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware, @giladchase, and @MohammadNassar1)

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 1 unresolved discussion (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 75 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4tMTaI-N57R4KNsnfZ:-O4taqZrEI0-KUUYjOSO:blkn31k) ([raw file](https://github.com/starkware-libs/sequencer/blob/f77733384796ac498f3e35bf9498d025fee97ed3/crates/mempool/src/transaction_queue.rs#L75)):*

No change here in the end? Why?
</blockquote></details>

Because I believe it should align with the `iter` method. Since we iterate over the priority queue, the `empty` method should return the state of the priority queue. 
Moreover, it's only used only [here](https://github.com/starkware-libs/sequencer/blob/f31354d3221bcf4d39ab19d5489d99cd89b45a26/crates/mempool/src/mempool.rs#L47) where it refers to the state of the priority queue.

elintul (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware, @giladchase, and @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 26 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4u8PjJCKSsrmXgnqfx:-O4u8PjJCKSsrmXgnqfy:b-fcmeux) ([raw file](https://github.com/starkware-libs/sequencer/blob/f77733384796ac498f3e35bf9498d025fee97ed3/crates/mempool/src/transaction_queue.rs#L26)):*
> ```XML
>     /// Panics: if given a duplicate tx.
>     // TODO(Mohammad): Add test for two transactions from the same address, expecting specific
>     // assert.
> ```

BTW, was this done in the end?



_Code quote:_
```XML
    // TODO(Mohammad): Add test for two transactions from the same address, expecting specific
    // assert.
```
___
*[`crates/mempool/src/transaction_queue.rs` line 75 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4tMTaI-N57R4KNsnfZ:-O4u6s3p9EwFrKIbsy5i:b151v59) ([raw file](https://github.com/starkware-libs/sequencer/blob/f77733384796ac498f3e35bf9498d025fee97ed3/crates/mempool/src/transaction_queue.rs#L75)):*

Because I believe it should align with the `iter` method. Since we iterate over the priority queue, the `empty` method should return the state of the priority queue. 
Moreover, it's only used only [here](https://github.com/starkware-libs/sequencer/blob/f31354d3221bcf4d39ab19d5489d99cd89b45a26/crates/mempool/src/mempool.rs#L47) where it refers to the state of the priority queue.
</blockquote></details>

Suddenly I'm worried the Q API isn't explicit enough; moving this discussion to Slack.
BTW, as the writer: when you're making such assumptions, please be sure to let your reviewer know by writing your thoughts in a comment.

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 26 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4u8PjJCKSsrmXgnqfx:-O4uLMsh8y2FCDJylTGs:b-ad2zaw) ([raw file](https://github.com/starkware-libs/sequencer/blob/f77733384796ac498f3e35bf9498d025fee97ed3/crates/mempool/src/transaction_queue.rs#L26)):*

BTW, was this done in the end?
</blockquote></details>

This scenario, where a transaction is added to a non-empty `tx_queue`, cannot occur.
The `insert` method is only used in the mempool, and before calling it, we ensure that the queue is empty.

Removing this TODO.
___
*[`crates/mempool/src/transaction_queue.rs` line 75 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4tMTaI-N57R4KNsnfZ:-O4uIraR0cmpvIp11qMk:b-yqixdv) ([raw file](https://github.com/starkware-libs/sequencer/blob/f77733384796ac498f3e35bf9498d025fee97ed3/crates/mempool/src/transaction_queue.rs#L75)):*

Suddenly I'm worried the Q API isn't explicit enough; moving this discussion to Slack.
BTW, as the writer: when you're making such assumptions, please be sure to let your reviewer know by writing your thoughts in a comment.
</blockquote></details>

As we decided, I'll rename these methods in a different PR.

elintul (changes requested):
> Reviewed 1 of 1 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware, @giladchase, and @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 26 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4u8PjJCKSsrmXgnqfx:-O4uQOo3CJDdVsDje8kq:b-y2020u) ([raw file](https://github.com/starkware-libs/sequencer/blob/f77733384796ac498f3e35bf9498d025fee97ed3/crates/mempool/src/transaction_queue.rs#L26)):*

This scenario, where a transaction is added to a non-empty `tx_queue`, cannot occur.
The `insert` method is only used in the mempool, and before calling it, we ensure that the queue is empty.

Removing this TODO.
</blockquote></details>

Empty, meaning: this address does not exist in the Q?
___
*[`crates/mempool/src/transaction_queue.rs` line 39 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4uQUd3D8cnNekOEjsw:-O4uQUd3D8cnNekOEjsx:b-m2f0vt) ([raw file](https://github.com/starkware-libs/sequencer/blob/153c41ac5930afbadcad2382eba36251b266b895/crates/mempool/src/transaction_queue.rs#L39)):*
> ```XML
>         };
> 
>         assert!(tx_already_existed, "Keys should be unique; duplicates are checked prior.");
> ```

You should assert that `tx_already_existed is None`. Let's call it "[optional/maybe]_already_existing_tx".



_Suggestion:_
```XML
        };
        assert!(tx_already_existed, "Keys should be unique; duplicates are checked prior.");
```

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 26 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4u8PjJCKSsrmXgnqfx:-O4uTT8LBSPdfBN_ftz4:b2nign6) ([raw file](https://github.com/starkware-libs/sequencer/blob/f77733384796ac498f3e35bf9498d025fee97ed3/crates/mempool/src/transaction_queue.rs#L26)):*

Empty, meaning: this address does not exist in the Q?
</blockquote></details>

yes
___
*[`crates/mempool/src/transaction_queue.rs` line 39 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4uQUd3D8cnNekOEjsw:-O4uTupa5Ben57lepPtP:b6bqzop) ([raw file](https://github.com/starkware-libs/sequencer/blob/153c41ac5930afbadcad2382eba36251b266b895/crates/mempool/src/transaction_queue.rs#L39)):*

You should assert that `tx_already_existed is None`. Let's call it "[optional/maybe]_already_existing_tx".
</blockquote></details>

It should return true, and that's what we check in the assert.

insert docstr:
/// - If the set did not previously contain an equal value, \`true\` is

/// returned.

/// - If the set already contained an equal value, \`false\` is returned, and

/// the entry is not updated.

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 39 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4uQUd3D8cnNekOEjsw:-O4uUtT76FjmLQbYIfzn:b-969kf5) ([raw file](https://github.com/starkware-libs/sequencer/blob/153c41ac5930afbadcad2382eba36251b266b895/crates/mempool/src/transaction_queue.rs#L39)):*

It should return true, and that's what we check in the assert.

insert docstr:
/// - If the set did not previously contain an equal value, \`true\` is

/// returned.

/// - If the set already contained an equal value, \`false\` is returned, and

/// the entry is not updated.
</blockquote></details>

renamed to `new_tx_inserted`. wdyt?

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O57qVVY4lznL2y0WSmb:-O57vNfZ0G7MowzJB_DW:bqgj1qf) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L70)):*

Who calls this function? what happens when there's a `false`?
</blockquote></details>

It's called in commit_block and add_tx.
Currently, it doesn't check the return value.

elintul (changes requested):
> Reviewed 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware, @giladchase, and @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 39 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4uQUd3D8cnNekOEjsw:-O57zBvtCuH7SYrZZZNr:bbzx8vx) ([raw file](https://github.com/starkware-libs/sequencer/blob/153c41ac5930afbadcad2382eba36251b266b895/crates/mempool/src/transaction_queue.rs#L39)):*

renamed to `new_tx_inserted`. wdyt?
</blockquote></details>

Shouldn't I'd be looking [here](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.insert)?
___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O57qVVY4lznL2y0WSmb:-O59v6BE-cZy3bnO6D25:b6zo9us) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L70)):*

It's called in commit_block and add_tx.
Currently, it doesn't check the return value.
</blockquote></details>

So we don't assume the address exists there?
Something to be mindful of when an already-used data structure changes the inside operation of its API, and hint the reviewer it's a thought path you've taken. :)

ayeletstarkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 4 unresolved discussions (waiting on @giladchase and @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 13 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O5CpMFHD1bCtcg4ISAe:-O5CpMFHD1bCtcg4ISAf:b-gewxja) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L13)):*
> ```XML
> #[derive(Debug, Default, Eq, PartialEq)]
> pub struct TransactionQueue {
>     // Transactions with gas price above base price (sorted by tip).
> ```

above gas price threshold?



_Code quote:_
```XML
above base price
```
___
*[`crates/mempool/src/transaction_queue.rs` line 17 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O5CofuI10nxNiDeeodk:-O5CofuJ2orKpK06J4qR:bongjox) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L17)):*
> ```XML
>     // Transactions with gas price below base price (sorted by price).
>     pending_queue: BTreeSet<PendingTransaction>,
>     gas_price_threshold: u128,
> ```

Should this be the first field of the struct? It might make the difference between the queues clearer.



_Code quote:_
```XML
gas_price_threshold: u128,
```

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on @ayeletstarkware, @elintul, and @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 39 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O4uQUd3D8cnNekOEjsw:-O5DC5vF9_6BIq1hcrzQ:b-m8c8vj) ([raw file](https://github.com/starkware-libs/sequencer/blob/153c41ac5930afbadcad2382eba36251b266b895/crates/mempool/src/transaction_queue.rs#L39)):*

Shouldn't I'd be looking [here](https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#method.insert)?
</blockquote></details>

We use `BTreeSet`, [here](https://doc.rust-lang.org/std/collections/struct.BTreeSet.html#method.insert).
___
*[`crates/mempool/src/transaction_queue.rs` line 17 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O5CofuI10nxNiDeeodk:-O5Cud1v2oaVokAF3Psz:b-ian6po) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L17)):*
<details><summary><i>Previously, ayeletstarkware (Ayelet Zilber) wrote…</i></summary><blockquote>

Should this be the first field of the struct? It might make the difference between the queues clearer.
</blockquote></details>

Good point, updated accordingly
___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O57qVVY4lznL2y0WSmb:-O5DCFF_3z4jpATEPybi:bok20v6) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L70)):*

So we don't assume the address exists there?
Something to be mindful of when an already-used data structure changes the inside operation of its API, and hint the reviewer it's a thought path you've taken. :)
</blockquote></details>

Let's add assert in our code.
I'll do it in a separate PR.

elintul (changes requested):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 2 unresolved discussions (waiting on @giladchase and @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O57qVVY4lznL2y0WSmb:-O5HsrEEBvOhf5IClgyS:b7ndye9) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L70)):*

Let's add assert in our code.
I'll do it in a separate PR.
</blockquote></details>

To assert what?
___
*[`crates/mempool/src/transaction_queue.rs` line 33 at r10](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O5Hshh47XG5X-MfiN2K:-O5Hshh47XG5X-MfiN2L:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/6565fa31de84f8bc02ee3a2aab4adffc7979cf0c/crates/mempool/src/transaction_queue.rs#L33)):*
> ```XML
>         );
> 
>         let new_tx_inserted = if tx_reference.get_l2_gas_price() < self.gas_price_threshold {
> ```





_Suggestion:_
```XML
new_tx_successfully_inserted
```

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 1 unresolved discussion (waiting on @elintul and @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O57qVVY4lznL2y0WSmb:-O5HwxXz42gY66D37TOM:b-6wjdol) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L70)):*

To assert what?
</blockquote></details>

That remove returns true.

elintul (approved):
:lgtm:

> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482#-:-O5I4HW02uGpbIipfdIP:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O57qVVY4lznL2y0WSmb:-O5I4Box2LWmY52iufSE:b-d6ym8j) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L70)):*

That remove returns true.
</blockquote></details>

Do we care about the result?

giladchase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 1 unresolved discussion (waiting on @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O57qVVY4lznL2y0WSmb:-O5IPRWM5W9dwKGSTgBD:bdtdk3x) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L70)):*

Do we care about the result?
</blockquote></details>

I'm not sure we do :thinking: 

Returning a result and asserting in the callsites:
As a pure DS it makes sense to return `bool` here, also this DS doesn't really need to know about our assumption that this removal can never fail in order to do its job.

Not returning a result and asserting inside the `remove`:
Documents our assumption that removals never fail in a central location, which prevents future callsites or refactorings of the callsites from violating this assertion (for example if someone calls this method in a test without asserting the result).
Less duplication of assertions in the callsites.

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O57qVVY4lznL2y0WSmb:-O5Ix2dZDCkYT_Wm9tMM:b-fo7qyh) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L70)):*

I'm not sure we do :thinking: 

Returning a result and asserting in the callsites:
As a pure DS it makes sense to return `bool` here, also this DS doesn't really need to know about our assumption that this removal can never fail in order to do its job.

Not returning a result and asserting inside the `remove`:
Documents our assumption that removals never fail in a central location, which prevents future callsites or refactorings of the callsites from violating this assertion (for example if someone calls this method in a test without asserting the result).
Less duplication of assertions in the callsites.
</blockquote></details>

I think that as a pure DS, we should return a boolean value. 
We can add an assertion in the `mempool.rs` code to ensure the consistency between the transaction pool and transaction queue. Specifically, we should check that the queue contains the transaction, and if so, the `remove` operation should return `true`.

giladchase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 1 unresolved discussion (waiting on @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O57qVVY4lznL2y0WSmb:-O5N0Eg43QwMQp0sggSY:bkz830z) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L70)):*
>Specifically, we should check that the queue contains the transaction, and if so, the `remove` operation should return `true`.

Is that the intended behavior in all callsites of this function?

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O57qVVY4lznL2y0WSmb:-O5N1JE5D2wTKnNHr699:b8wgf9e) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L70)):*

>Specifically, we should check that the queue contains the transaction, and if so, the `remove` operation should return `true`.

Is that the intended behavior in all callsites of this function?
</blockquote></details>

As mentioned earlier, we currently call this method in two places, and it should return `true`. To ensure consistency, we can add a check.

giladchase (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 1 unresolved discussion (waiting on @MohammadNassar1)

___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O57qVVY4lznL2y0WSmb:-O5N2kcX6f22K7fTTTiT:b-xi6qk7) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L70)):*

As mentioned earlier, we currently call this method in two places, and it should return `true`. To ensure consistency, we can add a check.
</blockquote></details>

You mentioned add_tx and commit_block, does the remove in commit_block always return true?

MohammadNassar1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

___
*[`crates/mempool/src/transaction_queue.rs` line 70 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/482#-O57qVVY4lznL2y0WSmb:-O5N48Xd4HAteiNfbB9V:br5uzt0) ([raw file](https://github.com/starkware-libs/sequencer/blob/2bbce4b8843e69a9e26bf305863a20f993357d5a/crates/mempool/src/transaction_queue.rs#L70)):*

You mentioned add_tx and commit_block, does the remove in commit_block always return true?
</blockquote></details>

You are correct, in commit_block it may return false.
While in add_tx it should always return true.

giladchase (approved):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/482)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @MohammadNassar1)

PR #595, opened 2024-08-26
feat(mempool): add suspended transaction pool to mempool, no implementation

ayeletstarkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/595)

### DIFF
--- a/crates/mempool/src/mempool.rs
+++ b/crates/mempool/src/mempool.rs
@@ -24,7 +24,7 @@ pub struct Mempool {
     // Transactions eligible for sequencing.
     tx_queue: TransactionQueue,
     // Transactions suspended because they are after a hole.
-    _suspended_tx_pool: SuspendedTransactionPool,
+    suspended_tx_pool: SuspendedTransactionPool,
     // Represents the current state of the mempool during block creation.
     mempool_state: HashMap<ContractAddress, AccountState>,
     // The most recent account nonces received, for all account in the pool.
@@ -79,7 +79,9 @@ impl Mempool {
         self.validate_input(&input)?;
         let MempoolInput { tx, account: Account { sender_address, state: AccountState { nonce } } } =
             input;
+        let tx_reference = TransactionReference::new(&tx);
         self.tx_pool.insert(tx)?;
+        self.insert_to_suspended_pool_if_eligible(tx_reference);
         self.align_to_account_state(sender_address, nonce);
         Ok(())
     }
@@ -181,6 +183,8 @@ impl Mempool {
 
         self.tx_pool.remove_up_to_nonce(address, nonce);
 
+        self.suspended_tx_pool.remove_up_to_nonce_and_sequential(address, nonce);
+
         // Maybe close nonce gap.
         if self.tx_queue.get_nonce(address).is_none() {
             if let Some(tx_reference) = self.tx_pool.get_by_address_and_nonce(address, nonce) {
@@ -188,6 +192,9 @@ impl Mempool {
             }
         }
     }
+
+    // TODO(Ayelet): Implement this function.
+    fn insert_to_suspended_pool_if_eligible(&mut self, _tx: TransactionReference) {}
 
     #[cfg(test)]
     pub(crate) fn tx_pool(&self) -> &TransactionPool {
--- a/crates/mempool/src/mempool_test.rs
+++ b/crates/mempool/src/mempool_test.rs
@@ -88,7 +88,7 @@ impl From<MempoolContent> for Mempool {
             tx_pool: tx_pool.unwrap_or_default(),
             tx_queue: tx_queue.unwrap_or_default(),
             // TODO: Add implementation when needed.
-            _suspended_tx_pool: Default::default(),
+            suspended_tx_pool: Default::default(),
             mempool_state: Default::default(),
             _account_nonces: account_nonces.unwrap_or_default(),
         }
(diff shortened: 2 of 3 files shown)


ayeletstarkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/595)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @elintul, @giladchase, and @MohammadNassar1)

___
*[`crates/mempool/src/mempool.rs` line 170 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/595#-O5hSzDn4-kxFu6eYMah:-O5i3XYD3cRkiYThZ8XH:b-bgvdla) ([raw file](https://github.com/starkware-libs/sequencer/blob/a163e265ec7cbaac5a447f07db0a238f4d4579c6/crates/mempool/src/mempool.rs#L170)):*

When it can be included in the suspended pool?
I think get_next_eligible would return None if there is a hole.
</blockquote></details>

good catch. Since we increment the eligible transaction by one, its sequential transaction should not be in the suspended pool.
___
*[`crates/mempool/src/mempool.rs` line 195 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/595#-O5hTwcY-wsu5IHCBRPJ:-O5i1fUGDLUYAfqKZaiB:b-wp7j59) ([raw file](https://github.com/starkware-libs/sequencer/blob/a163e265ec7cbaac5a447f07db0a238f4d4579c6/crates/mempool/src/mempool.rs#L195)):*

Why do you need to check if it contains? We're confident it's in the suspended pool. If a transaction is in the pool and the queue was empty, then it's definitely in the suspended pool.
</blockquote></details>

I'm not entirely sure about the scenario you're describing, but I agree that this check seems redundant. 
Here's why: 
we remove transactions up to the received nonce across the queue, pool, and suspended pool. 
If the next eligible transaction is still in the pool, it shouldn’t be in the suspended pool.

For example, if we have transactions 0, 1, and 3 in the mempool and receive nonce 3, we remove transactions from the queue if there are any, remove transactions 0 and 1 from the pool, and remove transaction 3 from the suspended pool.
___
*[`crates/mempool/src/suspended_transaction_pool.rs` line 16 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/595#-O5hUnFM6NHO8v1hdJ29:-O5i0bxREnn6BkquOB8j:b-p6jgi8) ([raw file](https://github.com/starkware-libs/sequencer/blob/a163e265ec7cbaac5a447f07db0a238f4d4579c6/crates/mempool/src/suspended_transaction_pool.rs#L16)):*

I think it can return bool.
so it would return false if there is such txs to remove.
</blockquote></details>

I wrote this function based on the transaction pool’s `remove_up_to_nonce` function, which doesn't return anything. Why would we need to return a `bool` value to indicate if there were transactions to remove?

elintul (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/595)* status: 1 of 3 files reviewed, 6 unresolved discussions (waiting on @ayeletstarkware, @giladchase, and @MohammadNassar1)

___
*[`crates/mempool/src/suspended_transaction_pool.rs` line 7 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/595#-O5mStA_Ca78yexu6jrA:-O5mStA_Ca78yexu6jrB:be8da7k) ([raw file](https://github.com/starkware-libs/sequencer/blob/bf80127e4a80e7da421d6b5be084ae767768e247/crates/mempool/src/suspended_transaction_pool.rs#L7)):*
> ```XML
> use crate::mempool::TransactionReference;
> 
> type AddressNonceToTransaction = HashMap<(ContractAddress, Nonce), TransactionReference>;
> ```

Please shortly explain the future use of this DS.



_Code quote:_
```XML
HashMap<(ContractAddress, Nonce), TransactionReference>
```
___
*[`crates/mempool/src/suspended_transaction_pool.rs` line 16 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/595#-O5mSzQuEGkyGdbB-HMs:-O5mSzQuEGkyGdbB-HMt:b-44jrun) ([raw file](https://github.com/starkware-libs/sequencer/blob/bf80127e4a80e7da421d6b5be084ae767768e247/crates/mempool/src/suspended_transaction_pool.rs#L16)):*
> ```XML
> impl SuspendedTransactionPool {
>     // TODO(Ayelet): Implement this function.
>     pub fn remove_up_to_nonce_and_sequential(&mut self, _address: ContractAddress, _nonce: Nonce) {}
> ```

Name not clear enough for me to understand what this function does.



_Code quote:_
```XML
remove_up_to_nonce_and_sequential
```
___
*[`crates/mempool/src/suspended_transaction_pool.rs` line 18 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/595#-O5mSoSc1iiR4VD3sCi0:-O5mSoSdA-a7O2VwgIFq:b-fatq4r) ([raw file](https://github.com/starkware-libs/sequencer/blob/bf80127e4a80e7da421d6b5be084ae767768e247/crates/mempool/src/suspended_transaction_pool.rs#L18)):*
> ```XML
>     pub fn remove_up_to_nonce_and_sequential(&mut self, _address: ContractAddress, _nonce: Nonce) {}
> 
>     // TODO(Ayelet): Implement this function.
> ```

Maybe `unimplemented!()` is better?



_Code quote:_
```XML
// TODO(Ayelet): Implement this function.
```

PR #735, opened 2024-09-09
feat(consensus): add active_proposals to context

asmaastarkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/735)

### DIFF
--- a/crates/sequencing/papyrus_consensus/src/manager_test.rs
+++ b/crates/sequencing/papyrus_consensus/src/manager_test.rs
@@ -54,16 +54,22 @@ mock! {
         type Block = TestBlock;
         type ProposalChunk = Transaction;
 
-        async fn build_proposal(&self, height: BlockNumber) -> (
+        async fn build_proposal(&mut self, height: BlockNumber) -> (
             mpsc::Receiver<Transaction>,
             oneshot::Receiver<ProposalContentId>
         );
 
         async fn validate_proposal(
-            &self,
+            &mut self,
             height: BlockNumber,
             content: mpsc::Receiver<Transaction>
         ) -> oneshot::Receiver<ProposalContentId>;
+
+        async fn get_proposal(
+            &self,
+            height: BlockNumber,
+            id: ProposalContentId,
+        ) -> mpsc::Receiver<Transaction>;
 
         async fn validators(&self, height: BlockNumber) -> Vec<ValidatorId>;
 
--- a/crates/sequencing/papyrus_consensus/src/test_utils.rs
+++ b/crates/sequencing/papyrus_consensus/src/test_utils.rs
@@ -30,16 +30,22 @@ mock! {
         type Block = TestBlock;
         type ProposalChunk = u32;
 
-        async fn build_proposal(&self, height: BlockNumber) -> (
+        async fn build_proposal(&mut self, height: BlockNumber) -> (
             mpsc::Receiver<u32>,
             oneshot::Receiver<ProposalContentId>
         );
 
         async fn validate_proposal(
-            &self,
+            &mut self,
             height: BlockNumber,
             content: mpsc::Receiver<u32>
         ) -> oneshot::Receiver<ProposalContentId>;
+
+        async fn get_proposal(
+            &self,
+            height: BlockNumber,
+            id: ProposalContentId,
+        ) -> mpsc::Receiver<u32>;
 
         async fn validators(&self, height: BlockNumber) -> Vec<ValidatorId>;
 
--- a/crates/sequencing/papyrus_consensus/src/types.rs
+++ b/crates/sequencing/papyrus_consensus/src/types.rs
@@ -45,7 +45,7 @@ pub trait ConsensusContext {
     ///   and building it. If the block fails to be built, the Sender will be dropped by
     ///   ConsensusContext.
     async fn build_proposal(
-        &self,
+        &mut self,
         height: BlockNumber,
     ) -> (mpsc::Receiver<Self::ProposalChunk>, oneshot::Receiver<ProposalContentId>);
 
@@ -62,10 +62,26 @@ pub trait ConsensusContext {
     /// - A receiver for the block id. If a valid block cannot be built the Sender will be dropped
     ///   by ConsensusContext.
     async fn validate_proposal(
-        &self,
+        &mut self,
         height: BlockNumber,
         content: mpsc::Receiver<Self::ProposalChunk>,
     ) -> oneshot::Receiver<ProposalContentId>;
+
+    /// This function is called by consensus to retrieve the content of a previously built or
+    /// validated proposal. It expects that this call will return immediately, allowing
+    /// consensus to stream the block's content.
+    ///
+    /// Params:
+    /// - `height`: The height of the block that was built or validated.
+    /// - `id`: The `ProposalContentId` associated with the block's content.
+    ///
+    /// Returns:
+    /// - A receiver for the stream of the block's content.
+    async fn get_proposal(
+        &self,
+        height: BlockNumber,
+        id: ProposalContentId,
+    ) -> mpsc::Receiver<Self::ProposalChunk>;
 
     /// Get the set of validators for a given height. These are the nodes that can propose and vote
     /// on blocks.
--- a/crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs
+++ b/crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs
@@ -3,6 +3,8 @@
 mod papyrus_consensus_context_test;
 
 use core::panic;
+use std::collections::{BTreeMap, HashMap};
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
 use async_trait::async_trait;
@@ -25,7 +27,7 @@ use papyrus_storage::{StorageError, StorageReader};
 use starknet_api::block::{BlockHash, BlockNumber};
 use starknet_api::core::ContractAddress;
 use starknet_api::transaction::Transaction;
-use tracing::{debug, debug_span, info, warn, Instrument};
+use tracing::{debug, debug_span, error, info, warn, Instrument};
 
 // TODO: add debug messages and span to the tasks.
 
@@ -36,17 +38,20 @@ pub struct PapyrusConsensusBlock {
     id: BlockHash,
 }
 
+type HeightToIdToContent = BTreeMap<BlockNumber, HashMap<ProposalContentId, Vec<Transaction>>>;
+
 pub struct PapyrusConsensusContext {
     storage_reader: StorageReader,
     network_broadcast_sender: BroadcastTopicSender<ConsensusMessage>,
     validators: Vec<ValidatorId>,
     sync_broadcast_sender: Option<BroadcastTopicSender<Vote>>,
+    // Proposal building/validating returns immediately, leaving the actual processing to a spawned
+    // task. The spawned task processes the proposal asynchronously and updates the
+    // valid_proposals map upon completion, ensuring consistency across tasks.
+    valid_proposals: Arc<Mutex<HeightToIdToContent>>,
 }
 
 impl PapyrusConsensusContext {
-    // TODO(dvir): remove the dead code attribute after we will use this function.
-    #[allow(dead_code)]
-    #[allow(missing_docs)]
     pub fn new(
         storage_reader: StorageReader,
         network_broadcast_sender: BroadcastTopicSender<ConsensusMessage>,
@@ -58,6 +63,7 @@ impl PapyrusConsensusContext {
             network_broadcast_sender,
             validators: (0..num_validators).map(ContractAddress::from).collect(),
             sync_broadcast_sender,
+            valid_proposals: Arc::new(Mutex::new(BTreeMap::new())),
         }
     }
 }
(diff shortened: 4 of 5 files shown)


matan-starkware (changes requested):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/735)* status: all files reviewed, 2 unresolved discussions (waiting on @asmaastarkware)

___
*[`crates/sequencing/papyrus_consensus/src/papyrus_consensus_context.rs` line 50 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/735#-O6pCfEn4VOCG7RnoYfd:-O6pCfEo-Y0ojzf_KmKb:b-gmc9it) ([raw file](https://github.com/starkware-libs/sequencer/blob/334e3c14aeaa72aeb99091e3ef3854c14d45a858/crates/sequencing/papyrus_consensus/src/papyrus_consensus_context.rs#L50)):*
> ```XML
>     validators: Vec<ValidatorId>,
>     sync_broadcast_sender: Option<BroadcastTopicSender<Vote>>,
>     active_proposals: Arc<Mutex<ActiveProposals>>,
> ```

Hmm I don't like that we have to do this, but I think I understand the need:
1. We want to return immediately for build/validate which leads to actual handling being done in a separate task
2. We want to update the state of context when that completes

I can think of one other solution that avoids the mutex. The context holds a channel. When it spawns a proposal task it passes in the sender side of the channel. When the proposal is built, the spawned task sends the proposal on the channel. This is then handled lazily. Whenever `get_proposal` is called, the context takes all of the proposals sent on the channel and updates `active_proposals`. WDYT?



_Code quote:_
```XML
    active_proposals: Arc<Mutex<ActiveProposals>>,
```
___
*[`crates/sequencing/papyrus_consensus/src/papyrus_consensus_context.rs` line 82 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/735#-O6pD4hrDMKL9TxIdCkn:-O6pD4hrDMKL9TxIdCko:b-lyda8n) ([raw file](https://github.com/starkware-libs/sequencer/blob/334e3c14aeaa72aeb99091e3ef3854c14d45a858/crates/sequencing/papyrus_consensus/src/papyrus_consensus_context.rs#L82)):*
> ```XML
>             }
>         }
>         rx
> ```

I still prefer to avoid multi nesting when possible :)



_Suggestion:_
```XML
        let Some(proposals_at_height) = self.active_proposals.lock().await.get(&height) else {
            return rx
        };
        
        if let Some(transactions) = proposals_at_height.get(&id) {
            let _ = tx.send(transactions.clone()).await;
        }
        rx
```

matan-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/735)* status: all files reviewed, 4 unresolved discussions (waiting on @asmaastarkware)

___
*[`crates/sequencing/papyrus_consensus/src/papyrus_consensus_context.rs` line 73 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/735#-O6v5vF84Tb-0fniV4ox:-O6v5vF96ofGR9xYRmxz:bl0ys69) ([raw file](https://github.com/starkware-libs/sequencer/blob/01fb2f79dcdfe3192b58b3060e1e1306919ffa7b/crates/sequencing/papyrus_consensus/src/papyrus_consensus_context.rs#L73)):*
> ```XML
>         id: ProposalContentId,
>     ) -> mpsc::Receiver<Vec<Transaction>> {
>         let (mut tx, rx) = mpsc::channel(1);
> ```

we also use `tx` for transaction. So let's stick to writing sender and receiver.



_Code quote:_
```XML
tx, rx
```
___
*[`crates/sequencing/papyrus_consensus/src/papyrus_consensus_context.rs` line 86 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/735#-O6v68sJA_Dd7aM_MKba:-O6v68sJA_Dd7aM_MKbb:br615nh) ([raw file](https://github.com/starkware-libs/sequencer/blob/01fb2f79dcdfe3192b58b3060e1e1306919ffa7b/crates/sequencing/papyrus_consensus/src/papyrus_consensus_context.rs#L86)):*
> ```XML
>         };
>         let _ = tx.send(transactions).await;
>         rx
> ```

We need to be spawning here like we do with build/validate. The point is to return immediately the receiver and let the spawned task handle streaming out in parallel.



_Code quote:_
```XML
        let transactions = {
            let active_proposals_lock =
                self.active_proposals.lock().expect("Failed to lock active proposals");
            let Some(proposals_at_height) = active_proposals_lock.get(&height) else {
                return rx;
            };
            let Some(transactions) = proposals_at_height.get(&id) else {
                return rx;
            };
            transactions.clone()
        };
        let _ = tx.send(transactions).await;
        rx
```
___
*[`crates/sequencing/papyrus_consensus/src/papyrus_consensus_context.rs` line 135 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/735#-O6v6UpE8W_tddRn0Us3:-O6v6UpFAEgsRzCvK1_d:bd7cxrn) ([raw file](https://github.com/starkware-libs/sequencer/blob/01fb2f79dcdfe3192b58b3060e1e1306919ffa7b/crates/sequencing/papyrus_consensus/src/papyrus_consensus_context.rs#L135)):*
> ```XML
> 
>                 let mut proposals =
>                     active_proposals.lock().expect("Failed to lock active proposals");
> ```

It's not exactly failed to lock. IIRC this is about the lock being "poisoned" because a thread holding the lock panic'd. Look at the documentation on the std mutex.



_Code quote:_
```XML
                let mut proposals =
                    active_proposals.lock().expect("Failed to lock active proposals");
```

matan-starkware (changes requested):
> Reviewed 7 of 7 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/735)* status: all files reviewed, 5 unresolved discussions (waiting on @asmaastarkware)

___
*[`crates/sequencing/papyrus_consensus/src/types.rs` line 81 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/735#-O78FPqG9GtvaYxxyqsg:-O78FPqHABf1qs-l38_Q:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/761758f33e2e3679154512742d2aedf1107815ce/crates/sequencing/papyrus_consensus/src/types.rs#L81)):*
> ```XML
>     ///
>     /// Returns:
>     /// - A receiver for the stream of the block's content.
> ```





_Suggestion:_
```XML
    /// This function is called by consensus to retrieve the content of a previously built or
    /// validated proposal. It expects that this call will return immediately, allowing
    /// consensus to stream the block's content.
    ///
    /// Params:
    /// - `height`: The height of the block that was built or validated.
    /// - `id`: The `ProposalContentId` associated with the block's content.
    ///
    /// Returns:
    /// - A receiver for the stream of the block's content.
```
___
*[`crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs` line 41 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/735#-O78FoWK6BBGMi18Zzz9:-O78FoWK6BBGMi18ZzzA:b-d2dcmr) ([raw file](https://github.com/starkware-libs/sequencer/blob/761758f33e2e3679154512742d2aedf1107815ce/crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs#L41)):*
> ```XML
> }
> 
> type ActiveProposals = BTreeMap<BlockNumber, HashMap<ProposalContentId, Vec<Transaction>>>;
> ```

I don't love the term Active, since it can be for old heights.



_Suggestion:_
```XML
type HeightToIdToContent = BTreeMap<BlockNumber, HashMap<ProposalContentId, Vec<Transaction>>>;
```
___
*[`crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs` line 48 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/735#-O78FiMF9lr0CDM1nZ5w:-O78FiMF9lr0CDM1nZ5x:b-65k4ro) ([raw file](https://github.com/starkware-libs/sequencer/blob/761758f33e2e3679154512742d2aedf1107815ce/crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs#L48)):*
> ```XML
>     validators: Vec<ValidatorId>,
>     sync_broadcast_sender: Option<BroadcastTopicSender<Vote>>,
>     active_proposals: Arc<Mutex<ActiveProposals>>,
> ```

Add a comment explaining why we have the mutex

WDYT of this name? 



_Suggestion:_
```XML
    valid_proposals: Arc<Mutex<ActiveProposals>>,
```
___
*[`crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs` line 118 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/735#-O78HJI77o6uz8AKSW5m:-O78HJI77o6uz8AKSW5n:b-c81e61) ([raw file](https://github.com/starkware-libs/sequencer/blob/761758f33e2e3679154512742d2aedf1107815ce/crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs#L118)):*
> ```XML
>                 proposals.entry(height).or_default().insert(block_hash, transactions);
> 
>                 fin_sender.send(block_hash).expect("Send should succeed");
> ```

Done after inserting the proposal into the map to avoid race conditions between insertion and calls to `get_proposal`.

(and below)



_Code quote:_
```XML
                fin_sender.send(block_hash).expect("Send should succeed");
```
___
*[`crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs` line 206 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/735#-O78He9J5Go7PUrEXreY:-O78He9J5Go7PUrEXreZ:b1v7t9w) ([raw file](https://github.com/starkware-libs/sequencer/blob/761758f33e2e3679154512742d2aedf1107815ce/crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs#L206)):*
> ```XML
>                 let Some(transactions) = proposals_at_height.get(&id) else {
>                     return;
>                 };
> ```

Both of these should be an error log, since they represent an inconsistency between the state of consensus and the context.



_Code quote:_
```XML
                let Some(proposals_at_height) = active_proposals_lock.get(&height) else {
                    return;
                };
                let Some(transactions) = proposals_at_height.get(&id) else {
                    return;
                };
```

matan-starkware (changes requested):
> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/735)* status: all files reviewed, 1 unresolved discussion (waiting on @asmaastarkware)

___
*[`crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs` line 48 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/735#-O7P2yR05QQeTk4ZuQWP:-O7P2yR16dFT9uCXr__T:b-pb7ql6) ([raw file](https://github.com/starkware-libs/sequencer/blob/f002111c8d00ff6d3db10a9fb205df3ac42bb2e6/crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs#L48)):*
> ```XML
>     validators: Vec<ValidatorId>,
>     sync_broadcast_sender: Option<BroadcastTopicSender<Vote>>,
>     // Mutex ensures safe concurrent access to valid_proposals across async tasks.
> ```

This is more a definition of a mutex. I meant explain why we need it here. "proposal handling returns immediately, streaming in spawned task to process actual building, can only be added once proposal is complete" (but real sentences)



_Code quote:_
```XML
    // Mutex ensures safe concurrent access to valid_proposals across async tasks.
```

matan-starkware (changes requested):
> Reviewed 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/735)* status: all files reviewed, 1 unresolved discussion (waiting on @asmaastarkware)

___
*[`crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs` line 50 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/735#-O7TAd9v2G-u2NW47R2p:-O7TAd9wAumfw2Ia9yUN:b-g6u5no) ([raw file](https://github.com/starkware-libs/sequencer/blob/538e1284d37aa85889e20757447ced48cdb5205f/crates/sequencing/papyrus_consensus_orchestrator/src/papyrus_consensus_context.rs#L50)):*
> ```XML
>     // Proposal building/validating runs asynchronously in a spawned task.
>     // And the valid_proposals map is updated only after the entire proposal is completed,
>     // ensuring consistency across tasks.
> ```

Proposal building/validating returns immediately, actually processing the proposal in a spawned task. The spawned task, after finish handling the proposal, updates the `valid_proposals` map. 



_Code quote:_
```XML
    // Proposal building/validating runs asynchronously in a spawned task.
    // And the valid_proposals map is updated only after the entire proposal is completed,
    // ensuring consistency across tasks.
```

matan-starkware (approved):
> Reviewed 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/735)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @asmaastarkware)

PR #809, opened 2024-09-16
feat(blockifier): set sierra gas limits based on resource bounds

nimrod-starkware:
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)

### DIFF
--- a/crates/blockifier/src/blockifier/stateful_validator.rs
+++ b/crates/blockifier/src/blockifier/stateful_validator.rs
@@ -73,9 +73,8 @@ impl<S: StateReader> StatefulValidator<S> {
         }
 
         // `__validate__` call.
-        let versioned_constants = &tx_context.block_context.versioned_constants();
         let (_optional_call_info, actual_cost) =
-            self.validate(&tx, versioned_constants.tx_default_initial_gas())?;
+            self.validate(&tx, tx_context.initial_sierra_gas())?;
 
         // Post validations.
         PostValidationReport::verify(&tx_context, &actual_cost)?;
--- a/crates/blockifier/src/context.rs
+++ b/crates/blockifier/src/context.rs
@@ -5,11 +5,12 @@ use papyrus_config::{ParamPath, ParamPrivacyInput, SerializedParam};
 use serde::{Deserialize, Serialize};
 use starknet_api::block::GasPriceVector;
 use starknet_api::core::{ChainId, ContractAddress};
-use starknet_api::transaction::GasVectorComputationMode;
+use starknet_api::transaction::{AllResourceBounds, GasVectorComputationMode, ValidResourceBounds};
 
 use crate::blockifier::block::BlockInfo;
 use crate::bouncer::BouncerConfig;
 use crate::transaction::objects::{
+    CurrentTransactionInfo,
     FeeType,
     HasRelatedFeeType,
     TransactionInfo,
@@ -43,6 +44,23 @@ impl TransactionContext {
             .block_info
             .gas_prices
             .get_gas_prices_by_fee_type(&self.tx_info.fee_type())
+    }
+
+    /// Returns the initial Sierra gas of the transaction.
+    /// This value is used to limit the transaction's run.
+    // TODO(tzahi): replace returned value from u64 to GasAmount.
+    pub fn initial_sierra_gas(&self) -> u64 {
+        match &self.tx_info {
+            TransactionInfo::Deprecated(_)
+            | TransactionInfo::Current(CurrentTransactionInfo {
+                resource_bounds: ValidResourceBounds::L1Gas(_),
+                ..
+            }) => self.block_context.versioned_constants.default_initial_gas_cost(),
+            TransactionInfo::Current(CurrentTransactionInfo {
+                resource_bounds: ValidResourceBounds::AllResources(AllResourceBounds { l2_gas, .. }),
+                ..
+            }) => l2_gas.max_amount.0,
+        }
     }
 }
 
--- a/crates/blockifier/src/execution/contract_address_test.rs
+++ b/crates/blockifier/src/execution/contract_address_test.rs
@@ -34,7 +34,7 @@ fn test_calculate_contract_address() {
             calldata,
             entry_point_selector: selector_from_name("test_contract_address"),
             storage_address: deployer_address,
-            initial_gas: versioned_constants.os_constants.gas_costs.default_initial_gas_cost,
+            initial_gas: versioned_constants.default_initial_gas_cost(),
             ..Default::default()
         };
         let contract_address =
--- a/crates/blockifier/src/execution/entry_point_test.rs
+++ b/crates/blockifier/src/execution/entry_point_test.rs
@@ -198,7 +198,7 @@ fn run_security_test(
         entry_point_selector: selector_from_name(entry_point_name),
         calldata,
         storage_address: security_contract.get_instance_address(0),
-        initial_gas: versioned_constants.os_constants.gas_costs.default_initial_gas_cost,
+        initial_gas: versioned_constants.default_initial_gas_cost(),
         ..Default::default()
     };
     let error = match entry_point_call.execute_directly(state) {
--- a/crates/blockifier/src/execution/execution_utils.rs
+++ b/crates/blockifier/src/execution/execution_utils.rs
@@ -44,24 +44,33 @@ pub const SEGMENT_ARENA_BUILTIN_SIZE: usize = 3;
 
 /// A wrapper for execute_entry_point_call that performs pre and post-processing.
 pub fn execute_entry_point_call_wrapper(
-    call: CallEntryPoint,
+    mut call: CallEntryPoint,
     contract_class: ContractClass,
     state: &mut dyn State,
     resources: &mut ExecutionResources,
     context: &mut EntryPointExecutionContext,
     remaining_gas: &mut u64,
 ) -> EntryPointExecutionResult<CallInfo> {
-    let tracked_resource = contract_class
+    let contract_tracked_resource = contract_class
         .tracked_resource(&context.versioned_constants().min_compiler_version_for_sierra_gas);
     // Note: no return statements (explicit or implicit) should be added between the push and the
     // pop commands.
 
     // Once we ran with CairoSteps, we will continue to run using it for all nested calls.
-    if context.tracked_resource_stack.last().is_some_and(|x| *x == TrackedResource::CairoSteps) {
-        context.tracked_resource_stack.push(TrackedResource::CairoSteps);
-    } else {
-        context.tracked_resource_stack.push(tracked_resource);
-    }
+    match context.tracked_resource_stack.last() {
+        Some(TrackedResource::CairoSteps) => {
+            context.tracked_resource_stack.push(TrackedResource::CairoSteps)
+        }
+        Some(TrackedResource::SierraGas) => {
+            if contract_tracked_resource == TrackedResource::CairoSteps {
+                // Switching from SierraGas to CairoSteps: override initial_gas with a high value so
+                // it won't limit the run.
+                call.initial_gas = context.versioned_constants().default_initial_gas_cost();
+            };
+            context.tracked_resource_stack.push(contract_tracked_resource)
+        }
+        None => context.tracked_resource_stack.push(contract_tracked_resource),
+    };
 
     let res = execute_entry_point_call(call, contract_class, state, resources, context);
     context.tracked_resource_stack.pop();
--- a/crates/blockifier/src/execution/syscalls/hint_processor.rs
+++ b/crates/blockifier/src/execution/syscalls/hint_processor.rs
@@ -857,6 +857,7 @@ pub fn execute_library_call(
         storage_address: syscall_handler.storage_address(),
         caller_address: syscall_handler.caller_address(),
         call_type: CallType::Delegate,
+        // NOTE: this value might be overridden later on.
         initial_gas: *remaining_gas,
     };
 
--- a/crates/blockifier/src/execution/syscalls/mod.rs
+++ b/crates/blockifier/src/execution/syscalls/mod.rs
@@ -195,6 +195,7 @@ pub fn call_contract(
         storage_address,
         caller_address: syscall_handler.storage_address(),
         call_type: CallType::Call,
+        // NOTE: this value might be overridden later on.
         initial_gas: *remaining_gas,
     };
 
--- a/crates/blockifier/src/fee/fee_test.rs
+++ b/crates/blockifier/src/fee/fee_test.rs
@@ -1,10 +1,17 @@
 use assert_matches::assert_matches;
 use cairo_vm::types::builtin_name::BuiltinName;
 use rstest::rstest;
-use starknet_api::block::NonzeroGasPrice;
+use starknet_api::block::{GasPrice, NonzeroGasPrice};
 use starknet_api::execution_resources::{GasAmount, GasVector};
 use starknet_api::invoke_tx_args;
-use starknet_api::transaction::{Fee, GasVectorComputationMode, Resource, ValidResourceBounds};
+use starknet_api::transaction::{
+    AllResourceBounds,
+    Fee,
+    GasVectorComputationMode,
+    Resource,
+    ResourceBounds,
+    ValidResourceBounds,
+};
 
 use crate::blockifier::block::GasPrices;
 use crate::context::BlockContext;
@@ -26,7 +33,12 @@ use crate::test_utils::{
     MAX_L1_GAS_AMOUNT,
 };
 use crate::transaction::objects::FeeType;
-use crate::transaction::test_utils::{account_invoke_tx, all_resource_bounds, l1_resource_bounds};
+use crate::transaction::test_utils::{
+    account_invoke_tx,
+    all_resource_bounds,
+    block_context,
+    l1_resource_bounds,
+};
 use crate::utils::u64_from_usize;
 use crate::versioned_constants::VersionedConstants;
 
@@ -346,4 +358,33 @@ fn test_get_fee_by_gas_vector_overflow(
     let gas_vector =
         GasVector { l1_gas: l1_gas.into(), l1_data_gas: l1_data_gas.into(), l2_gas: l2_gas.into() };
     assert_eq!(get_fee_by_gas_vector(&block_info, gas_vector, &FeeType::Eth), Fee(u128::MAX));
+}
+
+#[rstest]
+#[case::default(
+    VersionedConstants::create_for_account_testing().default_initial_gas_cost(),
+    GasVectorComputationMode::NoL2Gas
+)]
+#[case::from_l2_gas(4321, GasVectorComputationMode::All)]
+fn test_initial_sierra_gas(
+    #[case] expected: u64,
+    #[case] gas_mode: GasVectorComputationMode,
+    block_context: BlockContext,
+) {
+    let resource_bounds = match gas_mode {
+        GasVectorComputationMode::NoL2Gas => ValidResourceBounds::L1Gas(ResourceBounds {
+            max_amount: GasAmount(1234),
+            max_price_per_unit: GasPrice(56),
+        }),
+        GasVectorComputationMode::All => ValidResourceBounds::AllResources(AllResourceBounds {
+            l2_gas: ResourceBounds {
+                max_amount: GasAmount(expected),
+                max_price_per_unit: GasPrice(1),
+            },
+            ..Default::default()
+        }),
+    };
+    let account_tx = account_invoke_tx(invoke_tx_args!(resource_bounds));
+    let actual = block_context.to_tx_context(&account_tx).initial_sierra_gas();
+    assert_eq!(actual, expected)
 }
--- a/crates/blockifier/src/transaction/account_transaction.rs
+++ b/crates/blockifier/src/transaction/account_transaction.rs
@@ -811,7 +811,7 @@ impl<U: UpdatableState> ExecutableTransaction<U> for AccountTransaction {
         )?;
 
         // Run validation and execution.
-        let mut remaining_gas = block_context.versioned_constants.tx_default_initial_gas();
+        let mut remaining_gas = tx_context.initial_sierra_gas();
         let ValidateExecuteCallInfo {
             validate_call_info,
             execute_call_info,
--- a/crates/blockifier/src/transaction/account_transactions_test.rs
+++ b/crates/blockifier/src/transaction/account_transactions_test.rs
@@ -478,8 +478,7 @@ fn test_max_fee_limit_validate(
     .unwrap_err()
     .to_string();
     block_context.versioned_constants.validate_max_n_steps = old_validate_max_n_steps;
-
-    assert!(error_trace.contains("no remaining steps"));
+    assert!(error_trace.contains("no remaining steps") | error_trace.contains("Out of gas"))
 }
 
 #[rstest]
--- a/crates/blockifier/src/transaction/transaction_execution.rs
+++ b/crates/blockifier/src/transaction/transaction_execution.rs
@@ -141,7 +141,7 @@ impl<U: UpdatableState> ExecutableTransaction<U> for L1HandlerTransaction {
 
         let mut execution_resources = ExecutionResources::default();
         let mut context = EntryPointExecutionContext::new_invoke(tx_context.clone(), true);
-        let mut remaining_gas = block_context.versioned_constants.tx_default_initial_gas();
+        let mut remaining_gas = tx_context.initial_sierra_gas();
         let execute_call_info =
             self.run_execute(state, &mut execution_resources, &mut context, &mut remaining_gas)?;
         let l1_handler_payload_size = self.payload_size();
--- a/crates/blockifier/src/transaction/transactions_test.rs
+++ b/crates/blockifier/src/transaction/transactions_test.rs
@@ -153,8 +153,8 @@ static VERSIONED_CONSTANTS: LazyLock<VersionedConstants> =
     LazyLock::new(VersionedConstants::create_for_testing);
 
 #[fixture]
-fn tx_default_initial_gas() -> u64 {
-    VERSIONED_CONSTANTS.tx_default_initial_gas()
+fn default_initial_gas_cost() -> u64 {
+    VERSIONED_CONSTANTS.default_initial_gas_cost()
 }
 
 #[fixture]
@@ -217,7 +217,8 @@ fn expected_validate_call_info(
             storage_address,
             caller_address: ContractAddress::default(),
             call_type: CallType::Call,
-            initial_gas: tx_default_initial_gas(),
+            // TODO(tzahi): make compatible with l2 gas in resource bounds once used in tests.
+            initial_gas: default_initial_gas_cost(),
         },
         // The account contract we use for testing has trivial `validate` functions.
         resources,
@@ -389,7 +390,7 @@ fn add_kzg_da_resources_to_resources_mapping(
         },
         validate_gas_consumed: 0,
         execute_gas_consumed: 0,
-        inner_call_initial_gas: versioned_constants_for_account_testing().os_constants.gas_costs.default_initial_gas_cost,
+        inner_call_initial_gas: versioned_constants_for_account_testing().default_initial_gas_cost(),
     },
     CairoVersion::Cairo0)]
 #[case::with_cairo1_account(
@@ -402,9 +403,10 @@ fn add_kzg_da_resources_to_resources_mapping(
         validate_gas_consumed: 4740, // The gas consumption results from parsing the input
             // arguments.
         execute_gas_consumed: 163280,
-        inner_call_initial_gas: 9999395189,
+        inner_call_initial_gas: versioned_constants_for_account_testing().default_initial_gas_cost(),
     },
     CairoVersion::Cairo1)]
+// TODO(Tzahi): Add calls to cairo1 test contracts (where gas flows to and from the inner call).
 fn test_invoke_tx(
     max_l1_resource_bounds: ValidResourceBounds,
     #[case] expected_arguments: ExpectedResultTestInvokeTx,
@@ -479,10 +481,11 @@ fn test_invoke_tx(
         call_type: CallType::Call,
         initial_gas: expected_arguments.inner_call_initial_gas,
     };
+    let expected_validated_call = expected_validate_call_info.as_ref().unwrap().call.clone();
     let expected_execute_call = CallEntryPoint {
         entry_point_selector: selector_from_name(constants::EXECUTE_ENTRY_POINT_NAME),
-        initial_gas: tx_default_initial_gas() - expected_arguments.validate_gas_consumed,
-        ..expected_validate_call_info.as_ref().unwrap().call.clone()
+        initial_gas: expected_validated_call.initial_gas - expected_arguments.validate_gas_consumed,
+        ..expected_validated_call
     };
     let expected_return_result_retdata = Retdata(expected_return_result_calldata);
     let expected_execute_call_info = Some(CallInfo {
@@ -1565,7 +1568,7 @@ fn test_deploy_account_tx(
             entry_point_type: EntryPointType::Constructor,
             entry_point_selector: selector_from_name(abi_constants::CONSTRUCTOR_ENTRY_POINT_NAME),
             storage_address: deployed_account_address,
-            initial_gas: tx_default_initial_gas(),
+            initial_gas: default_initial_gas_cost(),
             ..Default::default()
         },
         ..Default::default()
@@ -2063,7 +2066,7 @@ fn test_l1_handler(#[values(false, true)] use_kzg_da: bool) {
             storage_address: contract_address,
             caller_address: ContractAddress::default(),
             call_type: CallType::Call,
-            initial_gas: tx_default_initial_gas(),
+            initial_gas: default_initial_gas_cost(),
         },
         execution: CallExecution {
             retdata: Retdata(vec![value]),
--- a/crates/blockifier/src/versioned_constants.rs
+++ b/crates/blockifier/src/versioned_constants.rs
@@ -253,10 +253,9 @@ impl VersionedConstants {
         self.l1_to_l2_gas_price_ratio()
     }
 
-    /// Returns the default initial gas of any transaction to run with.
-    pub fn tx_default_initial_gas(&self) -> u64 {
-        let os_consts = &self.os_constants;
-        os_consts.gas_costs.default_initial_gas_cost - os_consts.gas_costs.transaction_gas_cost
+    /// Returns the default initial gas for VM mode transactions.
+    pub fn default_initial_gas_cost(&self) -> u64 {
+        self.os_constants.gas_costs.default_initial_gas_cost
     }
 
     pub fn vm_resource_fee_cost(&self) -> &VmResourceCosts {


TzahiTaub (changes requested):
> Reviewed 3 of 3 files at r1.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed (commit messages unreviewed), 3 unresolved discussions (waiting on @nimrod-starkware and @Yoni-Starkware)

*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O6zKjs8CWJ4t5PeAlkU:-O6zKjs8CWJ4t5PeAlkV:b1untln) (no related file):*
Can you add a test that a new contract (with version>min) that signed on L2 gas have it as it's initial gas, and if it's and old compiler, we get the default initial gas?

___
___
*[`crates/blockifier/src/context.rs` line 44 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O6zHbce9pbPnCWTi9kF:-O6zHbce9pbPnCWTi9kG:bfblyms) ([raw file](https://github.com/starkware-libs/sequencer/blob/8fc285556a5fe3523469372f10499f415502ae6f/crates/blockifier/src/context.rs#L44)):*
> ```XML
>     }
> 
>     pub fn initial_sierra_gas(&self) -> u64 {
> ```

I don't know why we don't doc, but I think we need something like this.



_Suggestion:_
```XML
    /// Returns initial sierra gas to limit the transaction's run. Returns the L2 gas the user signed on if it exists, or the default gas amount (should be very high to avoid limiting with this resource) otherwise.
    pub fn initial_sierra_gas(&self) -> u64 {
```
___
*[`crates/blockifier/src/execution/entry_point.rs` line 106 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O6zK3Fh1dsNObmvKRv4:-O6zK3Fh1dsNObmvKRv5:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/8fc285556a5fe3523469372f10499f415502ae6f/crates/blockifier/src/execution/entry_point.rs#L106)):*
> ```XML
>         let contract_class = state.get_compiled_contract_class(class_hash)?;
> 
>         // If contract class can't run with cairo native, set the initial gas to infinity.
> ```





_Suggestion:_
```XML
  // If the contract class should run in VM mode, set a high initial gas so it won't limit the run.
```

nimrod-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: 1 of 4 files reviewed, 2 unresolved discussions (waiting on @TzahiTaub and @Yoni-Starkware)

*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O6zKjs8CWJ4t5PeAlkU:-O6zpj_K69TtDLlfkWiR:b-896fix) (no related file):*

Can you add a test that a new contract (with version>min) that signed on L2 gas have it as it's initial gas, and if it's and old compiler, we get the default initial gas?
</blockquote></details>

Done.

___
___
*[`crates/blockifier/src/context.rs` line 44 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O6zHbce9pbPnCWTi9kF:-O6zZlgk-S1Bt7nlEXWh:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/8fc285556a5fe3523469372f10499f415502ae6f/crates/blockifier/src/context.rs#L44)):*

I don't know why we don't doc, but I think we need something like this.
</blockquote></details>

Done.

nimrod-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on @TzahiTaub and @Yoni-Starkware)

*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O6zKjs8CWJ4t5PeAlkU:-O73evjX2Nzm0k6SnACj:b-dwkkhj) (no related file):*

It's not the test I wanted (not sure if it's possible) - I wanted to verify that we give the correct bounds according to the compiler version, assuming the user signed on L2 gas. I.e., create a `CallEntryPoint` with `initial_sierra_gas`, call its `execute` and check the `initial_gas` field was modified if the compiler version is too small.
</blockquote></details>

לא אספיק לכתוב טסט כזה....

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O73RigJ2GKEHVz83OTY:-O73b6LR9Mxw02MXWvBw:b-896fix) (no related file):*

I've searched for `.tx_initial_gas` and there are a few more places that needs to be changed to the new `initial_sierra_gas` (I think), specifically the `L1HandlerTransaction` `execute\_raw` func (and possibly two other tests).
</blockquote></details>

Done.

___
___
*[`crates/blockifier/src/context.rs` line 46 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O7-dJHPB-Il54Mj4TRL:-O73X15CEwTZO6Qoyu3R:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/3007729f1b7c8cbb50cca78c39533d51f0f45065/crates/blockifier/src/context.rs#L46)):*

:facepalm:  sorry
</blockquote></details>

Done

TzahiTaub (commented):
> Reviewed 1 of 3 files at r2, 1 of 2 files at r3, 2 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: 6 of 7 files reviewed, 1 unresolved discussion (waiting on @nimrod-starkware and @Yoni-Starkware)

*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O6zKjs8CWJ4t5PeAlkU:-O73s7jm9-pNLtKjlCQ-:b-jcpdgi) (no related file):*

לא אספיק לכתוב טסט כזה....
</blockquote></details>

Add a todo (on me)?

___
___
*[`crates/blockifier/src/transaction/transactions_test.rs` line 461 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O73ru056HDn_RBh9w3u:-O73ru056HDn_RBh9w3v:b-m8v4xr) ([raw file](https://github.com/starkware-libs/sequencer/blob/44fe21973f7f94e041b529eaea7c58f6661607d0/crates/blockifier/src/transaction/transactions_test.rs#L461)):*
> ```XML
>     let expected_execute_call = CallEntryPoint {
>         entry_point_selector: selector_from_name(constants::EXECUTE_ENTRY_POINT_NAME),
>         initial_gas: tx_context.initial_sierra_gas() - expected_arguments.validate_gas_consumed,
> ```

I think this is better reverted as the resource bound we define above are of L1 only (so by setting the expected to `tx_initial_gas` we enforce the expected behaviour).

TzahiTaub (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @nimrod-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/context.rs` line 53 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O73sNBd2_siftJXKPOJ:-O73sNBd2_siftJXKPOK:bbxg8jr) ([raw file](https://github.com/starkware-libs/sequencer/blob/da62f7703f9c43a7a5ea347f2bf5b6150395eda9/crates/blockifier/src/context.rs#L53)):*
> ```XML
>                 ValidResourceBounds::L1Gas(_) => default_initial_gas,
>                 ValidResourceBounds::AllResources(AllResourceBounds { .. }) => {
>                     panic!("Unexpected variant!!!");
> ```

What happened here?

TzahiTaub (commented):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 2 unresolved discussions (waiting on @nimrod-starkware and @Yoni-Starkware)

Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 3 unresolved discussions (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/context.rs` line 46 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O73pqAI9sQ380jlCWzf:-O73pqAI9sQ380jlCWzg:b-dhjr1w) ([raw file](https://github.com/starkware-libs/sequencer/blob/44fe21973f7f94e041b529eaea7c58f6661607d0/crates/blockifier/src/context.rs#L46)):*
> ```XML
>     /// Returns initial sierra gas to limit the transaction's run. Returns the L2 gas the user
>     /// signed on if it exists, or the default gas amount otherwise (should be very high to avoid
>     /// limiting with this resource).
> ```

In general, func docstring should not contain impl details. You can add comments inside the impl for more info (e.g. about the default)



_Suggestion:_
```XML
    /// Returns the initial Sierra gas of the transaction. 
    /// This value is used to limit the transaction's run.
```

nimrod-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 3 unresolved discussions (waiting on @TzahiTaub)

___
*[`crates/blockifier/src/context.rs` line 53 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O73sNBd2_siftJXKPOJ:-O73tKvc3GYdbQve0AkC:b3hvmuw) ([raw file](https://github.com/starkware-libs/sequencer/blob/da62f7703f9c43a7a5ea347f2bf5b6150395eda9/crates/blockifier/src/context.rs#L53)):*

What happened here?
</blockquote></details>

Some tests fail and I can't run tests locally... trying to debug it at CI (it will be reverted)

TzahiTaub (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: 0 of 11 files reviewed, 1 unresolved discussion (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/context.rs` line 65 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8qRP2aEpp5MfOgJpX7:-O8qRP2aEpp5MfOgJpX8:b-93icg8) ([raw file](https://github.com/starkware-libs/sequencer/blob/fa8178ad48cd9c8006883694a94b4e8d087dad0c/crates/blockifier/src/context.rs#L65)):*
> ```XML
>                             .transaction_gas_cost,
>                     )
>                     .unwrap_or_default(),
> ```

If the ex const is higher than the bound, we return 0.



_Code quote:_
```XML
  .unwrap_or_default(),
```

TzahiTaub (approved):

dorimedini-starkware (changes requested):
> Reviewed 2 of 11 files at r6, 7 of 9 files at r7, 2 of 2 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 14 unresolved discussions (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/context.rs` line 50 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rBhF95G2WUlQEjR6w:-O8rBhF95G2WUlQEjR6x:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/context.rs#L50)):*
> ```XML
>     /// Returns the initial Sierra gas of the transaction.
>     /// This value is used to limit the transaction's run.
>     pub fn initial_sierra_gas(&self) -> u64 {
> ```





_Suggestion:_
```XML
GasAmount
```
___
*[`crates/blockifier/src/context.rs` line 53 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rBCQH7FWxHh8ZLwoe:-O8rBCQH7FWxHh8ZLwof:b-4vrw0e) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/context.rs#L53)):*
> ```XML
>         let default_initial_gas = self.block_context.versioned_constants.tx_default_initial_gas();
>         match &self.tx_info {
>             TransactionInfo::Current(tx_info) => match tx_info.resource_bounds {
> ```

non-blocking: if you combine the destructuring you wouldn't need to define the extra var



_Suggestion:_
```XML
 TransactionInfo::Current(
     TransactionInfo {
         resource_bounds: ValidResourceBounds::AllResources(AllResourceBounds { l2_gas, .. }),
         ..
     }
 ) => bla,
 _ => self.block_context.versioned_constants.tx_default_initial_gas()
```
___
*[`crates/blockifier/src/context.rs` line 65 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rAdJT1zVkWOzSuLLg:-O8rAdJT1zVkWOzSuLLh:brb7wc4) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/context.rs#L65)):*
> ```XML
>                             .transaction_gas_cost,
>                     )
>                     .unwrap_or_default(),
> ```

implement and use `saturating_sub` on `GasAmount`.

@Yoni-Starkware this should saturate (as user can provide l2_gas amount of 1), right?
is this related to [this item](https://starkware.monday.com/boards/6987445459/pulses/6987445571)?



_Suggestion:_
```XML
                    .max_amount
                    .saturating_sub(
                        self.block_context
                            .versioned_constants
                            .os_constants
                            .gas_costs
                            .transaction_gas_cost,
                    ),
```
___
*[`crates/blockifier/src/versioned_constants.rs` line 264 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rBkMm4NqbBw-u00YF:-O8rBkMm4NqbBw-u00YG:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/versioned_constants.rs#L264)):*
> ```XML
>     pub fn default_initial_gas_cost(&self) -> u64 {
>         self.os_constants.gas_costs.default_initial_gas_cost
>     }
> ```





_Suggestion:_
```XML
    pub fn default_initial_gas_cost(&self) -> GasAmount {
        self.os_constants.gas_costs.default_initial_gas_cost.into()
    }
```
___
*[`crates/blockifier/src/execution/contract_class.rs` line 71 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rBzcw8LWKHy355v7q:-O8rBzcw8LWKHy355v7r:bxsb667) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/execution/contract_class.rs#L71)):*
> ```XML
>         matches!(self, TrackedResource::CairoSteps)
>     }
> }
> ```

I don't like this because users should `match` on the `TrackedResource` enum.
consider using `match` below, or move this to a test util area (only usage is in test)



_Code quote:_
```XML
impl TrackedResource {
    pub fn is_cairo_steps(&self) -> bool {
        matches!(self, TrackedResource::CairoSteps)
    }
}
```
___
*[`crates/blockifier/src/execution/entry_point.rs` line 153 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rCMKQ0qYfhs_6nq8b:-O8rCMKQ0qYfhs_6nq8c:bt6rkgk) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/execution/entry_point.rs#L153)):*
> ```XML
>         {
>             self.initial_gas = versioned_constants.default_initial_gas_cost()
>         }
> ```

should this be next to the check where we check if we are running steps or not?
this should also depend on "parent context", right?



_Code quote:_
```XML
        // If the contract class should run in VM mode, override initial_gas with a high value so it
        // won't limit the run.
        let versioned_constants = context.tx_context.block_context.versioned_constants();
        if contract_class.tracked_resource(&versioned_constants.min_compiler_version_for_sierra_gas)
            == TrackedResource::CairoSteps
        {
            self.initial_gas = versioned_constants.default_initial_gas_cost()
        }
```
___
*[`crates/blockifier/src/fee/fee_test.rs` line 364 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rCVtu9DUR_TmpNi81:-O8rCVtu9DUR_TmpNi82:b-nheakx) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/fee/fee_test.rs#L364)):*
> ```XML
> #[case::default(VersionedConstants::create_for_account_testing().tx_default_initial_gas(),
>     GasVectorComputationMode::NoL2Gas
> )]
> ```

manually format macro contents



_Suggestion:_
```XML
#[case::default(
    VersionedConstants::create_for_account_testing().tx_default_initial_gas(),
    GasVectorComputationMode::NoL2Gas
)]
```
___
*[`crates/blockifier/src/fee/fee_test.rs` line 368 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rCbjbEC-QqJ2aWMJo:-O8rCbjbEC-QqJ2aWMJp:b-rtagwg) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/fee/fee_test.rs#L368)):*
> ```XML
> fn test_initial_sierra_gas(
>     #[case] expected: u64,
>     // #[case] resource_bounds: ValidResourceBounds,
> ```

delete



_Code quote:_
```XML
/ #[case] resource_bounds: ValidResourceBounds,
```
___
*[`crates/blockifier/src/fee/fee_test.rs` line 372 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rC_h-3DQk2VZaGT38:-O8rC_h-3DQk2VZaGT39:bpyx8j3) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/fee/fee_test.rs#L372)):*
> ```XML
>     block_context: BlockContext,
> ) {
>     use starknet_api::block::GasPrice;
> ```

move to top of module



_Code quote:_
```XML
use starknet_api::block::GasPrice;
```
___
*[`crates/blockifier/src/fee/fee_test.rs` line 378 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rD1Ti6s9nb5jYMmn5:-O8rD1Ti6s9nb5jYMmn6:bebpzi5) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/fee/fee_test.rs#L378)):*
> ```XML
>         GasVectorComputationMode::NoL2Gas => ValidResourceBounds::L1Gas(ResourceBounds {
>             max_amount: GasAmount(3000),
>             max_price_per_unit: GasPrice(9),
> ```

where are the magic numbers 5000, 3000 and 9 from...?



_Code quote:_
```XML
            max_amount: GasAmount(3000),
            max_price_per_unit: GasPrice(9),
```
___
*[`crates/blockifier/src/fee/fee_test.rs` line 389 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rD9bJ9WOk-w3uo5D3:-O8rD9bJ9WOk-w3uo5D4:betts4c) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/fee/fee_test.rs#L389)):*
> ```XML
>     };
>     let tx_args = invoke_tx_args!(resource_bounds);
>     let account_tx = account_invoke_tx(tx_args);
> ```

combine these lines



_Code quote:_
```XML
    let tx_args = invoke_tx_args!(resource_bounds);
    let account_tx = account_invoke_tx(tx_args);
```
___
*[`crates/blockifier/src/transaction/account_transactions_test.rs` line 536 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rDHlsDeO1WSOJddMl:-O8rDHlt1Cyp-kW8Cvpp:blem7go) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/transaction/account_transactions_test.rs#L536)):*
> ```XML
>     } else {
>         assert!(error_trace.contains("no remaining steps") | error_trace.contains("Out of gas"));
>     }
> ```

convert to `match`



_Code quote:_
```XML
    if resource_bounds.get_gas_vector_computation_mode() == GasVectorComputationMode::NoL2Gas {
        assert!(error_trace.contains("no remaining steps"));
    } else {
        assert!(error_trace.contains("no remaining steps") | error_trace.contains("Out of gas"));
    }
```
___
*[`crates/blockifier/src/transaction/transactions_test.rs` line 223 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rDKcP9TmzfK5ei3zn:-O8rDKcP9TmzfK5ei3zo:b63fhti) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/transaction/transactions_test.rs#L223)):*
> ```XML
>             } else {
>                 tx_default_initial_gas()
>             },
> ```

`match`



_Code quote:_
```XML
            initial_gas: if tracked_resource.is_cairo_steps() {
                VERSIONED_CONSTANTS.default_initial_gas_cost()
            } else {
                tx_default_initial_gas()
            },
```
___
*[`crates/blockifier/src/transaction/transactions_test.rs` line 262 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rDOP5Bc62L6E6vhy3:-O8rDOP5Bc62L6E6vhy4:b-3fn8de) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/transaction/transactions_test.rs#L262)):*
> ```XML
>         call_type: CallType::Call,
>         initial_gas: tx_context
>             .block_context
> ```

why this addition?



_Code quote:_
```XML
        initial_gas: tx_context
            .block_context
```

Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 14 unresolved discussions (waiting on @dorimedini-starkware and @nimrod-starkware)

___
*[`crates/blockifier/src/context.rs` line 65 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rAdJT1zVkWOzSuLLg:-O8rfA-Y9sziY2n9oZyZ:bf7rb) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/context.rs#L65)):*

implement and use `saturating_sub` on `GasAmount`.

@Yoni-Starkware this should saturate (as user can provide l2_gas amount of 1), right?
is this related to [this item](https://starkware.monday.com/boards/6987445459/pulses/6987445571)?
</blockquote></details>

Not relevant.

Note - the transaction should be reverted if the l2 gas is lower than the constant cost.
Also, note that we charge this also in VM resources at the end of the run ([here](https://github.com/starkware-libs/sequencer/blob/c094d8c52d7cff506a44d4822456df852be68d51/crates/blockifier/src/fee/receipt.rs#L75))

dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 14 unresolved discussions (waiting on @nimrod-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/context.rs` line 65 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rAdJT1zVkWOzSuLLg:-O8wvDIq7guUPHmIYWow:b-xb8w9t) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/context.rs#L65)):*

Not relevant.

Note - the transaction should be reverted if the l2 gas is lower than the constant cost.
Also, note that we charge this also in VM resources at the end of the run ([here](https://github.com/starkware-libs/sequencer/blob/c094d8c52d7cff506a44d4822456df852be68d51/crates/blockifier/src/fee/receipt.rs#L75))
</blockquote></details>

so it is a bug if this saturates? if so, this should be `checked_sub`, not saturating, but in any case this should be implemented on `GasAmount`

TzahiTaub (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 13 unresolved discussions (waiting on @dorimedini-starkware and @nimrod-starkware)

___
*[`crates/blockifier/src/context.rs` line 50 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rBhF95G2WUlQEjR6w:-O93ufRQ0IZ3pFAnpo87:bk2ew8j) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/context.rs#L50)):*
> ```XML
>     /// Returns the initial Sierra gas of the transaction.
>     /// This value is used to limit the transaction's run.
>     pub fn initial_sierra_gas(&self) -> u64 {
> ```

I prefer to do this change in the next PR, I'm afraid it will mess this one too much.
___
*[`crates/blockifier/src/context.rs` line 53 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rBCQH7FWxHh8ZLwoe:-O93xpVxADPQs_8-T0lw:b-y0h495) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/context.rs#L53)):*

non-blocking: if you combine the destructuring you wouldn't need to define the extra var
</blockquote></details>

Don't know if this is more readable but OK.
___
*[`crates/blockifier/src/context.rs` line 65 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rAdJT1zVkWOzSuLLg:-O93yKYX7ocdElm27zBk:b56tgp2) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/context.rs#L65)):*

so it is a bug if this saturates? if so, this should be `checked_sub`, not saturating, but in any case this should be implemented on `GasAmount`
</blockquote></details>

Not sure I understand what we want to happen here. Starting with an amount of gas sufficient only for the constant tx gas cost as currently done will revert the tx later on as it will have additional costs. Do you want this call to fail? And add some mechanism to have a failure here? 
Note, I think we should add this const (the tx_gas_cost to the prevalidation stage as it's not used there ATM.
___
*[`crates/blockifier/src/versioned_constants.rs` line 264 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rBkMm4NqbBw-u00YF:-O93zkjl5ffBFVfAWdNP:bje53k0) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/versioned_constants.rs#L264)):*
> ```XML
>     pub fn default_initial_gas_cost(&self) -> u64 {
>         self.os_constants.gas_costs.default_initial_gas_cost
>     }
> ```

Next PR please
___
*[`crates/blockifier/src/execution/contract_class.rs` line 71 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rBzcw8LWKHy355v7q:-O93kgI28qFWq9iuLzaL:bss4a2b) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/execution/contract_class.rs#L71)):*

I don't like this because users should `match` on the `TrackedResource` enum.
consider using `match` below, or move this to a test util area (only usage is in test)
</blockquote></details>

Removed.
___
*[`crates/blockifier/src/execution/entry_point.rs` line 153 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rCMKQ0qYfhs_6nq8b:-O93l3AaDwL3515C_mh-:bys8zdx) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/execution/entry_point.rs#L153)):*

should this be next to the check where we check if we are running steps or not?
this should also depend on "parent context", right?
</blockquote></details>

Not exactly, that check is later on (inside `execute_entry_point_call_wrapper`). This is only for (not) limiting the run when possibly running old cairo 1, it doesn't influence what resource we count in the end. And it's not tight, as if the condition doesn't hold we might still run with VM mode. I can move it into the wrapper function if it will be clearer, just need to make the `call` param in there mutable. WDYT?
___
*[`crates/blockifier/src/fee/fee_test.rs` line 378 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rD1Ti6s9nb5jYMmn5:-O941iiR-8pRXhLuDzyN:b-g1wb8b) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/fee/fee_test.rs#L378)):*

where are the magic numbers 5000, 3000 and 9 from...?
</blockquote></details>

All ,meaningless numbers Nimrod put. Is this better?
___
*[`crates/blockifier/src/transaction/transactions_test.rs` line 262 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rDOP5Bc62L6E6vhy3:-O93uZ7BBWifgrZ0PJiT:bnpi2al) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/transaction/transactions_test.rs#L262)):*

why this addition?
</blockquote></details>

No idea.

dorimedini-starkware (changes requested):
> Reviewed 5 of 5 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 3 unresolved discussions (waiting on @nimrod-starkware, @TzahiTaub, and @Yoni-Starkware)

___
*[`crates/blockifier/src/context.rs` line 65 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rAdJT1zVkWOzSuLLg:-O94ANeWA-TD882TRL1b:b-a4so6g) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/context.rs#L65)):*

Not sure I understand what we want to happen here. Starting with an amount of gas sufficient only for the constant tx gas cost as currently done will revert the tx later on as it will have additional costs. Do you want this call to fail? And add some mechanism to have a failure here? 
Note, I think we should add this const (the tx_gas_cost to the prevalidation stage as it's not used there ATM.
</blockquote></details>

Depends
1. If this is called *only in the top-level entry point*, then I would ask you add a test that providing an L2 bound that is less that `transaction_gas_cost` will fail pre-validation (reject), and then IMO you should do `checked_sub(...).expect(...)`, not `unwrap_or_default`. If the subtraction underflows, we have a bug! pre-validation should have failed.
2. If this is called for each entry point, then we can really underflow, right? And we would want to reject/revert?

In both cases I don't think we should saturate (which is equivalent to `checked_sub(..).unwrap_or_default()`)
___
*[`crates/blockifier/src/context.rs` line 73 at r10](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O94A02X4StP00JY3UmY:-O94A02X4StP00JY3UmZ:bsl1gvy) ([raw file](https://github.com/starkware-libs/sequencer/blob/5e6d09caedd895e0e225e567368966f7b44a3f3c/crates/blockifier/src/context.rs#L73)):*
> ```XML
>                 )
>                 .unwrap_or_default(),
>             TransactionInfo::Deprecated(_) => default_initial_gas,
> ```

this is what I meant (if it compiles).
if it does, delete the redundant `default_initial_gas` variable



_Suggestion:_
```XML
            TransactionInfo::Deprecated(_)
            | TransactionInfo::Current(CurrentTransactionInfo {
                resource_bounds: ValidResourceBounds::L1Gas(_),
                ..
            }) => default_initial_gas,
            TransactionInfo::Current(CurrentTransactionInfo {
                resource_bounds: ValidResourceBounds::AllResources(AllResourceBounds { l2_gas, .. }),
                ..
            }) => l2_gas
                .max_amount
                .0
                .checked_sub(
                    self.block_context
                        .versioned_constants
                        .os_constants
                        .gas_costs
                        .transaction_gas_cost,
                )
                .unwrap_or_default(),
            
```
___
*[`crates/blockifier/src/execution/entry_point.rs` line 153 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rCMKQ0qYfhs_6nq8b:-O94BQh2BFikmYEQmaTw:bjvan4i) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/execution/entry_point.rs#L153)):*

Not exactly, that check is later on (inside `execute_entry_point_call_wrapper`). This is only for (not) limiting the run when possibly running old cairo 1, it doesn't influence what resource we count in the end. And it's not tight, as if the condition doesn't hold we might still run with VM mode. I can move it into the wrapper function if it will be clearer, just need to make the `call` param in there mutable. WDYT?
</blockquote></details>

I want the `initial_gas` value to be set once. If you have to check, in two separate places, whether or not you need to set it to `default_initial_gas_cost`, then something is wrong IMO.
Is this possible?
Or also an option: will this make code *worse* somehow?
___
*[`crates/blockifier/src/fee/fee_test.rs` line 378 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rD1Ti6s9nb5jYMmn5:-O949SYJBmrdL8vmcc7n:bgawq48) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/fee/fee_test.rs#L378)):*

All ,meaningless numbers Nimrod put. Is this better?
</blockquote></details>

ah, I was just confused that the NoL2Gas case used a computed number and for the All case it's arbitrary

TzahiTaub (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/context.rs` line 73 at r10](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O94A02X4StP00JY3UmY:-O94CpDc2-tGcRvwXo8f:b-py294e) ([raw file](https://github.com/starkware-libs/sequencer/blob/5e6d09caedd895e0e225e567368966f7b44a3f3c/crates/blockifier/src/context.rs#L73)):*

this is what I meant (if it compiles).
if it does, delete the redundant `default_initial_gas` variable
</blockquote></details>

Ah, OK.
___
*[`crates/blockifier/src/execution/entry_point.rs` line 153 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rCMKQ0qYfhs_6nq8b:-O94DQZbEBzYNwgRp7hH:b-tsrlp9) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/execution/entry_point.rs#L153)):*

I want the `initial_gas` value to be set once. If you have to check, in two separate places, whether or not you need to set it to `default_initial_gas_cost`, then something is wrong IMO.
Is this possible?
Or also an option: will this make code *worse* somehow?
</blockquote></details>

It is intially set before the location we are at (on separate callers), and only overridden once, in here. Moving it forward might make the code a bit clearer, but again, the `call` param will become mutable. Check if the current change seems better, I need to fix the tests if it does.

dorimedini-starkware (changes requested):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: 11 of 13 files reviewed, 3 unresolved discussions (waiting on @nimrod-starkware, @TzahiTaub, and @Yoni-Starkware)

___
*[`crates/blockifier/src/execution/execution_utils.rs` line 71 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O94TEXq89_EMxiZ9Bkg:-O94TEXq89_EMxiZ9Bkh:b-4i4epe) ([raw file](https://github.com/starkware-libs/sequencer/blob/fe7ea1debb303e6426211bff38f5054c2395a385/crates/blockifier/src/execution/execution_utils.rs#L71)):*
> ```XML
>             context.tracked_resource_stack.push(tracked_resource)
>         }
>         _ => context.tracked_resource_stack.push(tracked_resource),
> ```

this is so the compiler will block you if you add another tracked resource variant



_Suggestion:_
```XML
None =>
```

TzahiTaub (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: 7 of 13 files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware and @nimrod-starkware)

___
*[`crates/blockifier/src/context.rs` line 65 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O8rAdJT1zVkWOzSuLLg:-O95F3Gi4qTEiVf0pwCr:b-oj46xw) ([raw file](https://github.com/starkware-libs/sequencer/blob/c927c7ff14c66d7125a29a1a99ffdf63da4f2262/crates/blockifier/src/context.rs#L65)):*

Depends
1. If this is called *only in the top-level entry point*, then I would ask you add a test that providing an L2 bound that is less that `transaction_gas_cost` will fail pre-validation (reject), and then IMO you should do `checked_sub(...).expect(...)`, not `unwrap_or_default`. If the subtraction underflows, we have a bug! pre-validation should have failed.
2. If this is called for each entry point, then we can really underflow, right? And we would want to reject/revert?

In both cases I don't think we should saturate (which is equivalent to `checked_sub(..).unwrap_or_default()`)
</blockquote></details>

Called only in top level entry point, removed this const entirely from the initial gas computation. It will be taken into account in the finalization of the fee via the VM const resources translation to L2.

dorimedini-starkware (changes requested):
> Reviewed 2 of 2 files at r12, 5 of 5 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 2 unresolved discussions (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/transaction/transactions_test.rs` line 219 at r13](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O95UO6N4tAgw4_VqmX2:-O95UO6N4tAgw4_VqmX3:bg5kztv) ([raw file](https://github.com/starkware-libs/sequencer/blob/8e00ea56998779710e38a79d4d5f46e5301fd9f6/crates/blockifier/src/transaction/transactions_test.rs#L219)):*
> ```XML
>             caller_address: ContractAddress::default(),
>             call_type: CallType::Call,
>             initial_gas: default_initial_gas_cost(),
> ```

what happened here?
if tracked resource is steps, don't you need the value from the VC?



_Code quote:_
```XML
 initial_gas: default_initial_gas_cost(),
```

dorimedini-starkware (changes requested):
> Reviewed 1 of 1 files at r14, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 1 unresolved discussion (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/transaction/transactions_test.rs` line 219 at r13](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O95UO6N4tAgw4_VqmX2:-O95XA7Q1rnIm_bvqurr:b29uch1) ([raw file](https://github.com/starkware-libs/sequencer/blob/8e00ea56998779710e38a79d4d5f46e5301fd9f6/crates/blockifier/src/transaction/transactions_test.rs#L219)):*

We get it from the VC as before, I just renamed the function as removing the tx-const from the tx_default_inital_gas func made equal to the default_inital_gas_cost.
</blockquote></details>

so, in what situation will you insert the "real" initial_gas (depending on user bounds)?
only when mutated in the execution logic?

dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 1 unresolved discussion (waiting on @nimrod-starkware)

___
*[`crates/blockifier/src/transaction/transactions_test.rs` line 219 at r13](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O95UO6N4tAgw4_VqmX2:-O95XKPr-3w09Nj1PaiC:bfy26yr) ([raw file](https://github.com/starkware-libs/sequencer/blob/8e00ea56998779710e38a79d4d5f46e5301fd9f6/crates/blockifier/src/transaction/transactions_test.rs#L219)):*

so, in what situation will you insert the "real" initial_gas (depending on user bounds)?
only when mutated in the execution logic?
</blockquote></details>

if so please doc it with a comment

TzahiTaub (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: 12 of 13 files reviewed, 1 unresolved discussion (waiting on @dorimedini-starkware)

___
*[`crates/blockifier/src/transaction/transactions_test.rs` line 219 at r13](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O95UO6N4tAgw4_VqmX2:-O95Y6ow0XQBoKWgtErh:bw91ng5) ([raw file](https://github.com/starkware-libs/sequencer/blob/8e00ea56998779710e38a79d4d5f46e5301fd9f6/crates/blockifier/src/transaction/transactions_test.rs#L219)):*

if so please doc it with a comment
</blockquote></details>

Oh, I now understand you asked about the change done a few commits ago. No current use in L2 gas in the caller tests.

Yoni-Starkware (changes requested):
> Reviewed 2 of 11 files at r6, 3 of 9 files at r7, 2 of 5 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: 12 of 13 files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware and @nimrod-starkware)

___
*[`crates/blockifier/src/context.rs` line 56 at r15](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O96RFiR4Vx9NdOgxr3y:-O96RFiR4Vx9NdOgxr3z:b7nto54) ([raw file](https://github.com/starkware-libs/sequencer/blob/a328cf1567c1869d679d407fbac737644d13f037/crates/blockifier/src/context.rs#L56)):*
> ```XML
>             TransactionInfo::Deprecated(_)
>             | TransactionInfo::Current(CurrentTransactionInfo {
>                 resource_bounds: ValidResourceBounds::L1Gas(_),
> ```

This reminds me: we need to decide how we handle txs with old resources and new sierra classes
(what you did here implies that old resources txs cannot run native - which is fine by me, just need to remember that. I.e., the first TrackedResource should be a function of the resource types )



_Code quote:_
```XML
ValidResourceBounds::L1Gas(_),
```
___
*[`crates/blockifier/src/transaction/account_transactions_test.rs` line 535 at r15](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O96SSHp2RhJinwycNQO:-O96SSHp2RhJinwycNQP:bbvxslx) ([raw file](https://github.com/starkware-libs/sequencer/blob/a328cf1567c1869d679d407fbac737644d13f037/crates/blockifier/src/transaction/account_transactions_test.rs#L535)):*
> ```XML
>         GasVectorComputationMode::NoL2Gas => assert!(error_trace.contains("no remaining steps")),
>         _ => {
>             assert!(error_trace.contains("no remaining steps") | error_trace.contains("Out of gas"))
> ```

Why `or`? there is exactly one case with mode All, how can it have two error messages?



_Code quote:_
```XML
assert!(error_trace.contains("no remaining steps") | error_trace.contains("Out of gas"))
```

Yoni-Starkware (commented):
> Reviewed 1 of 5 files at r13.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: 12 of 13 files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware and @nimrod-starkware)

TzahiTaub (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: 12 of 13 files reviewed, 4 unresolved discussions (waiting on @dorimedini-starkware, @nimrod-starkware, and @Yoni-Starkware)

*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O96V-sn3cTojdIOMjyb:-O96V-sn3cTojdIOMjyc:b-16t5oc) (no related file):*
Python side [PR](https://reviewable.io/reviews/starkware-industries/starkware/36080) (with os change)

___
___
*[`crates/blockifier/src/context.rs` line 56 at r15](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O96RFiR4Vx9NdOgxr3y:-O96TZPm5epCH_cg37DJ:b-bzl66q) ([raw file](https://github.com/starkware-libs/sequencer/blob/a328cf1567c1869d679d407fbac737644d13f037/crates/blockifier/src/context.rs#L56)):*

This reminds me: we need to decide how we handle txs with old resources and new sierra classes
(what you did here implies that old resources txs cannot run native - which is fine by me, just need to remember that. I.e., the first TrackedResource should be a function of the resource types )
</blockquote></details>

Haven't thought of this option. I guess we can either translate the L1 gas to sierra gas (probably don't want to do this), or as you said, make an L1 resource bounds tx run in VM mode in any case.
___
*[`crates/blockifier/src/transaction/account_transactions_test.rs` line 535 at r15](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O96SSHp2RhJinwycNQO:-O96UKbWEsW3YZlnCJyp:b-npd06e) ([raw file](https://github.com/starkware-libs/sequencer/blob/a328cf1567c1869d679d407fbac737644d13f037/crates/blockifier/src/transaction/account_transactions_test.rs#L535)):*

Why `or`? there is exactly one case with mode All, how can it have two error messages?
</blockquote></details>

I used the or because it can theoretically be OK to have a test with all resource bounds fail due to insufficient steps (depending on the tracked resource), didn't make it match the current test data exactly.
___
*[`crates/blockifier/src/transaction/transactions_test.rs` line 219 at r13](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O95UO6N4tAgw4_VqmX2:-O96Uoy_BrWiJq5GaTHG:b-7g6inw) ([raw file](https://github.com/starkware-libs/sequencer/blob/8e00ea56998779710e38a79d4d5f46e5301fd9f6/crates/blockifier/src/transaction/transactions_test.rs#L219)):*

Oh, I now understand you asked about the change done a few commits ago. No current use in L2 gas in the caller tests.
</blockquote></details>

Added a todo.

dorimedini-starkware (approved):
> Reviewed 1 of 1 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 3 unresolved discussions (waiting on @nimrod-starkware and @Yoni-Starkware)

Yoni-Starkware (changes requested):
> Reviewed 1 of 5 files at r13.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 5 unresolved discussions (waiting on @nimrod-starkware and @TzahiTaub)

___
*[`crates/blockifier/src/execution/execution_utils.rs` line 54 at r15](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O98csQP9YkQaoG5DVgp:-O98csQQEtnjieT-He5C:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/a328cf1567c1869d679d407fbac737644d13f037/crates/blockifier/src/execution/execution_utils.rs#L54)):*
> ```XML
>     context: &mut EntryPointExecutionContext,
> ) -> EntryPointExecutionResult<CallInfo> {
>     let tracked_resource = contract_class
> ```





_Suggestion:_
```XML
let contract_tracked_resource = contract_class
```
___
*[`crates/blockifier/src/execution/execution_utils.rs` line 66 at r15](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O98cuvCA0oV6aZ3rJ68:-O98cuvCA0oV6aZ3rJ69:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/a328cf1567c1869d679d407fbac737644d13f037/crates/blockifier/src/execution/execution_utils.rs#L66)):*
> ```XML
>         Some(TrackedResource::SierraGas) => {
>             if tracked_resource == TrackedResource::CairoSteps {
>                 // Override initial_gas with a high value so it won't limit the run.
> ```





_Suggestion:_
```XML
            if tracked_resource == TrackedResource::CairoSteps {
                // Switching from SierraGas to CairoSteps: override initial_gas with a high value so it won't limit the run.
```
___
*[`crates/blockifier/src/execution/execution_utils.rs` line 66 at r15](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O98eBjL4Ut69w7LaXFK:-O98eBjL4Ut69w7LaXFL:bir0fgx) ([raw file](https://github.com/starkware-libs/sequencer/blob/a328cf1567c1869d679d407fbac737644d13f037/crates/blockifier/src/execution/execution_utils.rs#L66)):*
> ```XML
>         Some(TrackedResource::SierraGas) => {
>             if tracked_resource == TrackedResource::CairoSteps {
>                 // Override initial_gas with a high value so it won't limit the run.
> ```

Oh boy. Add a comment
`NOTE: this value might be overridden later on.`
[Here](https://github.com/starkware-libs/sequencer/blob/3ae5b34d2cd99d49086bc45d24bea4233bf16566/crates/blockifier/src/execution/syscalls/hint_processor.rs#L859) and [here](https://github.com/starkware-libs/sequencer/blob/3ae5b34d2cd99d49086bc45d24bea4233bf16566/crates/blockifier/src/execution/syscalls/mod.rs#L199)



_Code quote:_
```XML
// Override initial_gas with a high value so it won't limit the ru
```
___
*[`crates/blockifier/src/transaction/account_transactions_test.rs` line 535 at r15](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O96SSHp2RhJinwycNQO:-O98Zz474d-K7rhN9nik:btqf3fo) ([raw file](https://github.com/starkware-libs/sequencer/blob/a328cf1567c1869d679d407fbac737644d13f037/crates/blockifier/src/transaction/account_transactions_test.rs#L535)):*

I used the or because it can theoretically be OK to have a test with all resource bounds fail due to insufficient steps (depending on the tracked resource), didn't make it match the current test data exactly.
</blockquote></details>

Please keep just the specific error for now. 
We may want to check later to see if the second one is reachable, so we should add a specific case.

TzahiTaub (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: 11 of 15 files reviewed, 5 unresolved discussions (waiting on @dorimedini-starkware, @nimrod-starkware, and @Yoni-Starkware)

___
*[`crates/blockifier/src/transaction/account_transactions_test.rs` line 535 at r15](https://reviewable.io/reviews/starkware-libs/sequencer/809#-O96SSHp2RhJinwycNQO:-O991ExE5minpLQaUC1L:behb84y) ([raw file](https://github.com/starkware-libs/sequencer/blob/a328cf1567c1869d679d407fbac737644d13f037/crates/blockifier/src/transaction/account_transactions_test.rs#L535)):*

Please keep just the specific error for now. 
We may want to check later to see if the second one is reachable, so we should add a specific case.
</blockquote></details>

OK, seems to be leftovers from changes in the gas computations I had in this PR. Current error is only of steps.

Yoni-Starkware (approved):
:lgtm:

> Reviewed 1 of 1 files at r15, 4 of 4 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809#-:-O9B37vuCWLqsTc-H8Xm:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @nimrod-starkware)

dorimedini-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: all files reviewed, 1 unresolved discussion (waiting on @nimrod-starkware)

TzahiTaub (commented):
> Reviewed 1 of 11 files at r6, 2 of 9 files at r7, 2 of 5 files at r13, 8 of 8 files at r17, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/809)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @nimrod-starkware)

PR #1053, opened 2024-09-26
fix(tests-integration) use feature account

giladchase:
Fix bug in tx generator where txs used `FeatureContract` mock addresses, rather than a consistent `sender_address` per deployed account, as derived from its deploy account tx.

Changes:
- Main part is in `starknet_api_test_utils.rs`, the rest is mostly getting things to compile, in particular, the code in end_to_end_test.rs and integration_test_setup.rs is disabled at the moment anyway.

There are two main changes (in `starknet_api_test_utils.rs`)
- Loading accounts to the generator must be done one at a time, this is for very-near-future-proofing for when we start supporting deploy account txs, where every account loaded to the tx generator will
 generate a deploy-account tx that has to accepted before invokes can
 pass validation.
 - Once accounts are loaded to the system, we use FeatureAccount instead of FeatureContract directly, which makes sure that every account is matched with exactly one sender_address, as determined by a unique deploy_account tx.

### DIFF
--- a/crates/mempool_test_utils/src/starknet_api_test_utils.rs
+++ b/crates/mempool_test_utils/src/starknet_api_test_utils.rs
@@ -1,5 +1,4 @@
 use std::cell::RefCell;
-use std::collections::HashMap;
 use std::env;
 use std::fs::File;
 use std::path::Path;
@@ -190,9 +189,9 @@ pub fn invoke_tx(cairo_version: CairoVersion) -> RpcTransaction {
 pub fn executable_invoke_tx(cairo_version: CairoVersion) -> Transaction {
     let default_account = FeatureContract::AccountWithoutValidations(cairo_version);
 
-    MultiAccountTransactionGenerator::new_for_account_contracts([default_account])
-        .account_with_id(0)
-        .generate_default_executable_invoke()
+    let mut tx_generator = MultiAccountTransactionGenerator::new();
+    tx_generator.register_account(default_account);
+    tx_generator.account_with_id(0).generate_default_executable_invoke()
 }
 
 pub fn generate_default_deploy_account_with_salt(
@@ -211,7 +210,6 @@ pub fn generate_default_deploy_account_with_salt(
 // TODO: when moving this to Starknet API crate, move this const into a module alongside
 // MultiAcconutTransactionGenerator.
 type AccountId = usize;
-type ContractInstanceId = u16;
 
 type SharedNonceManager = Rc<RefCell<NonceManager>>;
 
@@ -225,48 +223,50 @@ type SharedNonceManager = Rc<RefCell<NonceManager>>;
 /// # Example
 ///
 /// ```
+/// use blockifier::test_utils::contracts::FeatureContract;
+/// use blockifier::test_utils::CairoVersion;
 /// use mempool_test_utils::starknet_api_test_utils::MultiAccountTransactionGenerator;
 ///
-/// let mut tx_generator = MultiAccountTransactionGenerator::new(2); // Initialize with 2 accounts.
+/// let mut tx_generator = MultiAccountTransactionGenerator::new();
+/// let some_account_type = FeatureContract::AccountWithoutValidations(CairoVersion::Cairo1);
+/// // Initialize multiple accounts, these can be any account type in `FeatureContract`.
+/// tx_generator.register_account_for_flow_test(some_account_type.clone());
+/// tx_generator.register_account_for_flow_test(some_account_type);
+///
 /// let account_0_tx_with_nonce_0 = tx_generator.account_with_id(0).generate_default_invoke();
 /// let account_1_tx_with_nonce_0 = tx_generator.account_with_id(1).generate_default_invoke();
 /// let account_0_tx_with_nonce_1 = tx_generator.account_with_id(0).generate_default_invoke();
 /// ```
 // Note: when moving this to starknet api crate, see if blockifier's
 // [blockifier::transaction::test_utils::FaultyAccountTxCreatorArgs] can be made to use this.
+#[derive(Default)]
 pub struct MultiAccountTransactionGenerator {
     // Invariant: coupled with the nonce manager.
     account_tx_generators: Vec<AccountTransactionGenerator>,
     // Invariant: nonces managed internally thorugh `generate` API of the account transaction
     // generator.
-    // Only used by single account transaction generators, but owning it here is preferable over
-    // only distributing the ownership among the account generators.
-    _nonce_manager: SharedNonceManager,
+    nonce_manager: SharedNonceManager,
 }
 
 impl MultiAccountTransactionGenerator {
-    pub fn new(n_accounts: usize) -> Self {
-        let default_account_contract =
-            FeatureContract::AccountWithoutValidations(CairoVersion::Cairo1);
-        let accounts = std::iter::repeat(default_account_contract).take(n_accounts);
-        Self::new_for_account_contracts(accounts)
+    pub fn new() -> Self {
+        Self::default()
     }
 
-    pub fn new_for_account_contracts(accounts: impl IntoIterator<Item = FeatureContract>) -> Self {
-        let mut account_tx_generators = vec![];
-        let mut account_type_to_n_instances = HashMap::new();
-        let nonce_manager = SharedNonceManager::default();
-        for account in accounts {
-            let n_current_contract = account_type_to_n_instances.entry(account).or_insert(0);
-            account_tx_generators.push(AccountTransactionGenerator {
-                account,
-                contract_instance_id: *n_current_contract,
-                nonce_manager: nonce_manager.clone(),
-            });
-            *n_current_contract += 1;
-        }
+    pub fn register_account(
+        &mut self,
+        account_contract: FeatureContract,
+    ) -> (FeatureAccount, RpcTransaction) {
+        let new_account_id = self.account_tx_generators.len();
+        let (account_tx_generator, default_deploy_account_tx) = AccountTransactionGenerator::new(
+            new_account_id,
+            account_contract,
+            self.nonce_manager.clone(),
+        );
+        self.account_tx_generators.push(account_tx_generator);
+        let account_tx_generator = self.account_with_id(new_account_id);
 
-        Self { account_tx_generators, _nonce_manager: nonce_manager }
+        (account_tx_generator.account.clone(), default_deploy_account_tx)
     }
 
     pub fn account_with_id(&mut self, account_id: AccountId) -> &mut AccountTransactionGenerator {
@@ -277,6 +277,16 @@ impl MultiAccountTransactionGenerator {
             )
         })
     }
+
+    // TODO(deploy_account_support): once we support deploy account in tests, remove this method and
+    // only use new_account_default in tests. In particular, deploy account txs must be then sent to
+    // the GW via the add tx endpoint just like other txs.
+    pub fn register_account_for_flow_test(
+        &mut self,
+        account_contract: FeatureContract,
+    ) -> FeatureAccount {
+        self.register_account(account_contract).0
+    }
 }
 
 /// Manages transaction generation for a single account.
@@ -286,9 +296,9 @@ impl MultiAccountTransactionGenerator {
 /// with room for future extensions.
 ///
 /// TODO: add more transaction generation methods as needed.
+#[derive(Debug)]
 pub struct AccountTransactionGenerator {
-    account: FeatureContract,
-    contract_instance_id: ContractInstanceId,
+    account: FeatureAccount,
     nonce_manager: SharedNonceManager,
 }
 
@@ -299,7 +309,7 @@ impl AccountTransactionGenerator {
             sender_address: self.sender_address(),
             resource_bounds: test_resource_bounds_mapping(),
             nonce: self.next_nonce(),
-            calldata: create_trivial_calldata(self.test_contract_address()),
+            calldata: create_trivial_calldata(self.sender_address()),
         );
         rpc_invoke_tx(invoke_args)
     }
@@ -309,16 +319,10 @@ impl AccountTransactionGenerator {
             sender_address: self.sender_address(),
             resource_bounds: test_valid_resource_bounds(),
             nonce: self.next_nonce(),
-            calldata: create_trivial_calldata(self.test_contract_address()),
+            calldata: create_trivial_calldata(self.sender_address()),
         );
 
         Transaction::Invoke(starknet_api::test_utils::invoke::executable_invoke_tx(invoke_args))
-    }
-
-    // TODO: support more contracts, instead of this hardcoded type.
-    pub fn test_contract_address(&mut self) -> ContractAddress {
-        let cairo_version = self.account.cairo_version();
-        FeatureContract::TestContract(cairo_version).get_instance_address(0)
     }
 
     /// Generates an `RpcTransaction` with fully custom parameters.
@@ -335,13 +339,35 @@ impl AccountTransactionGenerator {
     }
 
     pub fn sender_address(&mut self) -> ContractAddress {
-        self.account.get_instance_address(self.contract_instance_id)
+        self.account.sender_address
     }
 
     /// Retrieves the nonce for the current account, and __increments__ it internally.
     pub fn next_nonce(&mut self) -> Nonce {
         let sender_address = self.sender_address();
         self.nonce_manager.borrow_mut().next(sender_address)
+    }
+
+    /// Private constructor, since only the multi-account transaction generator should create this
+    /// struct.
+    // TODO: add a version that doesn't rely on the default deploy account constructor, but takes
+    // deploy account args.
+    fn new(
+        account_id: usize,
+        account: FeatureContract,
+        nonce_manager: SharedNonceManager,
+    ) -> (Self, RpcTransaction) {
+        let contract_address_salt = ContractAddressSalt(account_id.into());
+        // A DeployAccount tx must be created now in order to affix an address to it.
+        // If this doesn't happen now it'll be difficult to fund the account during test setup.
+        let default_deploy_account_tx =
+            generate_default_deploy_account(&account, contract_address_salt);
+
+        let tx_generator = Self {
+            account: FeatureAccount::new(account, &default_deploy_account_tx),
+            nonce_manager,
+        };
+        (tx_generator, default_deploy_account_tx)
     }
 }
 
@@ -361,7 +387,7 @@ impl FeatureAccount {
         self.account.get_class_hash()
     }
 
-    fn _new(account: FeatureContract, deploy_account_tx: &RpcTransaction) -> Self {
+    fn new(account: FeatureContract, deploy_account_tx: &RpcTransaction) -> Self {
         assert_matches!(
             deploy_account_tx,
             RpcTransaction::DeployAccount(_),
--- a/crates/starknet_api/src/test_utils.rs
+++ b/crates/starknet_api/src/test_utils.rs
@@ -11,7 +11,7 @@ pub mod declare;
 pub mod deploy_account;
 pub mod invoke;
 
-#[derive(Default)]
+#[derive(Debug, Default)]
 pub struct NonceManager {
     next_nonce: HashMap<ContractAddress, Felt>,
 }
--- a/crates/tests-integration/src/integration_test_setup.rs
+++ b/crates/tests-integration/src/integration_test_setup.rs
@@ -1,7 +1,6 @@
 use std::net::SocketAddr;
 
-use blockifier::test_utils::contracts::FeatureContract;
-use blockifier::test_utils::CairoVersion;
+use mempool_test_utils::starknet_api_test_utils::FeatureAccount;
 use starknet_api::executable_transaction::Transaction;
 use starknet_api::rpc_transaction::RpcTransaction;
 use starknet_api::transaction::TransactionHash;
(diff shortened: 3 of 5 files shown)


giladchase (commented):
Moving unresolved comments from [prev version](https://reviewable.io/reviews/starkware-libs/sequencer/1046#-).

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-:-O7jCdd703mQHGp0opEJ:b-agy9i1)* status: 0 of 5 files reviewed, 2 unresolved discussions

___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 193 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7jCkdFBVO3RQMtESP0:-O7jCkdFBVO3RQMtESP1:b58xu42) ([raw file](https://github.com/starkware-libs/sequencer/blob/9a301fe4be02b740ce2f77aee22ae27838f3debd/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L193)):*
>What about `new_with_account[_contract]s`?

That won't fit the flow I have in mind (which is kind of hard to spot at this moment :sweat_smile:).
The idea is to register multiple accounts in order, where every registration generates a deploy account tx: it is necessary to do so since every account must be tied to some sender_address that is calced from its deploy account tx.
Generating them in bulk will then have to output a vector of deploy_account_txs, which might be somewhat inconvenient to send to the GW afterwards, but maybe not. WDYT? this is how I expect tests to look once I'm done with these PRs:

```rust
fn test() {
// Setup
...
    let mut tx_generator = MultiAccountTransactionGenerator::new();
    let deploy_account_tx_0 = tx_generator.register_account(FeatureAccount::SomeAccountType);
    let deploy_account_tx_1 = tx_generator.register_account(FeatureAccount::SomeAccountType2);
    let deploy_account_tx_2 = ...
...

// Test

// First lets deploy one (or more) of the accounts and add to block.
assert_eq!(add_tx(deploy_account_2), Ok());  

batcher_get_txs(1);
batcher_close_block_now();

// Now invokes from the the deployed accounts can pass validation.
let account_2_invoke_0 = tx_generator.with_account(2).generate_invoke();
let account_2_invoke_1 = tx_generator.with_account(2).generate_invoke();
assert_eq!(add_tx(account_2_invoke_0), Ok())
assert_eq!(add_tx(account_2_invoke_1), Ok())

// Now we can keep generating txs from account 2, and/or add them to blocks. 
// Or we can now deploy some other account if we want in the same way:
assert_eq!(add_tx(deploy_account_0), Ok());  
// Note 1: after deploy_account_0 is added to a block we can generate txs for it just like with account to only using `with_account(0)`.
// Note 2: whether or not we deploy account 0, we can of course still continue generating invokes for account 2.
```




_Code quote:_
```XML
    tx_generator.register_account(default_account);
```
___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 364 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7jHhBd-y_ak1Ib-vTx:-O7jHhBd-y_ak1Ib-vTy:bywma5i) ([raw file](https://github.com/starkware-libs/sequencer/blob/066eeaf87bcb642e535316db8e2cc4e048069c09/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L364)):*
>Is that the default in `deploy_account_tx_args!()` macro?

Nope, default is a bad name here and in `generate_default_invoke`: most of the fields are indeed the defaults as in the macro, but a few are just the "default choice" when using the tx generator to ensure consistency: for example they all use the same sender_address, and in the invoke the invoke is always some trivial invoke from FeatureContract::TestContract

How about I'll rename both here and in the invoke one `default->test`?



_Code quote:_
```XML
        let default_deploy_account_tx =
            generate_default_deploy_account_with_salt(&account, contract_address_salt);
```

elintul (changes requested):
> Reviewed 4 of 5 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1053)* status: all files reviewed, 12 unresolved discussions (waiting on @giladchase and @Yael-Starkware)

___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 193 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7jCkdFBVO3RQMtESP0:-O7rkQdL-t-UfSyx89Rq:b-vs1prh) ([raw file](https://github.com/starkware-libs/sequencer/blob/9a301fe4be02b740ce2f77aee22ae27838f3debd/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L193)):*

>What about `new_with_account[_contract]s`?

That won't fit the flow I have in mind (which is kind of hard to spot at this moment :sweat_smile:).
The idea is to register multiple accounts in order, where every registration generates a deploy account tx: it is necessary to do so since every account must be tied to some sender_address that is calced from its deploy account tx.
Generating them in bulk will then have to output a vector of deploy_account_txs, which might be somewhat inconvenient to send to the GW afterwards, but maybe not. WDYT? this is how I expect tests to look once I'm done with these PRs:

```rust
fn test() {
// Setup
...
    let mut tx_generator = MultiAccountTransactionGenerator::new();
    let deploy_account_tx_0 = tx_generator.register_account(FeatureAccount::SomeAccountType);
    let deploy_account_tx_1 = tx_generator.register_account(FeatureAccount::SomeAccountType2);
    let deploy_account_tx_2 = ...
...

// Test

// First lets deploy one (or more) of the accounts and add to block.
assert_eq!(add_tx(deploy_account_2), Ok());  

batcher_get_txs(1);
batcher_close_block_now();

// Now invokes from the the deployed accounts can pass validation.
let account_2_invoke_0 = tx_generator.with_account(2).generate_invoke();
let account_2_invoke_1 = tx_generator.with_account(2).generate_invoke();
assert_eq!(add_tx(account_2_invoke_0), Ok())
assert_eq!(add_tx(account_2_invoke_1), Ok())

// Now we can keep generating txs from account 2, and/or add them to blocks. 
// Or we can now deploy some other account if we want in the same way:
assert_eq!(add_tx(deploy_account_0), Ok());  
// Note 1: after deploy_account_0 is added to a block we can generate txs for it just like with account to only using `with_account(0)`.
// Note 2: whether or not we deploy account 0, we can of course still continue generating invokes for account 2.
```

</blockquote></details>

Looks good!
___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 364 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7jHhBd-y_ak1Ib-vTx:-O7rkMloAmQEMoW1Vebq:bpd6ccq) ([raw file](https://github.com/starkware-libs/sequencer/blob/066eeaf87bcb642e535316db8e2cc4e048069c09/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L364)):*

>Is that the default in `deploy_account_tx_args!()` macro?

Nope, default is a bad name here and in `generate_default_invoke`: most of the fields are indeed the defaults as in the macro, but a few are just the "default choice" when using the tx generator to ensure consistency: for example they all use the same sender_address, and in the invoke the invoke is always some trivial invoke from FeatureContract::TestContract

How about I'll rename both here and in the invoke one `default->test`?
</blockquote></details>

Yup, later. :)
___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 216 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rlvJY2bBD1lnMH7qe:-O7rlvJY2bBD1lnMH7qf:b-8nqg33) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L216)):*
> ```XML
> type SharedNonceManager = Rc<RefCell<NonceManager>>;
> 
> /// Manages transaction generation for multiple pre-funded accounts, internally bumping nonces
> ```

Does this code run in tests?
___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 267 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rkdWj2ckXjnoyCSHx:-O7rkdWj2ckXjnoyCSHy:btx0c0k) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L267)):*
> ```XML
>         );
>         self.account_tx_generators.push(account_tx_generator);
>         let account_tx_generator = self.account_with_id(new_account_id);
> ```

Why not continue with the one you've just created? Is this a sanity check?



_Code quote:_
```XML
let account_tx_generator = self.account_with_id(new_account_id);
```
___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 269 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rkkVJ4fS1wXVG4Ho_:-O7rkkVJ4fS1wXVG4Hoa:brmuvud) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L269)):*
> ```XML
>         let account_tx_generator = self.account_with_id(new_account_id);
> 
>         (account_tx_generator.account.clone(), default_deploy_account_tx)
> ```

Cheap?



_Code quote:_
```XML
clone()
```
___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 362 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rj5nzA_UthMPeiZRX:-O7rj5nzA_UthMPeiZRY:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L362)):*
> ```XML
>         let contract_address_salt = ContractAddressSalt(account_id.into());
>         // A DeployAccount tx must be created now in order to affix an address to it.
>         // If this doesn't happen now it'll be difficult to fund the account during test setup.
> ```





_Suggestion:_
```XML
        // A deploy account transaction must be created now in order to affix an address to it.
        // If this doesn't happen now it'll be difficult to fund the account during test setup.
        let contract_address_salt = ContractAddressSalt(account_id.into());
```
___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 370 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rjp3O8pbQnTPYtiHX:-O7rjp3O8pbQnTPYtiHY:b6y2pp5) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L370)):*
> ```XML
>             nonce_manager,
>         };
>         // Bump the nonce to account for the creation of the deploy account tx.
> ```

A bit fragile; can we somehow tie this to happen right after transaction creation (kind of `__exit__` of a context manager)? Not here though.



_Suggestion:_
```XML
// Bump the account nonce after transaction creation.
```
___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 382 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rkysa8n_BqKmzs7o0:-O7rkysa8n_BqKmzs7o1:b-suv37h) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L382)):*
> ```XML
> // not related to an actual deploy account transaction, which is the way real account addresses are
> // calculated.
> #[derive(Debug, Clone)]
> ```

Alphabetize?



_Code quote:_
```XML
#[derive(Debug, Clone)]
```
___
*[`crates/tests-integration/tests/end_to_end_test.rs` line 13 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rlFsjAbthSWUgpL9A:-O7rlFsjAbthSWUgpL9B:b2t1t5q) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/tests-integration/tests/end_to_end_test.rs#L13)):*
> ```XML
> }
> 
> #[rstest]
> ```

Needed in order to be able to use fixtures?



_Code quote:_
```XML
#[rstest]
```
___
*[`crates/tests-integration/tests/end_to_end_test.rs` line 19 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rlZU_58s7FxF0LgwA:-O7rlZU_58s7FxF0LgwB:b-x60b41) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/tests-integration/tests/end_to_end_test.rs#L19)):*
> ```XML
> async fn test_end_to_end(mut tx_generator: MultiAccountTransactionGenerator) {
>     // Setup.
>     let accounts: Vec<_> = [
> ```

Would actually prefer to see what's generated here, for readability.



_Code quote:_
```XML
<_>
```
___
*[`crates/tests-integration/tests/end_to_end_test.rs` line 25 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rlQOR4m_ET5NT3fRY:-O7rlQOS59qrFc4SFoHl:b-vyq6t2) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/tests-integration/tests/end_to_end_test.rs#L25)):*
> ```XML
>     .into_iter()
>     .map(|account| tx_generator.register_account_for_flow_test(account))
>     .collect();
> ```

Is this needed, if you're passing a slice to `new`? Maybe pass iterable?



_Code quote:_
```XML
.collect()
```
___
*[`crates/tests-integration/tests/end_to_end_test.rs` line 35 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rli6Q-2ODS8DtKBsP:-O7rli6Q-2ODS8DtKBsQ:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/tests-integration/tests/end_to_end_test.rs#L35)):*
> ```XML
>     // Test.
> 
>     mock_running_system.assert_add_tx_success(&account0_invoke_nonce1).await;
> ```





_Suggestion:_
```XML
    // Test.
    mock_running_system.assert_add_tx_success(&account0_invoke_nonce1).await;
```

elintul (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1053)* status: all files reviewed, 12 unresolved discussions (waiting on @giladchase and @Yael-Starkware)

___
*[`crates/tests-integration/tests/end_to_end_test.rs` line 19 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rlZU_58s7FxF0LgwA:-O7s1O8K7_uh5AK4b-Z6:bb2auw9) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/tests-integration/tests/end_to_end_test.rs#L19)):*

Would actually prefer to see what's generated here, for readability.
</blockquote></details>

Actually, can we avoid this annotation altogether, given the compiler can deduce the type from `new_for_accounts`'s signature?

elintul (approved):
:lgtm:

> Reviewed 3 of 3 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-:-O7xMrP86jc7Ec73KTBM:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on @giladchase and @Yael-Starkware)

___
*[`crates/tests-integration/tests/end_to_end_test.rs` line 22 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7xMad64O2JRFkTOf8-:-O7xMad64O2JRFkTOf80:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/758d403bbe7aa50968d728aa4ce73200440c2073/crates/tests-integration/tests/end_to_end_test.rs#L22)):*
> ```XML
> async fn test_end_to_end(mut tx_generator: MultiAccountTransactionGenerator) {
>     // Setup.
>     for contract in [
> ```





_Suggestion:_
```XML
account
```

elintul (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1053)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase and @Yael-Starkware)

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1053)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Yael-Starkware)

___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 216 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rlvJY2bBD1lnMH7qe:-O7wvo382LmhZz_6MtcS:bnsqelc) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L216)):*

Does this code run in tests?
</blockquote></details>

Yep, to run it specifically you need to do `cargo test --doc`, since these are called "doc tests".
___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 267 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rkdWj2ckXjnoyCSHx:-O7ww66pE9tMY8eLLRBH:bd575r0) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L267)):*

Why not continue with the one you've just created? Is this a sanity check?
</blockquote></details>

Ya that bugged me to, but the preceding line that pushes it into self.account_tx_generators requires ownership (because it's pushing it into a Vec).

I can solve this by adding a setter as a convenience method:
```rust
fn add_account_tx_generator(&mut self, account_id, account_tx_generator) -> &mut AccountTxGenerator {
        self.account_tx_generators.push(account_tx_generator);
        // Can also replace below with self.account_with_id(self.account_tx_generators.len()-1), and then we won't need 
        // account_id arg, but it's a bit unclear.
        self.account_with_id(account_id) 
}
```
Then we could replace the two lines with:
```rust
let account_tx_generator = self.add_account_tx_generator(new_account_id, account_tx_generator);
```
WDYT?
___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 269 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rkkVJ4fS1wXVG4Ho_:-O7wy2ChDuKrFaQ8xpqR:b-qvgf1o) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L269)):*

Cheap?
</blockquote></details>

Nope, thanks for catching.
Discussed offline and changed accordingly.
___
*[`crates/mempool_test_utils/src/starknet_api_test_utils.rs` line 370 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rjp3O8pbQnTPYtiHX:-O7xHOToADcTcr46AdfE:b-ldgmll) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/mempool_test_utils/src/starknet_api_test_utils.rs#L370)):*
Ya i don't like this either, but putting it in a private `new` is the best option I found :cry:  I'll try to think of something.

>after transaction creation

We don't have the address yet :( I can calc it, but that is pretty much the reason `FeatureAccount::new` exists, to calc that and save the result, and then `account_tx_generator.next_nonce()` knows where to grab the address from and so what nonce to bump.
Basically I think the abstraction that ties this address and the nonce together is the account_tx_generator itself :sweat_smile:.

Another option would be to pass nonce_manager as an arg into `generate_default_deploy_account_with_salt` and it'll handle nonce bumping, but that func shouldn't really care about nonce management I think.

Another option which I don't like is to add a generate_deploy_account_tx static method to AcccountTransactionGenerator, which will get the argss for the free function `generate_default_deploy_account_with_salt` and the nonce_manager, then it'll call the free function and bump the nonce.
But I don't like static functions :smiling_face_with_tear: 
___
*[`crates/tests-integration/tests/end_to_end_test.rs` line 13 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rlFsjAbthSWUgpL9A:-O7xJNN04JciW_J1UYub:b2nign3) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/tests-integration/tests/end_to_end_test.rs#L13)):*

Needed in order to be able to use fixtures?
</blockquote></details>

yep
___
*[`crates/tests-integration/tests/end_to_end_test.rs` line 19 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rlZU_58s7FxF0LgwA:-O7xJSNoCk5qGIH7L5xa:b59s4gj) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/tests-integration/tests/end_to_end_test.rs#L19)):*

Actually, can we avoid this annotation altogether, given the compiler can deduce the type from `new_for_accounts`'s signature?
</blockquote></details>

Stale, as discussed offline, we're getting rid of this var.
___
*[`crates/tests-integration/tests/end_to_end_test.rs` line 25 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/1053#-O7rlQOR4m_ET5NT3fRY:-O7xJf6N36Wd2WlqXC7v:b59s4gj) ([raw file](https://github.com/starkware-libs/sequencer/blob/c9e702a099f6a67b81bb6613f53411635d31e085/crates/tests-integration/tests/end_to_end_test.rs#L25)):*

Is this needed, if you're passing a slice to `new`? Maybe pass iterable?
</blockquote></details>

Stale, as discussed offline, we're getting rid of this var.

elintul (approved):
> Reviewed 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1053)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Yael-Starkware)

PR #1242, opened 2024-10-08
feat(blockifier): compile cairo native runtime

rodrigo-pino:
Add the Cairo Native runtime binary to the root after executing `./scripts/dependencies.sh`

- - -
This change is [Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,4 +1,5 @@
 [env]
+CAIRO_NATIVE_RUNTIME_LIBRARY = "./libcairo_native_runtime.so"
 LLVM_SYS_191_PREFIX = "/usr/lib/llvm-19/"
 MLIR_SYS_190_PREFIX = "/usr/lib/llvm-19/"
 TABLEGEN_190_PREFIX = "/usr/lib/llvm-19/"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,5 +1,5 @@
 [env]
-CAIRO_NATIVE_RUNTIME_LIBRARY = "./libcairo_native_runtime.so"
+CAIRO_NATIVE_RUNTIME_LIBRARY = "./libcairo_native_runtime.a"
 LLVM_SYS_191_PREFIX = "/usr/lib/llvm-19/"
 MLIR_SYS_190_PREFIX = "/usr/lib/llvm-19/"
 TABLEGEN_190_PREFIX = "/usr/lib/llvm-19/"
--- a/.github/workflows/blockifier_ci.yml
+++ b/.github/workflows/blockifier_ci.yml
@@ -44,6 +44,10 @@ jobs:
     runs-on: starkware-ubuntu-20-04-medium
     steps:
       - uses: actions/checkout@v4
+        with:
+          # required to clone native as a gitsubmodule
+          submodules: recursive
+          fetch-depth: 0
       - uses: ./.github/actions/bootstrap
       # No features - build blockifier without features activated by dependencies in the workspace.
       - run: cargo build -p blockifier
@@ -51,6 +55,6 @@ jobs:
       # transaction_serde is not activated by any workspace crate; test the build.
       - run: cargo build -p blockifier --features transaction_serde
       - run: cargo test -p blockifier --features transaction_serde
-        # cairo_native is not activated by any workspace crate; test the build.
+      # cairo_native is not activated by any workspace crate; test the build.
       - run: cargo build -p blockifier --features cairo_native
       - run: cargo test -p blockifier --features cairo_native
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,6 +31,8 @@ jobs:
       # Environment setup.
       - uses: actions/checkout@v4
         with:
+          # required to clone native as a git submodule
+          submodules: recursive
           # Fetch the entire history. Required to checkout the merge target commit, so the diff can
           # be computed.
           fetch-depth: 0
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,6 @@ scripts/papyrus/p2p_sync_e2e_test/data_server/
 # Papyrus helm chart
 deployments/papyrus/helm/config/*
 !deployments/papyrus/helm/config/example.json
+
+# Generated file used for running contracts compiled with Cairo Native
+libcairo_native_runtime.so
--- a/.gitignore
+++ b/.gitignore
@@ -39,4 +39,4 @@ deployments/papyrus/helm/config/*
 !deployments/papyrus/helm/config/example.json
 
 # Generated file used for running contracts compiled with Cairo Native
-libcairo_native_runtime.so
+crates/blockifier/libcairo_native_runtime.a
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "crates/blockifier/cairo_native"]
+    path = crates/blockifier/cairo_native
+    url = https://github.com/lambdaclass/cairo_native
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -96,6 +96,7 @@ cairo-lang-sierra = "=2.8.4"
 cairo-lang-sierra-to-casm = "2.8.4"
 cairo-lang-starknet-classes = "2.8.4"
 cairo-lang-utils = "2.8.4"
+# Important: when updated, make sure to update the cairo-native submodule as well.
 cairo-native = "0.2.0-alpha.4"
 cairo-vm = "=1.0.1"
 camelpaste = "0.1.0"
--- a/crates/blockifier/build.rs
+++ b/crates/blockifier/build.rs
@@ -1,0 +1,68 @@
+use std::path::PathBuf;
+use std::process::Command;
+
+fn compile_cairo_native_aot_runtime() {
+    let cairo_native_dir = std::env::current_dir()
+        .expect("Failed to get current directory")
+        .join(PathBuf::from("cairo_native"));
+
+    if !cairo_native_dir.exists() || !cairo_native_dir.join(".git").exists() {
+        panic!(
+            "It seems git submodule at {} doesn't exist or it is not initialized, please \
+             run:\n\ngit submodule update --init --recursive\n",
+            cairo_native_dir.to_str().unwrap()
+        );
+    }
+
+    let runtime_target_dir = cairo_native_dir.join(PathBuf::from("target"));
+    let status = Command::new("cargo")
+        .args([
+            "build",
+            "--release",
+            "-p",
+            "cairo-native-runtime",
+            "--message-format=json",
+            "--target-dir",
+            runtime_target_dir.to_str().unwrap(),
+        ])
+        .current_dir(cairo_native_dir)
+        .status()
+        .expect("Failed to execute cargo");
+    if !status.success() {
+        panic!("Building cairo native runtime failed: {status}")
+    }
+
+    let runtime_target_path =
+        runtime_target_dir.join(PathBuf::from("release/libcairo_native_runtime.a"));
+
+    const RUNTIME_LIBRARY: &str = "CAIRO_NATIVE_RUNTIME_LIBRARY";
+    let runtime_expected_path = {
+        let expected_path_env =
+            std::env::var(RUNTIME_LIBRARY).expect("Cairo Native rutime path variable is not set");
+        let expected_path = PathBuf::from(&expected_path_env);
+
+        if expected_path.is_absolute() {
+            expected_path
+        } else {
+            std::env::current_dir().expect("Failed to get current directory").join(expected_path)
+        }
+    };
+
+    std::fs::copy(&runtime_target_path, &runtime_expected_path)
+        .expect("Failed to copy native runtime");
+
+    println!("cargo::rerun-if-changed=./cairo_native/runtime/");
+    // todo(rodrigo): this directive seems to cause the build script to trigger everytime on
+    // Linux based machines. Investigate the issue further.
+    println!("cargo::rerun-if-changed={}", runtime_expected_path.to_str().unwrap());
+    println!("cargo::rerun-if-env-changed={RUNTIME_LIBRARY}");
+}
+
+fn main() {
+    // `CARGO_FEATURE_CAIRO_NATIVE` env var is set by Cargo when compiling with the `cairo_native`
+    // feature flag. Build instructions are defined behind this condition since they are only
+    // relevant when using Cairo Native.
+    if std::env::var("CARGO_FEATURE_CAIRO_NATIVE").is_ok() {
+        compile_cairo_native_aot_runtime();
+    }
+}
--- a/scripts/dependencies.sh
+++ b/scripts/dependencies.sh
@@ -66,7 +66,7 @@ function compile_cairo_native_runtime() {
     fi
 
     # Then we clone and build the runtime from the repo
-    git clone https://github.com/lambdaclass/cairo_native.git
+    $SUDO git clone https://github.com/lambdaclass/cairo_native.git
     pushd ./cairo_native || exit 1
     git switch v0.2.0-alpha.2 --detach
     cargo build -p cairo-native-runtime --release --all-features --quiet
--- a/scripts/dependencies.sh
+++ b/scripts/dependencies.sh
@@ -48,41 +48,10 @@ function setup_llvm_deps() {
     esac
 }
 
-function compile_cairo_native_runtime() {
-    # First we need to make sure Cargo exists
-    if command -v cargo >/dev/null 2>&1; then
-        echo "Rust is already installed with cargo available in PATH."
-    else
-        echo "cargo not found. Installing Rust..."
-        if ! curl -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path; then
-            echo >&2 "Failed to install Rust. Aborting."
-            return 1
-        fi
-        # shellcheck disable=SC1090
-        source "$HOME/.cargo/env" || {
-            echo >&2 "Failed to source Rust environment. Aborting."
-            return 1
-        }
-    fi
-
-    # Then we clone and build the runtime from the repo
-    git clone https://github.com/lambdaclass/cairo_native.git
-    pushd ./cairo_native || exit 1
-    git switch v0.2.0-alpha.2 --detach
-    cargo build -p cairo-native-runtime --release --all-features --quiet
-    popd || exit 1
-
-    mv ./cairo_native/target/release/libcairo_native_runtime.a ./libcairo_native_runtime.a
-    rm -rf ./cairo_native
-}
-
 function main() {
     [ "$(uname)" = "Linux" ] && install_essential_deps_linux
     setup_llvm_deps
     echo "LLVM dependencies installed successfully."
-
-    compile_cairo_native_runtime
-    echo "Cairo Native runtime compiled successfully."
 }
 
 main "$@"
--- a/scripts/dependencies.sh
+++ b/scripts/dependencies.sh
@@ -48,10 +48,24 @@ function setup_llvm_deps() {
     esac
 }
 
+function compile_cairo_native_runtime() {
+    git clone https://github.com/lambdaclass/cairo_native.git
+    pushd ./cairo_native || exit 1
+    git switch v0.2.0-alpha.2 --detach
+    cargo build -p cairo-native-runtime --release --all-features --quiet
+    popd || exit 1
+
+    mv ./cairo_native/target/release/libcairo_native_runtime.a ./libcairo_native_runtime.so
+    rm -rf ./cairo_native
+}
+
 function main() {
     [ "$(uname)" = "Linux" ] && install_essential_deps_linux
     setup_llvm_deps
     echo "LLVM dependencies installed successfully."
+
+    compile_cairo_native_runtime
+    echo "Cairo Native runtime compiled successfully."
 }
 
 main "$@"
--- a/scripts/dependencies.sh
+++ b/scripts/dependencies.sh
@@ -66,7 +66,7 @@ function compile_cairo_native_runtime() {
     fi
 
     # Then we clone and build the runtime from the repo
-    $SUDO git clone https://github.com/lambdaclass/cairo_native.git
+    git clone https://github.com/lambdaclass/cairo_native.git
     pushd ./cairo_native || exit 1
     git switch v0.2.0-alpha.2 --detach
     cargo build -p cairo-native-runtime --release --all-features --quiet
--- a/scripts/dependencies.sh
+++ b/scripts/dependencies.sh
@@ -49,6 +49,23 @@ function setup_llvm_deps() {
 }
 
 function compile_cairo_native_runtime() {
+    # First we need to make sure Cargo exists
+    if command -v cargo >/dev/null 2>&1; then
+        echo "Rust is already installed with cargo available in PATH."
+    else
+        echo "cargo not found. Installing Rust..."
+        if ! curl -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path; then
+            echo >&2 "Failed to install Rust. Aborting."
+            return 1
+        fi
+        # shellcheck disable=SC1090
+        source "$HOME/.cargo/env" || {
+            echo >&2 "Failed to source Rust environment. Aborting."
+            return 1
+        }
+    fi
+
+    # Then we clone and build the runtime from the repo
     git clone https://github.com/lambdaclass/cairo_native.git
     pushd ./cairo_native || exit 1
     git switch v0.2.0-alpha.2 --detach
--- a/scripts/dependencies.sh
+++ b/scripts/dependencies.sh
@@ -72,7 +72,7 @@ function compile_cairo_native_runtime() {
     cargo build -p cairo-native-runtime --release --all-features --quiet
     popd || exit 1
 
-    mv ./cairo_native/target/release/libcairo_native_runtime.a ./libcairo_native_runtime.so
+    mv ./cairo_native/target/release/libcairo_native_runtime.a ./libcairo_native_runtime.a
     rm -rf ./cairo_native
 }
 
--- a/scripts/sequencer-ci.Dockerfile
+++ b/scripts/sequencer-ci.Dockerfile
@@ -12,8 +12,6 @@ RUN echo "%${USERNAME}        ALL=(ALL)       NOPASSWD: ALL" >> /etc/sudoers.d/d
 
 USER ${USERNAME}
 
-WORKDIR /app
-
 ENV RUSTUP_HOME=/var/tmp/rust
 ENV CARGO_HOME=${RUSTUP_HOME}
 ENV PATH=$PATH:${RUSTUP_HOME}/bin
--- a/scripts/sequencer-ci.Dockerfile
+++ b/scripts/sequencer-ci.Dockerfile
@@ -12,6 +12,8 @@ RUN echo "%${USERNAME}        ALL=(ALL)       NOPASSWD: ALL" >> /etc/sudoers.d/d
 
 USER ${USERNAME}
 
+WORKDIR /app
+
 ENV RUSTUP_HOME=/var/tmp/rust
 ENV CARGO_HOME=${RUSTUP_HOME}
 ENV PATH=$PATH:${RUSTUP_HOME}/bin
--- a/taplo.toml
+++ b/taplo.toml
@@ -1,4 +1,7 @@
-exclude = ["crates/native_blockifier/.cargo/config.toml"]
+exclude = [
+  "crates/blockifier/cairo_native/**/*.toml",
+  "crates/native_blockifier/.cargo/config.toml",
+]
 
 [formatting]
 column_width = 100


avi-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @rodrigo-pino)

___
*[`scripts/dependencies.sh` line 73 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O967xDQ7lSapoJvdnqD:-O967xDQ7lSapoJvdnqE:boad6l4) ([raw file](https://github.com/starkware-libs/sequencer/blob/f46f92715844eb3cda9e5e6b10b253ca55ca4c91/scripts/dependencies.sh#L73)):*
> ```Shell
>     popd || exit 1
> 
>     mv ./cairo_native/target/release/libcairo_native_runtime.a ./libcairo_native_runtime.so
> ```

Why did you rename the file extension `.a` -> `.so`?



_Code quote:_
```Shell
    mv ./cairo_native/target/release/libcairo_native_runtime.a ./libcairo_native_runtime.so
```

avi-starkware (commented):
> Reviewed 1 of 2 files at r1, 2 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @rodrigo-pino)

rodrigo-pino (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 1 of 4 files reviewed, all discussions resolved (waiting on @avi-starkware)

___
*[`scripts/dependencies.sh` line 73 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O967xDQ7lSapoJvdnqD:-O97stgV0sxr_oeVLFj5:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/cfe157b752f3c46c124ebe9952e75d96a386b2d6/scripts/dependencies.sh#L73)):*
<details><summary><i>Previously, avi-starkware wrote…</i></summary><blockquote>

Why did you rename the file extension `.a` -> `.so`?
</blockquote></details>

Done

avi-starkware (commented):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 2 of 4 files reviewed, all discussions resolved (waiting on @rodrigo-pino)

noaov1 (changes requested):
> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, 1 of 1 files at r8, 1 of 2 files at r11, 2 of 2 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: all files reviewed, 4 unresolved discussions (waiting on @rodrigo-pino)

___
*[`scripts/dependencies.sh` line 50 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eIgsEE0cmw3uoOyOJ:-O9eIgsEE0cmw3uoOyOK:b3ksxgx) ([raw file](https://github.com/starkware-libs/sequencer/blob/81bd5eb8f0dcf4dda5dcfef210787ce4b7594572/scripts/dependencies.sh#L50)):*
> ```Shell
> 
> function compile_cairo_native_runtime() {
>     # First we need to make sure Cargo exists
> ```

Can we assume it is installed?
@alon-dotan-starkware 



_Code quote:_
```Shell
# First we need to make sure Cargo exists
```
___
*[`scripts/dependencies.sh` line 66 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eHlC14vRY7TO4e6lA:-O9eHlC14vRY7TO4e6lB:bln9c2y) ([raw file](https://github.com/starkware-libs/sequencer/blob/81bd5eb8f0dcf4dda5dcfef210787ce4b7594572/scripts/dependencies.sh#L66)):*
> ```Shell
>     fi
> 
>     # Then we clone and build the runtime from the repo
> ```

Why is it needed? To get the `libcairo_native_runtime.a`?



_Code quote:_
```Shell
# Then we clone and build the runtime from the repo
```
___
*[`scripts/dependencies.sh` line 69 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eI4vAAsQckiRbiyii:-O9eI4vAAsQckiRbiyij:boshphq) ([raw file](https://github.com/starkware-libs/sequencer/blob/81bd5eb8f0dcf4dda5dcfef210787ce4b7594572/scripts/dependencies.sh#L69)):*
> ```Shell
>     git clone https://github.com/lambdaclass/cairo_native.git
>     pushd ./cairo_native || exit 1
>     git switch v0.2.0-alpha.2 --detach
> ```

Do we want to hardcode here the cairo native version? Can we use the one defined in the Cargo.toml?



_Code quote:_
```Shell
    git switch v0.2.0-alpha.2 --detach
```
___
*[`scripts/sequencer-ci.Dockerfile` line 15 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eGEf5BVSkYzsURUo3:-O9eGEf5BVSkYzsURUo4:btyvrix) ([raw file](https://github.com/starkware-libs/sequencer/blob/81bd5eb8f0dcf4dda5dcfef210787ce4b7594572/scripts/sequencer-ci.Dockerfile#L15)):*
> ```
> USER ${USERNAME}
> 
> WORKDIR /app
> ```

What does it mean?
@alon-dotan-starkware 



_Code quote:_
```
WORKDIR /app
```

noaov1 (commented):
> Reviewed 1 of 2 files at r18, 2 of 2 files at r19, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @alon and @rodrigo-pino)

PearsonWhite (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @alon, @alon-dotan-starkware, and @noaov1)

___
*[`scripts/sequencer-ci.Dockerfile` line 15 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eGEf5BVSkYzsURUo3:-O9lDLsl5FZvYSNqY4LD:b825ide) ([raw file](https://github.com/starkware-libs/sequencer/blob/22984131e3eb90664d0d5663bc0c0f3ab94b7a18/scripts/sequencer-ci.Dockerfile#L15)):*

What does it mean?
@alon-dotan-starkware
</blockquote></details>

`WORKDIR <dir>` makes all subsequent commands run under `<dir>`. Additionally, when running the docker the default path with be `<dir>`. For example `docker run -it sequencer-ci /bin/bash` lands the user in a bash shell in the directory `/app`.

This is so we don't run everything in the root dir (`/`), which has tighter user permissions, is owned by root, and sometimes has other restrictions.

alon-dotan-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @alon and @noaov1)

___
*[`scripts/dependencies.sh` line 66 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eHlC14vRY7TO4e6lA:-O9msUuO0G9MiI0PIZhl:bj0tffp) ([raw file](https://github.com/starkware-libs/sequencer/blob/22984131e3eb90664d0d5663bc0c0f3ab94b7a18/scripts/dependencies.sh#L66)):*
<details><summary><i>Previously, rodrigo-pino (Rodrigo) wrote…</i></summary><blockquote>

Right now we are building it, we can make download a release from the github as well.
</blockquote></details>

if we need to buid use git submodule, if we downloading use build.rs

alon-dotan-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @alon and @noaov1)

___
*[`scripts/dependencies.sh` line 50 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eIgsEE0cmw3uoOyOJ:-O9msfjBEuy8J6ct3BML:bw9nl61) ([raw file](https://github.com/starkware-libs/sequencer/blob/22984131e3eb90664d0d5663bc0c0f3ab94b7a18/scripts/dependencies.sh#L50)):*
<details><summary><i>Previously, rodrigo-pino (Rodrigo) wrote…</i></summary><blockquote>

Without it certain workflows failed
</blockquote></details>

so lets fix the workflows...
remove this flow please

alon-dotan-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @alon and @noaov1)

___
*[`scripts/dependencies.sh` line 69 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eI4vAAsQckiRbiyii:-O9msphbAGII5tmS-0Ba:b-ssig33) ([raw file](https://github.com/starkware-libs/sequencer/blob/22984131e3eb90664d0d5663bc0c0f3ab94b7a18/scripts/dependencies.sh#L69)):*
<details><summary><i>Previously, rodrigo-pino (Rodrigo) wrote…</i></summary><blockquote>

Let me look into it and see if there is a way of doing it without sacrificing simplicity
</blockquote></details>

remove this flow, if cairo_native became  dependency for this project use cargo to manage it or git submodule

alon-dotan-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @alon and @noaov1)

___
*[`scripts/sequencer-ci.Dockerfile` line 15 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eGEf5BVSkYzsURUo3:-O9mtBS69PTCtToH9rSY:b-q0vila) ([raw file](https://github.com/starkware-libs/sequencer/blob/22984131e3eb90664d0d5663bc0c0f3ab94b7a18/scripts/sequencer-ci.Dockerfile#L15)):*
<details><summary><i>Previously, PearsonWhite wrote…</i></summary><blockquote>

`WORKDIR <dir>` makes all subsequent commands run under `<dir>`. Additionally, when running the docker the default path with be `<dir>`. For example `docker run -it sequencer-ci /bin/bash` lands the user in a bash shell in the directory `/app`.

This is so we don't run everything in the root dir (`/`), which has tighter user permissions, is owned by root, and sometimes has other restrictions.
</blockquote></details>

why we need to use different dir?

rodrigo-pino (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 1 of 8 files reviewed, 4 unresolved discussions (waiting on @alon and @noaov1)

___
*[`scripts/dependencies.sh` line 50 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eIgsEE0cmw3uoOyOJ:-O9nrb2D5OZDOtvP7V11:banqhq9) ([raw file](https://github.com/starkware-libs/sequencer/blob/22984131e3eb90664d0d5663bc0c0f3ab94b7a18/scripts/dependencies.sh#L50)):*
<details><summary><i>Previously, alon-dotan-starkware wrote…</i></summary><blockquote>

so lets fix the workflows...
remove this flow please
</blockquote></details>

Yep, makes sense!
___
*[`scripts/dependencies.sh` line 66 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eHlC14vRY7TO4e6lA:-O9o6XkWDqGPVAeUwdTS:b8rtoqp) ([raw file](https://github.com/starkware-libs/sequencer/blob/22984131e3eb90664d0d5663bc0c0f3ab94b7a18/scripts/dependencies.sh#L66)):*
<details><summary><i>Previously, alon-dotan-starkware wrote…</i></summary><blockquote>

if we need to buid use git submodule, if we downloading use build.rs
</blockquote></details>

We can use `build.rs` for just downloading but we need both `build.rs` and git submoule in case of compiling. I chose compiling over downloading. 

Whatever route we are to pick, both are going to have the same caveats:

1. If there is any modification to the actual runtime (like deleting it) it won't get recompiled/redownloaded if Cargo thinks the projects needs no recompilation (hence build.rs won't get triggered).


2. There is no way to have a "unified" version control over Native:
-  With the **compilation route** when updating the native version in Cargo.toml then the submodule commit should be updated as well. The native version in Cargo.toml cannot point towards the submodule because both projects (sequencer and Cairo native) are defined as workspace roots.
- With the **download route** there is no way inside the build.rs of communicating the  cairo native version we want to use because it's a runtime dependency and build.rs don't have access to it.


I'll try to think a bit more on this, maybe there is a way reduce this cons, or try another way altogether. Looking forward to hearing your thoughts.
___
*[`scripts/dependencies.sh` line 69 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eI4vAAsQckiRbiyii:-O9y500X8sE0ejXJxdxd:b-8fx2u7) ([raw file](https://github.com/starkware-libs/sequencer/blob/22984131e3eb90664d0d5663bc0c0f3ab94b7a18/scripts/dependencies.sh#L69)):*
<details><summary><i>Previously, alon-dotan-starkware wrote…</i></summary><blockquote>

remove this flow, if cairo_native became  dependency for this project use cargo to manage it or git submodule
</blockquote></details>

:+1: 
___
*[`scripts/sequencer-ci.Dockerfile` line 15 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eGEf5BVSkYzsURUo3:-O9y57-e7tAQjbLyI_UU:btdbqyi) ([raw file](https://github.com/starkware-libs/sequencer/blob/22984131e3eb90664d0d5663bc0c0f3ab94b7a18/scripts/sequencer-ci.Dockerfile#L15)):*
<details><summary><i>Previously, alon-dotan-starkware wrote…</i></summary><blockquote>

why we need to use different dir?
</blockquote></details>

No longer relevant

noaov1 (changes requested):
> Reviewed 6 of 7 files at r22, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @alon-dotan-starkware, @avi-starkware, @PearsonWhite, and @rodrigo-pino)

___
*[`Cargo.toml` line 99 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAH1NS7D3wi-5XdCGEX:-OAH1NS7D3wi-5XdCGEY:b-u17zji) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/Cargo.toml#L99)):*
> ```TOML
> cairo-lang-starknet-classes = "2.8.2"
> cairo-lang-utils = "2.8.2"
> # Cannot use the git submodule at crates/blockifier/cairo_native/ because it is also declared as a "workspace"
> ```

Can you please elaborate?



_Suggestion:_
```TOML
# Important: when updated, make sure to update the cairo-native submodule as well.
```

alon-dotan-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 7 of 8 files reviewed, 2 unresolved discussions (waiting on @avi-starkware, @noaov1, @PearsonWhite, and @rodrigo-pino)

___
*[`scripts/dependencies.sh` line 66 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eHlC14vRY7TO4e6lA:-OAH9JMx104FVIQck_KF:bqmor7c) ([raw file](https://github.com/starkware-libs/sequencer/blob/22984131e3eb90664d0d5663bc0c0f3ab94b7a18/scripts/dependencies.sh#L66)):*
<details><summary><i>Previously, rodrigo-pino (Rodrigo) wrote…</i></summary><blockquote>

We can use `build.rs` for just downloading but we need both `build.rs` and git submoule in case of compiling. I chose compiling over downloading. 

Whatever route we are to pick, both are going to have the same caveats:

1. If there is any modification to the actual runtime (like deleting it) it won't get recompiled/redownloaded if Cargo thinks the projects needs no recompilation (hence build.rs won't get triggered).


2. There is no way to have a "unified" version control over Native:
-  With the **compilation route** when updating the native version in Cargo.toml then the submodule commit should be updated as well. The native version in Cargo.toml cannot point towards the submodule because both projects (sequencer and Cairo native) are defined as workspace roots.
- With the **download route** there is no way inside the build.rs of communicating the  cairo native version we want to use because it's a runtime dependency and build.rs don't have access to it.


I'll try to think a bit more on this, maybe there is a way reduce this cons, or try another way altogether. Looking forward to hearing your thoughts.
</blockquote></details>

cant we define something like this?https://stackoverflow.com/questions/65766269/can-cargo-toml-contain-custom-properties

alon-dotan-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 7 of 8 files reviewed, 3 unresolved discussions (waiting on @avi-starkware, @noaov1, @PearsonWhite, and @rodrigo-pino)

___
*[`.gitmodules` line 3 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAH9yJU7t-9K3SglEP2:-OAH9yJU7t-9K3SglEP3:b-pys6vk) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/.gitmodules#L3)):*
> ```gitconfig
> [submodule "crates/blockifier/cairo_native"]
> 	path = crates/blockifier/cairo_native
> 	url = https://github.com/lambdaclass/cairo_native
> ```

* fix tabs
* use specific commit



_Code quote:_
```gitconfig
[submodule "crates/blockifier/cairo_native"]
	path = crates/blockifier/cairo_native
	url = https://github.com/lambdaclass/cairo_native
```

avi-starkware:
*[`crates/blockifier/build.rs` line 47 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAHarIL08a_6HBNBPFw:-OAHarIL08a_6HBNBPFx:b-wcizwd) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/crates/blockifier/build.rs#L47)):*
> ```XML
> fn main() {
>     // todo(rdr): known issue of this approach is that if the user deletes libcairo_native_runtime.a
>     // and re-compiles with further changes, `build.rs` won't activate
> ```

You can add a `cargo:rerun-if-` flag here to make sure that the build is rerun if the file has been changed since the last build command.
(Make sure the path to the file is correct)

Do we additionally need a `cargo:rerun-if-env-change` flag to check the env var is still valid as well?



_Suggestion:_
```XML
    println!("cargo:rerun-if-changed=libcairo_native_runtime.a");
```

rodrigo-pino (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 7 of 8 files reviewed, 4 unresolved discussions (waiting on @avi-starkware, @noaov1, and @PearsonWhite)

___
*[`Cargo.toml` line 99 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAH1NS7D3wi-5XdCGEX:-OAHkevXBerRCLshXGhE:b-mdruhy) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/Cargo.toml#L99)):*

Can you please elaborate?
</blockquote></details>


Both the sequencer repo and Cairo Native are defined as cargo workspaces. When using the git-submodule approach, the whole Cairo Native repo is added as an extra folder (as if defining another crate just like `blockifier`, starknet_api`, etc.)

If we set Cairo Native as a path dependency in Cargo.toml, it being treated as another crate which is considered a Workspace won't allow it to compile.
___
*[`crates/blockifier/build.rs` line 47 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAHarIL08a_6HBNBPFw:-OAHmctc0ZePsf2XBPQB:blce68p) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/crates/blockifier/build.rs#L47)):*
> You can add a `cargo:rerun-if-` flag here to make sure that the build is rerun if the file has been changed since the last build command.  
(Make sure the path to the file is correct)

Will try this out

> Do we additionally need a `cargo:rerun-if-env-change` flag to check the env var is still valid as well?

It might be useful for development, will add it as well
___
*[`scripts/dependencies.sh` line 66 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-O9eHlC14vRY7TO4e6lA:-OAHtaFA7UyGHvdMuZyj:brm8mso) ([raw file](https://github.com/starkware-libs/sequencer/blob/22984131e3eb90664d0d5663bc0c0f3ab94b7a18/scripts/dependencies.sh#L66)):*
<details><summary><i>Previously, alon-dotan-starkware wrote…</i></summary><blockquote>

cant we define something like this?https://stackoverflow.com/questions/65766269/can-cargo-toml-contain-custom-properties
</blockquote></details>

Mmm I personally don't see how compilation features can help us with the problem at hand. They are usually used to add/alter some specific behaviour of the code but I don't know how we can use that in our advantage in this specific case.

There was a suggestion up here https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAHarIL08a_6HBNBPFw , I'll try that out

rodrigo-pino (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 0 of 11 files reviewed, 4 unresolved discussions (waiting on @alon-dotan-starkware, @avi-starkware, @noaov1, and @PearsonWhite)

___
*[`.gitmodules` line 3 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAH9yJU7t-9K3SglEP2:-OAHyl1641Nd2LLNdwzl:b1isn9t) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/.gitmodules#L3)):*
<details><summary><i>Previously, alon-dotan-starkware wrote…</i></summary><blockquote>

* fix tabs
* use specific commit
</blockquote></details>

Done. 

Regarding specific commits is not possible only branch are "targettables", at least from here. 

The only way to updating the git commit/release tag is by accesing the submodule through git and changing the commit and adding this action as a commit.

noaov1 (commented):
> Reviewed 1 of 2 files at r44, 6 of 10 files at r45, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 7 of 11 files reviewed, 5 unresolved discussions (waiting on @alon-dotan-starkware, @avi-starkware, @PearsonWhite, and @rodrigo-pino)

___
*[`taplo.toml` line 4 at r46](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAmP--j5D3yGzBMSCtE:-OAmP--j5D3yGzBMSCtF:btrof9t) ([raw file](https://github.com/starkware-libs/sequencer/blob/0d70854a71a5cc570d5e04c0319a6584ef12c7a7/taplo.toml#L4)):*
> ```TOML
>   "crates/blockifier/cairo_native/**/*.toml",
>   "crates/native_blockifier/.cargo/config.toml",
> ]
> ```

@dorimedini-starkware PTAL



_Code quote:_
```TOML
exclude = [
  "crates/blockifier/cairo_native/**/*.toml",
  "crates/native_blockifier/.cargo/config.toml",
]
```
___
*[`.github/workflows/blockifier_ci.yml` line 48 at r46](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAmPEzW9dsA5vy83tKf:-OAmPEzW9dsA5vy83tKg:b-u17ziw) ([raw file](https://github.com/starkware-libs/sequencer/blob/0d70854a71a5cc570d5e04c0319a6584ef12c7a7/.github/workflows/blockifier_ci.yml#L48)):*
> ```YAML
>       - uses: actions/checkout@v4
>         with:
>           # required to clone native as a gitsubmodule
> ```

Can you please elaborate? 



_Code quote:_
```YAML
          # required to clone native as a gitsubmodule
```

noaov1 (commented):
> Reviewed 2 of 10 files at r45.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 9 of 11 files reviewed, 4 unresolved discussions (waiting on @alon-dotan-starkware, @avi-starkware, @PearsonWhite, and @rodrigo-pino)

dorimedini-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 9 of 11 files reviewed, 4 unresolved discussions (waiting on @alon-dotan-starkware, @avi-starkware, @noaov1, @PearsonWhite, and @rodrigo-pino)

___
*[`taplo.toml` line 4 at r46](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAmP--j5D3yGzBMSCtE:-OAmwZBqExBeFWN0nGFl:b9uxliq) ([raw file](https://github.com/starkware-libs/sequencer/blob/0d70854a71a5cc570d5e04c0319a6584ef12c7a7/taplo.toml#L4)):*

@dorimedini-starkware PTAL
</blockquote></details>

if we are going with the git submodule approach, then this seems fine.
no other lints need to be turned off explicitly for the submodule? clippy, rustfmt etc won't run on it?

avi-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 9 of 11 files reviewed, 4 unresolved discussions (waiting on @alon-dotan-starkware, @noaov1, @PearsonWhite, and @rodrigo-pino)

___
*[`crates/blockifier/build.rs` line 47 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAHarIL08a_6HBNBPFw:-OAnBTwzDieoLXrqAJaW:bfskjym) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/crates/blockifier/build.rs#L47)):*
<details><summary><i>Previously, rodrigo-pino (Rodrigo) wrote…</i></summary><blockquote>

Done
</blockquote></details>

I think the `println!` directives for the `rerun-if` flags have to be at the beginning of `main()` to work. Additionally, it doesn't make sense to have the flags that check if the script should be rerun depend on variables that only exist if the entire script has indeed been rerun.

Did you check these flags have the expected behavior?

I tried to test if the flags work properly on my machine locally, but I can't get the build script to run without error.
This is the error I am getting:
```
error: failed to run custom build command for `blockifier v0.0.0 (/home/avi/workspace/sequencer2/crates/blockifier)`

Caused by:
  process didn't exit successfully: `/home/avi/workspace/sequencer2/target/debug/build/blockifier-151d1608d6c400a7/build-script-build` (exit status: 101)
  --- stderr
  error: package ID specification `cairo-native-runtime` did not match any packages
  thread 'main' panicked at crates/blockifier/build.rs:24:9:
  Building cairo native runtime failed: exit status: 101
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
```

alon-dotan-starkware (changes requested):
> Reviewed 1 of 10 files at r39, 1 of 10 files at r45.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 9 of 11 files reviewed, 5 unresolved discussions (waiting on @noaov1, @PearsonWhite, and @rodrigo-pino)

___
*[`.cargo/config.toml` line 2 at r46](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAqEhWu8gcLvgJ7SoEJ:-OAqEhWv9mhCd6V-fi1E:b-ml0te4) ([raw file](https://github.com/starkware-libs/sequencer/blob/0d70854a71a5cc570d5e04c0319a6584ef12c7a7/.cargo/config.toml#L2)):*
> ```TOML
> [env]
> CAIRO_NATIVE_RUNTIME_LIBRARY = "./libcairo_native_runtime.a"
> ```

cant we handle it better? its cant be part of build.rs? I prefer to use env vars less as possible



_Code quote:_
```TOML
CAIRO_NATIVE_RUNTIME_LIBRARY = "./libcairo_native_runtime.a"
```

alon-dotan-starkware (commented):
> Reviewed 1 of 10 files at r45.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 10 of 11 files reviewed, 6 unresolved discussions (waiting on @noaov1, @PearsonWhite, and @rodrigo-pino)

___
*[`crates/blockifier/cairo_native` line 1 at r46](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAqFCbu5fGm70AeMPgY:-OAqFCbu5fGm70AeMPgZ:bpen699) ([raw file](https://github.com/starkware-libs/sequencer/blob/0d70854a71a5cc570d5e04c0319a6584ef12c7a7/crates/blockifier/cairo_native#L1)):*
> ```
> Subproject commit b5769e4f6ba914b36eef68e0b1f71c791d7d075c
> ```

what does this stand for? Who source this file? cargo?

alon-dotan-starkware (changes requested):
> Reviewed 1 of 2 files at r44, 5 of 10 files at r45, 1 of 1 files at r46, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: all files reviewed, 6 unresolved discussions (waiting on @noaov1, @PearsonWhite, and @rodrigo-pino)

___
*[`.gitmodules` line 2 at r46](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAqFttQDjRAuaY1dgK1:-OAqFttQDjRAuaY1dgK2:bv63z74) ([raw file](https://github.com/starkware-libs/sequencer/blob/0d70854a71a5cc570d5e04c0319a6584ef12c7a7/.gitmodules#L2)):*
> ```gitconfig
> [submodule "crates/blockifier/cairo_native"]
>     path = crates/blockifier/cairo_native
> ```

why not at the root of the src dir?  create dedicated dir for it



_Code quote:_
```gitconfig
 path = crates/blockifier/cairo_native
```

alon-dotan-starkware (commented):
> Reviewed 1 of 10 files at r45.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: all files reviewed, 5 unresolved discussions (waiting on @noaov1, @PearsonWhite, and @rodrigo-pino)

rodrigo-pino (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: all files reviewed, 5 unresolved discussions (waiting on @alon-dotan-starkware, @avi-starkware, @noaov1, and @PearsonWhite)

___
*[`.gitmodules` line 2 at r46](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAqFttQDjRAuaY1dgK1:-OAr-BKU5wm0ylEUW4Ov:b2ps6an) ([raw file](https://github.com/starkware-libs/sequencer/blob/0d70854a71a5cc570d5e04c0319a6584ef12c7a7/.gitmodules#L2)):*
<details><summary><i>Previously, alon-dotan-starkware wrote…</i></summary><blockquote>

why not at the root of the src dir?  create dedicated dir for it
</blockquote></details>

It is at the root of the src dir of the only crate that needs it, which is the blockifier. For me it doesn't make sense to be outside of it but ok
___
*[`taplo.toml` line 4 at r46](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAmP--j5D3yGzBMSCtE:-OAr-SxJ2KeccoX5Arz-:b4zq9li) ([raw file](https://github.com/starkware-libs/sequencer/blob/0d70854a71a5cc570d5e04c0319a6584ef12c7a7/taplo.toml#L4)):*

if we are going with the git submodule approach, then this seems fine.
no other lints need to be turned off explicitly for the submodule? clippy, rustfmt etc won't run on it?
</blockquote></details>

All the other lints are working over it, taplo is the only one ignoring it. We can make a PR to Cairo Native that fix their taplo issues and then remove this exclusion later on
___
*[`.cargo/config.toml` line 2 at r46](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAqEhWu8gcLvgJ7SoEJ:-OAr-r0N4u1SIhiu3r0O:b-jfi0oz) ([raw file](https://github.com/starkware-libs/sequencer/blob/0d70854a71a5cc570d5e04c0319a6584ef12c7a7/.cargo/config.toml#L2)):*
<details><summary><i>Previously, alon-dotan-starkware wrote…</i></summary><blockquote>

cant we handle it better? its cant be part of build.rs? I prefer to use env vars less as possible
</blockquote></details>

No it cannot be since it is used as runtime dependency, meaning the var has to be set during runtime. It is used during building just to put it in the correct place (a.k.a. the place where the execution is going to look for it).
___
*[`.github/workflows/blockifier_ci.yml` line 48 at r46](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAmPEzW9dsA5vy83tKf:-OAr0BxPBVD4VaazOPiC:b40jyl7) ([raw file](https://github.com/starkware-libs/sequencer/blob/0d70854a71a5cc570d5e04c0319a6584ef12c7a7/.github/workflows/blockifier_ci.yml#L48)):*

Can you please elaborate?
</blockquote></details>

If we don't add the `submodule: recursive` then git by default won't clone it making blockifier compilation fail afterwards (due to the new `build.rs`)
___
*[`crates/blockifier/build.rs` line 47 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAHarIL08a_6HBNBPFw:-OAr0jFo5LILLT7Ki8TK:b-h00dmk) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/crates/blockifier/build.rs#L47)):*
> think the `println!` directives for the `rerun-if` flags have to be at the beginning of `main()` to work.

Nop, they location is not important, I put them there to re-use variables. Can put them in front if you like that more.

> it doesn't make sense to have the flags that check if the script should be rerun depend on variables that only exist if the entire script has indeed been rerun.

I see no relationship between both mechanisms. The directives are there in case there is a change in the env-var. The script fails to run if the env vars are unset.
 If the script never got built then directives are not important since nothing go built.

As I mentioned before, the directives can go anywhere. In our specific context they are only relevant when `cairo_native` feature is enabled, so the only requirement is that they get printed if that condition is met. Again, I put them there so that any change in the `build.rs` will keep it consistent with the directives, I can put it in front if you think it's better.

> I tried to test if the flags work properly on my machine locally, but I can't get the build script to run without error.  
This is the error I am getting:

The error is caused because you don't have the submodule. You can fix it by running:
```
git submodule update --init --recursive
```
I believe this can be detected easily in the `build.rs`, I'll add to it
___
*[`crates/blockifier/cairo_native` line 1 at r46](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAqFCbu5fGm70AeMPgY:-OAr3FDIBBB-bUq2Qs3W:b-apci6l) ([raw file](https://github.com/starkware-libs/sequencer/blob/0d70854a71a5cc570d5e04c0319a6584ef12c7a7/crates/blockifier/cairo_native#L1)):*
<details><summary><i>Previously, alon-dotan-starkware wrote…</i></summary><blockquote>

what does this stand for? Who source this file? cargo?
</blockquote></details>

That's the representation of the git submodule inside git, so that then git knows what to pull. It's created automatically by git

avi-starkware:
*[`crates/blockifier/build.rs` line 47 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAHarIL08a_6HBNBPFw:-OArXy1y36ZY6wE4xhRC:b-oxx370) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/crates/blockifier/build.rs#L47)):*
<details><summary><i>Previously, avi-starkware wrote…</i></summary><blockquote>

Now it works thanks!

Can you please add a comment documenting the need for this additional command?

I played around with building the crate and rerunning it, and it seems that the `cargo build` always reruns the `build.rs` script as long as the `rerun-if-changed` directives are present, regardless of whether the files have actually changed. This holds true also when I put the directives at the beginning of `main` (with explicit paths - no variables).

When I experimented with `rerun-if-changed` directive on a simpler `build.rs` script, it behaved as expected: rerunning when there were changes and skipping `build.rs` if there were no changes, so I am not sure what is happening here.

Did you find a case where the script is not rerun in its current state? (I can see the script is rerun since `crates/blockifier/libcairo_native_runtime.a` is modified each time I run the script.)
</blockquote></details>

Regardless, this is non-blocking.

rodrigo-pino (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: all files reviewed, 4 unresolved discussions (waiting on @alon-dotan-starkware, @noaov1, and @PearsonWhite)

___
*[`crates/blockifier/build.rs` line 47 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAHarIL08a_6HBNBPFw:-OArT4ED1c45Rvx2hIpW:b-ouazlp) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/crates/blockifier/build.rs#L47)):*
> Can you please add a comment documenting the need for this additional command?

The error can be detected, would it be good enough suggesting the fix when facing it the first time?
Maybe setting the command as part of the dependencies.sh script? I am unsure how it would affect the CI. Will try this out as well (cc: @alon-dotan-starkware)

Regarding your last three paragraphs, I find it as an odd behaviour. After I implemented the solution I tested it locally and it worked as expected (tried each directive individually) and the build was successfully triggered each time. Then, tried two+ builds in a row and after the first one, the others were no-op (also verified that the `libcairo_native_runtime.a` remained unmodified)

I'll ask for someone from the team to check if they have the same behaviour as your. We are all using macOS making our results biased. Can you see if your results persist on other Linux machines. (Thinking OS file system differences as a root cause). Finally, to completely rule out human error, are you sure that when running this script (as it is presented in this branch) twice in a row the second re-builds Native Runtime? Could you detect which directive of the three is causing the issue?

rodrigo-pino (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: all files reviewed, 5 unresolved discussions (waiting on @alon-dotan-starkware, @noaov1, and @PearsonWhite)

___
*[`crates/blockifier/build.rs` line 47 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAHarIL08a_6HBNBPFw:-OAr_6Yx5bRzu2WcbSnJ:b2ngvnt) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/crates/blockifier/build.rs#L47)):*
<details><summary><i>Previously, rodrigo-pino (Rodrigo) wrote…</i></summary><blockquote>

> Can you please add a comment documenting the need for this additional command?

The error can be detected, would it be good enough suggesting the fix when facing it the first time?
Maybe setting the command as part of the dependencies.sh script? I am unsure how it would affect the CI. Will try this out as well (cc: @alon-dotan-starkware)

Regarding your last three paragraphs, I find it as an odd behaviour. After I implemented the solution I tested it locally and it worked as expected (tried each directive individually) and the build was successfully triggered each time. Then, tried two+ builds in a row and after the first one, the others were no-op (also verified that the `libcairo_native_runtime.a` remained unmodified)

I'll ask for someone from the team to check if they have the same behaviour as your. We are all using macOS making our results biased. Can you see if your results persist on other Linux machines. (Thinking OS file system differences as a root cause). Finally, to completely rule out human error, are you sure that when running this script (as it is presented in this branch) twice in a row the second re-builds Native Runtime? Could you detect which directive of the three is causing the issue?
</blockquote></details>

...

avi-starkware:
*[`crates/blockifier/build.rs` line 47 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAHarIL08a_6HBNBPFw:-OAr_Kps3Tba7oj4Z8Od:bdmebw5) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/crates/blockifier/build.rs#L47)):*
> The error can be detected, would it be good enough suggesting the fix when facing it the first time?

This is a good idea! I think adding the explanation to the error message, as you suggested, would be sufficient.

I'll do some more testing on the rerun behavior on my machine and other linux machines.

rodrigo-pino (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: all files reviewed, 5 unresolved discussions (waiting on @alon-dotan-starkware, @noaov1, and @PearsonWhite)

___
*[`crates/blockifier/build.rs` line 47 at r22](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAHarIL08a_6HBNBPFw:-OArkbtzE7wdtqLJYrkk:biqzi65) ([raw file](https://github.com/starkware-libs/sequencer/blob/8181f5041d2af4134c3d52773f8b519ed0d8ae5b/crates/blockifier/build.rs#L47)):*
> I'll do some more testing on the rerun behavior on my machine and other linux machines.

Thank you!

> ...

This previous message of mine was because I mistakenly set the thread as resolved :man_facepalming: , I've just noticed that it's meaning wasn't completely clear

alon-dotan-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 0 of 11 files reviewed, 2 unresolved discussions (waiting on @noaov1 and @PearsonWhite)

noaov1 (approved):
> Reviewed 1 of 2 files at r52, 9 of 10 files at r53, 1 of 1 files at r54, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: all files reviewed, 3 unresolved discussions (waiting on @alon-dotan-starkware, @PearsonWhite, and @rodrigo-pino)

___
*[`taplo.toml` line 4 at r46](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAmP--j5D3yGzBMSCtE:-OAvjlWTAJI-lEdsXdl0:b-i52qw5) ([raw file](https://github.com/starkware-libs/sequencer/blob/0d70854a71a5cc570d5e04c0319a6584ef12c7a7/taplo.toml#L4)):*
<details><summary><i>Previously, rodrigo-pino (Rodrigo) wrote…</i></summary><blockquote>

All the other lints are working over it, taplo is the only one ignoring it. We can make a PR to Cairo Native that fix their taplo issues and then remove this exclusion later on
</blockquote></details>

No need IMO
___
*[`crates/blockifier/build.rs` line 59 at r54](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAvloS504zn6_2NDuv2:-OAvloS504zn6_2NDuv3:bg865ce) ([raw file](https://github.com/starkware-libs/sequencer/blob/10efcfa2f3b5cc2cebd4cf71cbb9ec6f62518bdb/crates/blockifier/build.rs#L59)):*
> ```XML
>     println!("cargo::rerun-if-changed={}", runtime_expected_path.to_str().unwrap());
>     println!("cargo::rerun-if-env-changed={RUNTIME_LIBRARY}");
> }
> ```

Why do we need both?



_Code quote:_
```XML
    println!("cargo::rerun-if-changed={}", runtime_expected_path.to_str().unwrap());
    println!("cargo::rerun-if-env-changed={RUNTIME_LIBRARY}");
}
```
___
*[`crates/blockifier/build.rs` line 62 at r54](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAvluYrE8Rs-wZNY7ud:-OAvluYrE8Rs-wZNY7ue:b-nzctcb) ([raw file](https://github.com/starkware-libs/sequencer/blob/10efcfa2f3b5cc2cebd4cf71cbb9ec6f62518bdb/crates/blockifier/build.rs#L62)):*
> ```XML
> 
> fn main() {
>     if std::env::var("CARGO_FEATURE_CAIRO_NATIVE").is_ok() {
> ```

Can you please explain this check? (and maybe document)



_Code quote:_
```XML
if std::env::var("CARGO_FEATURE_CAIRO_NATIVE").is_ok() {
```

noaov1 (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: all files reviewed, 2 unresolved discussions (waiting on @alon-dotan-starkware, @PearsonWhite, and @rodrigo-pino)

rodrigo-pino (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: 10 of 11 files reviewed, 1 unresolved discussion (waiting on @alon-dotan-starkware, @noaov1, and @PearsonWhite)

___
*[`crates/blockifier/build.rs` line 59 at r54](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAvloS504zn6_2NDuv2:-OAvmvOH3grHjJYbGLq4:b-5zlz6q) ([raw file](https://github.com/starkware-libs/sequencer/blob/10efcfa2f3b5cc2cebd4cf71cbb9ec6f62518bdb/crates/blockifier/build.rs#L59)):*

Why do we need both?
</blockquote></details>

The first one in case the runtime gets deleted, since ti is a binary I think no one will be modifying it.

The second one is in case a developer changes the runtime library path to re-compile it in that location.
___
*[`crates/blockifier/build.rs` line 62 at r54](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-OAvluYrE8Rs-wZNY7ud:-OAvnNPu2nzJEKq4bKW4:bnl6ww6) ([raw file](https://github.com/starkware-libs/sequencer/blob/10efcfa2f3b5cc2cebd4cf71cbb9ec6f62518bdb/crates/blockifier/build.rs#L62)):*

Can you please explain this check? (and maybe document)
</blockquote></details>

Done , explanation added as a comment

noaov1 (approved):
> Reviewed 1 of 1 files at r55, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242)* status: all files reviewed, 1 unresolved discussion (waiting on @alon-dotan-starkware and @PearsonWhite)

avi-starkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1242#-:-OAw0t6R0Tk2DNJV_hpP:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @PearsonWhite)

PR #1591, opened 2024-10-28
build(blockifier_reexecution): serialize raw tx

aner-starkware:


dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1591)* status: all files reviewed, 3 unresolved discussions (waiting on @aner-starkware and @AvivYossef-starkware)

___
*[`crates/blockifier_reexecution/src/state_reader/test_state_reader.rs` line 140 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1591#-OAIUuMYBAac7XBjnPQf:-OAIUuMYBAac7XBjnPQg:b5p2l05) ([raw file](https://github.com/starkware-libs/sequencer/blob/b59012f0a3b0f438b3a738de6c1a40e955372079/crates/blockifier_reexecution/src/state_reader/test_state_reader.rs#L140)):*
> ```XML
>         });
>         Ok(self.0.send_rpc_request(method, params)?)
>     }
> ```

this should not be a public function - the API of the state reader shouldn't return serialized (raw) data, only deserialized objects.
if you want to test it, I would revert this function split and in the above test, compare `Transaction`s instead of `Value`s



_Code quote:_
```XML
    pub fn get_raw_tx_by_hash(&self, tx_hash: &str) -> ReexecutionResult<Value> {
        let method = "starknet_getTransactionByHash";
        let params = json!({
            "transaction_hash": tx_hash,
        });
        Ok(self.0.send_rpc_request(method, params)?)
    }
```
___
*[`crates/blockifier_reexecution/src/state_reader/utils.rs` line 65 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1591#-OAIWgmk-VkDkrKu1O3Y:-OAIWgmk-VkDkrKu1O3Z:b76650n) ([raw file](https://github.com/starkware-libs/sequencer/blob/b59012f0a3b0f438b3a738de6c1a40e955372079/crates/blockifier_reexecution/src/state_reader/utils.rs#L65)):*
> ```XML
> #[allow(dead_code)]
> pub(crate) fn serialize_from_raw(raw_object: &serde_json::Value) -> ReexecutionResult<String> {
>     Ok(serde_json::to_string_pretty(&raw_object).expect("Could not serialize json"))
> ```

consider also outputting the raw object that failed to serialize (non-blocking)



_Code quote:_
```XML
"Could not serialize json"
```
___
*[`crates/blockifier_reexecution/src/state_reader/utils.rs` line 66 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1591#-OAIWMMr-VD72GENne0k:-OAIWMMr-VD72GENne0l:b-s3pjf8) ([raw file](https://github.com/starkware-libs/sequencer/blob/b59012f0a3b0f438b3a738de6c1a40e955372079/crates/blockifier_reexecution/src/state_reader/utils.rs#L66)):*
> ```XML
> pub(crate) fn serialize_from_raw(raw_object: &serde_json::Value) -> ReexecutionResult<String> {
>     Ok(serde_json::to_string_pretty(&raw_object).expect("Could not serialize json"))
> }
> ```

this is just a test util?
if so, move it to the test module (where it's used);
if not, replace the `dead_code` with a `cfg(test)` (which you can remove when this function will be used in business logic)



_Code quote:_
```XML
#[allow(dead_code)]
pub(crate) fn serialize_from_raw(raw_object: &serde_json::Value) -> ReexecutionResult<String> {
    Ok(serde_json::to_string_pretty(&raw_object).expect("Could not serialize json"))
}
```

aner-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1591)* status: 1 of 3 files reviewed, 3 unresolved discussions (waiting on @AvivYossef-starkware and @dorimedini-starkware)

___
*[`crates/blockifier_reexecution/src/state_reader/test_state_reader.rs` line 140 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1591#-OAIUuMYBAac7XBjnPQf:-OAIaVMd04u3jLhdIwl0:b-386mzx) ([raw file](https://github.com/starkware-libs/sequencer/blob/b59012f0a3b0f438b3a738de6c1a40e955372079/crates/blockifier_reexecution/src/state_reader/test_state_reader.rs#L140)):*

this should not be a public function - the API of the state reader shouldn't return serialized (raw) data, only deserialized objects.
if you want to test it, I would revert this function split and in the above test, compare `Transaction`s instead of `Value`s
</blockquote></details>

I'll need the raw data in future development, in order to create the JSON files. I took your suggestion on the test, though (since comparing `Value`s seems to be problematic).
___
*[`crates/blockifier_reexecution/src/state_reader/utils.rs` line 66 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1591#-OAIWMMr-VD72GENne0k:-OAI_t3_CKUMo9XMW9c4:b-w15e97) ([raw file](https://github.com/starkware-libs/sequencer/blob/b59012f0a3b0f438b3a738de6c1a40e955372079/crates/blockifier_reexecution/src/state_reader/utils.rs#L66)):*

this is just a test util?
if so, move it to the test module (where it's used);
if not, replace the `dead_code` with a `cfg(test)` (which you can remove when this function will be used in business logic)
</blockquote></details>

This is not just a test util. It will be used in the future.

PR #1609, opened 2024-10-28
feat: integrate streaming with consensus proposals

guy-starkware:
This is still a draft and does not compile! Please look only at the flow and don't worry about the details. 

This PR combines the StreamHandler and ProposalParts structs into the flow of Consensus accepting proposals. 

The Consensus is given a receiver from the StreamHandler, which, in turn, listens to the network and collects messages, sending them as soon as they arrive in the correct order. 

The manager's `run_height` functions takes an additional receiver, and listens separately to regular messages and for proposals (both inside the same `select` statement), 

The new receiver is produced at the begining of the call to `run_consensus` by making a channel and giving one side of it to a newly created StreamHandler and the other side to the `run_height` function. The StreamHandler gets a separate spawned thread to run on, listening to the network. In the future it would also be listening to the consensus context for outgoing proposals. 

TODO: 
- [x] There are two implementations of ConsensusContext, I only updated the "sequencer" one. 
- [ ] There's a difference between Transaction and ExecutableTransaction that I don't really understand. Right now there's an error because some places want a vector of transactions and other places want a vector of executable transactions.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7420,6 +7420,7 @@ dependencies = [
  "papyrus_monitoring_gateway",
  "papyrus_network",
  "papyrus_p2p_sync",
+ "papyrus_protobuf",
  "papyrus_rpc",
  "papyrus_storage",
  "papyrus_sync",
--- a/crates/papyrus_node/Cargo.toml
+++ b/crates/papyrus_node/Cargo.toml
@@ -39,6 +39,7 @@ papyrus_consensus_orchestrator.workspace = true
 papyrus_monitoring_gateway.workspace = true
 papyrus_network.workspace = true
 papyrus_p2p_sync.workspace = true
+papyrus_protobuf.workspace = true
 papyrus_rpc = { workspace = true, optional = true }
 papyrus_storage.workspace = true
 papyrus_sync.workspace = true
--- a/crates/papyrus_node/src/run.rs
+++ b/crates/papyrus_node/src/run.rs
@@ -13,14 +13,16 @@ use papyrus_common::pending_classes::PendingClasses;
 use papyrus_config::presentation::get_config_presentation;
 use papyrus_config::validators::config_validate;
 use papyrus_consensus::config::ConsensusConfig;
+use papyrus_consensus::stream_handler::StreamHandler;
 use papyrus_consensus_orchestrator::papyrus_consensus_context::PapyrusConsensusContext;
 use papyrus_monitoring_gateway::MonitoringServer;
 use papyrus_network::gossipsub_impl::Topic;
-use papyrus_network::network_manager::NetworkManager;
+use papyrus_network::network_manager::{BroadcastTopicChannels, NetworkManager};
 use papyrus_network::{network_manager, NetworkConfig};
 use papyrus_p2p_sync::client::{P2PSyncClient, P2PSyncClientChannels};
 use papyrus_p2p_sync::server::{P2PSyncServer, P2PSyncServerChannels};
 use papyrus_p2p_sync::{Protocol, BUFFER_SIZE};
+use papyrus_protobuf::consensus::{ProposalPart, StreamMessage};
 #[cfg(feature = "rpc")]
 use papyrus_rpc::run_server;
 use papyrus_storage::{open_storage, update_storage_metrics, StorageReader, StorageWriter};
@@ -49,6 +51,7 @@ const DEFAULT_LEVEL: LevelFilter = LevelFilter::INFO;
 // different genesis hash.
 // TODO: Consider moving to a more general place.
 const GENESIS_HASH: &str = "0x0";
+pub const NETWORK_TOPIC: &str = "consensus_proposals";
 
 // TODO(dvir): add this to config.
 // Duration between updates to the storage metrics (those in the collect_storage_metrics function).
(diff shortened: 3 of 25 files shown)


matan-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1609)* status: 0 of 5 files reviewed, 6 unresolved discussions (waiting on @asmaastarkware and @guy-starkware)

___
*[`crates/sequencing/papyrus_consensus/src/manager.rs` line 172 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1609#-OARrzfjEQn1jKlLGApk:-OARrzfjEQn1jKlLGApl:b9ulpco) ([raw file](https://github.com/starkware-libs/sequencer/blob/93d5db27630b5633f298caa61222d82f1a20e4ae/crates/sequencing/papyrus_consensus/src/manager.rs#L172)):*
> ```XML
>                         }
>                     };
>                     self.handle_proposal(context, height, &mut shc, proposal_init.into(), content_receiver).await?
> ```

Minor style point, I prefer to put as little code as possible within macro blocks since it's harder for the IDE to work with



_Code quote:_
```XML
                    let Some(first_part) = content_receiver.next().await else {
                        return Err(ConsensusError::InternalNetworkError("Proposal receiver closed".to_string()));
                    };
                    let proposal_init = match first_part {
                        ProposalPart::Init(init) => init,
                        _ => {
                            return Err(ConsensusError::InternalNetworkError(
                                "Expected first part of proposal to be Init variant".to_string(),
                            ));
                        }
                    };
                    self.handle_proposal(context, height, &mut shc, proposal_init.into(), content_receiver).await?
```
___
*[`crates/sequencing/papyrus_consensus/src/manager.rs` line 202 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1609#-OARsRmg5_wUUcXdi3wn:-OARsRmg5_wUUcXdi3wo:b-b1jyyo) ([raw file](https://github.com/starkware-libs/sequencer/blob/93d5db27630b5633f298caa61222d82f1a20e4ae/crates/sequencing/papyrus_consensus/src/manager.rs#L202)):*
> ```XML
>         ContextT: ConsensusContext,
>     {
>         // TODO(guyn): what is the right thing to do if proposal's height doesn't match?
> ```

We should do the same as before. Cache the channel and when we get to the height send it to SHC (create a new cache for this type)



_Code quote:_
```XML
        // TODO(guyn): what is the right thing to do if proposal's height doesn't match?
```
___
*[`crates/sequencing/papyrus_consensus/src/single_height_consensus.rs` line 127 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1609#-OARsjy668hyuUUmcMS2:-OARsjy668hyuUUmcMS3:b-7zi8ni) ([raw file](https://github.com/starkware-libs/sequencer/blob/93d5db27630b5633f298caa61222d82f1a20e4ae/crates/sequencing/papyrus_consensus/src/single_height_consensus.rs#L127)):*
> ```XML
> 
>         // TODO(guyn): I think we should rename "block" and "fin" to be more descriptive!
>         let (block, fin) = match block_receiver.await {
> ```

WDYT?



_Suggestion:_
```XML
block_built_from_content, fin_sent_by_proposer
```
___
*[`crates/sequencing/papyrus_consensus/src/types.rs` line 72 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1609#-OARr1Ob5c3NOP0Sn5r_:-OARr1Ob5c3NOP0Sn5ra:b89f9c8) ([raw file](https://github.com/starkware-libs/sequencer/blob/a3b26a19892d20022cee02b1deddd0c6915ffe0f/crates/sequencing/papyrus_consensus/src/types.rs#L72)):*
> ```XML
>         height: BlockNumber,
>         timeout: Duration,
>         content: mpsc::Receiver<Self::ProposalChunk>,
> ```

I see that you are explicitly using the protobuf `ProposalPart` throughout. I do not want this to be the case. What I want is a type which is generic and implements `Into/From ProposalInit/ProposalFin`.  Let's rename `ProposalChunk` here to `ProposalPart` actually. Nowhere in the `papyrus_consensus` crate should the actual protobuf type `ProposalPart` appear, since the crate should remain generic on the proposal's content.



_Code quote:_
```XML
Self::ProposalChunk
```
___
*[`crates/sequencing/papyrus_consensus_orchestrator/src/sequencer_consensus_context.rs` line 139 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1609#-OARszvn5uY0IEf4ABhK:-OARszvn5uY0IEf4ABhL:b-8ubr62) ([raw file](https://github.com/starkware-libs/sequencer/blob/93d5db27630b5633f298caa61222d82f1a20e4ae/crates/sequencing/papyrus_consensus_orchestrator/src/sequencer_consensus_context.rs#L139)):*
> ```XML
>         height: BlockNumber,
>         timeout: Duration,
>         content_receiver: mpsc::Receiver<ProposalPart>,
> ```

Only here in this implementation should we actually reference the protobuf ProposalPart struct, instead of being generic.



_Code quote:_
```XML
      content_receiver: mpsc::Receiver<ProposalPart>,
```
___
*[`crates/sequencing/papyrus_consensus_orchestrator/src/sequencer_consensus_context.rs` line 342 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1609#-OARtHID80QynLtzps2V:-OARtHID80QynLtzps2W:b-b2l29s) ([raw file](https://github.com/starkware-libs/sequencer/blob/93d5db27630b5633f298caa61222d82f1a20e4ae/crates/sequencing/papyrus_consensus_orchestrator/src/sequencer_consensus_context.rs#L342)):*
> ```XML
>                 }
>             }
>             ProposalPart::Fin(ProposalFin { proposal_content_id: id }) => {
> ```

let's break out of the loop here and move this code out of the loop. And then we can explicitly check that the receiver is closed with no other content.



_Code quote:_
```XML
            ProposalPart::Fin(ProposalFin { proposal_content_id: id }) => {
```

PR #1674, opened 2024-10-29
chore(sequencer_node): re-apply set chain_id as required pointer param

Itay-Tsabary-Starkware:


nadin-Starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1674)* status: 0 of 5 files reviewed, 2 unresolved discussions (waiting on @Itay-Tsabary-Starkware)

___
*[`crates/mempool_node/src/utils.rs` line 21 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1674#-OAR0-Iw7T86IaYZ_WYX:-OAR0-Iw7T86IaYZ_WYY:bak6c5m) ([raw file](https://github.com/starkware-libs/sequencer/blob/4293a67d9205f74fe5ace5b582941b80764c4c7d/crates/mempool_node/src/utils.rs#L21)):*
> ```XML
> 
> /// Returns the absolute path from the project root.
> pub fn get_absolute_path(relative_path: &str) -> PathBuf {
> ```

This function appears multiple times in the code, I think we should add a TODO to consolidate them into a single location.



_Code quote:_
```XML
pub fn get_absolute_path(relative_path: &str) -> PathBuf 
```
___
*[`crates/mempool_node/src/config/component_config.rs` line 46 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1674#-OAR06yA0n0wA4YMeJjM:-OAR06yA0n0wA4YMeJjN:brk2o17) ([raw file](https://github.com/starkware-libs/sequencer/blob/7a4efb301d804b5b7e911f292a1d1112ab6de2c3/crates/mempool_node/src/config/component_config.rs#L46)):*
> ```XML
>     fn dump(&self) -> BTreeMap<ParamPath, SerializedParam> {
>         #[allow(unused_mut)]
>         let mut sub_configs = vec![
> ```

Why did we need this initially, and why are we removing it now?



_Code quote:_
```XML
        #[allow(unused_mut)]
        let mut sub_configs = vec![
```

PR #1783, opened 2024-11-04
chore: add lints to papyrus_common

giladchase:

Nearly all `as` uses can be replaced with `[Try]From` which doesn't have implicit coercions like `as` (we encountered several bugs due to these coersions).

Motivation: we are standardizing lints across the repo and CI, instead of each crate having separate sets of lints.

### DIFF
--- a/crates/batcher_types/Cargo.toml
+++ b/crates/batcher_types/Cargo.toml
@@ -5,7 +5,6 @@ edition.workspace = true
 license.workspace = true
 repository.workspace = true
 
-
 [lints]
 workspace = true
 
--- a/crates/consensus_manager/Cargo.toml
+++ b/crates/consensus_manager/Cargo.toml
@@ -5,7 +5,6 @@ edition.workspace = true
 license.workspace = true
 repository.workspace = true
 
-
 [lints]
 workspace = true
 
--- a/crates/mempool_test_utils/Cargo.toml
+++ b/crates/mempool_test_utils/Cargo.toml
@@ -5,7 +5,8 @@ edition.workspace = true
 repository.workspace = true
 license.workspace = true
 
-[features]
+[lints]
+workspace = true
 
 [dependencies]
 assert_matches.workspace = true
--- a/crates/mempool_test_utils/src/starknet_api_test_utils.rs
+++ b/crates/mempool_test_utils/src/starknet_api_test_utils.rs
@@ -275,7 +275,7 @@ impl MultiAccountTransactionGenerator {
 }
 
 /// Manages transaction generation for a single account.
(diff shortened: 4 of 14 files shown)


dorimedini-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1783)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase and @ShahakShama)

___
*[`crates/papyrus_common/src/python_json.rs` line 46 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1783#-OArEx8A4YttImvOjQZZ:-OArEx8A4YttImvOjQZ_:b-e8cuuv) ([raw file](https://github.com/starkware-libs/sequencer/blob/3795db56e5fafb19fc51e3966acf059901669d74/crates/papyrus_common/src/python_json.rs#L46)):*
> ```XML
>                     write!(writer, r"\u{:4x}", num)?;
>                 }
>             }
> ```

is this equivalent...? non-blocking



_Suggestion:_
```XML
            match u8::try_from(ch) {
                Ok(ascii) => writer.write_all(&[ascii])?,
                Err(_) => {
                let slice = ch.encode_utf16(&mut buf);
                for num in slice {
                    write!(writer, r"\u{:4x}", num)?;
                }
                }
            }
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1783)* status: all files reviewed, 1 unresolved discussion (waiting on @ShahakShama)

___
*[`crates/papyrus_common/src/python_json.rs` line 46 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1783#-OArEx8A4YttImvOjQZZ:-OArURTj1FXhqrN_Q3Bh:bif8awl) ([raw file](https://github.com/starkware-libs/sequencer/blob/3795db56e5fafb19fc51e3966acf059901669d74/crates/papyrus_common/src/python_json.rs#L46)):*

is this equivalent...? non-blocking
</blockquote></details>

`is_ascii` checks if 0<= char <= 127, whereas u8::try_from will succeed up to 255.
So, for example, for 128, in the original code, it'll encode it as utf16 (with `\uX` escaped chars?), but in the proposed code, it will write it as is to the writer.
This could change expected output wherever this is used (it's related to `Display`).

I'm not sure what was the original intention though without delving into this code (which is out of scope) unless @ShahakShama chimes in.

dorimedini-starkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1783#-:-OArYGT2BkY1c0RC45P9:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ShahakShama)

PR #1841, opened 2024-11-05
chore: add lint to p2p_sync

giladchase:

Nearly all as uses can be replaced with [Try]From which doesn't have implicit coercions like as (we encountered several bugs due to these coercions).

Motivation: we are standardizing lints across the repo and CI, instead of each crate having separate sets of lints.

Changes:
1. header/state_diff: added clippy ignore to `gauge!` metric usage which requires f64. This is probably unnecessary since metrics used are int values, but fixing this is out of scope, and there are no other sensible conversions from unsigned to floats.
2. test.rs: most usages of these two vars were for usize, and only a couple needed u64, so i flipped the usages.


dorimedini-starkware (approved):
:lgtm:

> Reviewed 2 of 4 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1841#-:-OB4vZD61lLQHnleEqDc:bnfp4nl)* status: 3 of 4 files reviewed, all discussions resolved (waiting on @elintul)

PR #1842, opened 2024-11-05
chore: code dedup in blockifier transaction from api tx

ArniStarkware:


ArniStarkware:
Already covered in the main PR.

PR #1845, opened 2024-11-05
chore: add lints to protobuf

giladchase:

Changes: use try_from instead of relying on implicit truncation due of `u128 as u64`.


dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1845)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase and @ShahakShama)

___
*[`crates/papyrus_protobuf/src/converters/common.rs` line 58 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1845#-OB-gM-P83tezewDHpfL:-OB-gM-P83tezewDHpfM:by4o6qo) ([raw file](https://github.com/starkware-libs/sequencer/blob/83f3ca6647bde052211d1b368546675a934631a8/crates/papyrus_protobuf/src/converters/common.rs#L58)):*
> ```XML
>         Self {
>             high: u64::try_from(value >> 64).unwrap(),
>             low: u64::try_from(value & u128::from(u64::MAX)).unwrap(),
> ```

for future, if we can ban `unwrap` in non-test code



_Suggestion:_
```XML
            high: u64::try_from(value >> 64).expect("64-bit shift of u128 is a u64."),
            low: u64::try_from(value & u128::from(u64::MAX)).expect("Bottom 64 bits of u128 is a u64."),
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1845)* status: 2 of 3 files reviewed, 1 unresolved discussion (waiting on @dorimedini-starkware and @ShahakShama)

___
*[`crates/papyrus_protobuf/src/converters/common.rs` line 58 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1845#-OB-gM-P83tezewDHpfL:-OB4hmKn4WkfH3U62Xr2:bpmym2h) ([raw file](https://github.com/starkware-libs/sequencer/blob/83f3ca6647bde052211d1b368546675a934631a8/crates/papyrus_protobuf/src/converters/common.rs#L58)):*

for future, if we can ban `unwrap` in non-test code
</blockquote></details>

Added the `expect`s (changed `is` to `are`, I think that is more correct?).

But for the recond, i think that this is an OK example of why `unwrap`s are sometimes better than `expect`.
Without lighting any torches, especialy for the first example, was the code unclear with the unwrap? Not asking rethorically, I think that unwrap implies that this cast is trivial, whereas the expect implies deeper reasoning but ends up repeating the same thing the code said.

With that said, yr call.

dorimedini-starkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1845#-:-OB4uhd5Dn6eknROBbfn:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ShahakShama)

___
*[`crates/papyrus_protobuf/src/converters/common.rs` line 58 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1845#-OB-gM-P83tezewDHpfL:-OB4uXoc6b2hUQJ2mtE6:bncj91g) ([raw file](https://github.com/starkware-libs/sequencer/blob/83f3ca6647bde052211d1b368546675a934631a8/crates/papyrus_protobuf/src/converters/common.rs#L58)):*

Added the `expect`s (changed `is` to `are`, I think that is more correct?).

But for the recond, i think that this is an OK example of why `unwrap`s are sometimes better than `expect`.
Without lighting any torches, especialy for the first example, was the code unclear with the unwrap? Not asking rethorically, I think that unwrap implies that this cast is trivial, whereas the expect implies deeper reasoning but ends up repeating the same thing the code said.

With that said, yr call.
</blockquote></details>

You're right, it's clear from code, but easier to enforce `expects` and occasionally be overly verbose than allowing cryptic `unwrap`s

PR #1852, opened 2024-11-05
chore: add lints to rpc

giladchase:

Nearly all as uses can be replaced with [Try]From which doesn't have implicit 
coercions like as (we encountered several bugs due to these coercions).

Motivation: we are standardizing lints across the repo and CI, instead of each 
crate having separate sets of lints.

### DIFF
--- a/crates/papyrus_rpc/Cargo.toml
+++ b/crates/papyrus_rpc/Cargo.toml
@@ -62,7 +62,5 @@ starknet_client = { workspace = true, features = ["testing"] }
 strum.workspace = true
 strum_macros.workspace = true
 
-[lints.rust]
-# See [here](https://github.com/taiki-e/cargo-llvm-cov/issues/370) for a discussion on why this is
-# needed (from rust 1.80).
-unexpected_cfgs = { level = "warn", check-cfg = ['cfg(coverage_nightly)'] }
+[lints]
+workspace = true
--- a/crates/papyrus_rpc/src/v0_8/api/test.rs
+++ b/crates/papyrus_rpc/src/v0_8/api/test.rs
@@ -2811,7 +2811,7 @@ async fn test_get_events(
     let mut parent_hash = BlockHash(felt!(GENESIS_HASH));
     let mut rw_txn = storage_writer.begin_rw_txn().unwrap();
     for (i, block_metadata) in block_metadatas.iter().enumerate() {
-        let block_number = BlockNumber(i as u64);
+        let block_number = BlockNumber(u64::try_from(i).expect("usize should fit in u64"));
         let block = block_metadata.generate_block(&mut rng, parent_hash, block_number);
 
         parent_hash = block.header.block_hash;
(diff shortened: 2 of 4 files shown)


dorimedini-starkware (changes requested):
> Reviewed 4 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1852)* status: all files reviewed, 1 unresolved discussion (waiting on @dan-starkware and @giladchase)

___
*[`crates/papyrus_rpc/src/v0_8/transaction.rs` line 1240 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1852#-OB-pSbb21U7357oqFHV:-OB-pSbb21U7357oqFHW:b-yceb20) ([raw file](https://github.com/starkware-libs/sequencer/blob/62c4ff87ff56d3fa60160a6fda472d33bc71ceab/crates/papyrus_rpc/src/v0_8/transaction.rs#L1240)):*
> ```XML
>     let selector = Token::Bytes(entry_point_selector.0.to_bytes_be().to_vec());
>     let payload_length_as_felt =
>         Felt::from(u64::try_from(payload.len()).expect("usize should fit in usize"));
> ```

or reword the `expect` (usize should fit in u64)



_Suggestion:_
```XML
u64_from_usize(payload.len())
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1852)* status: all files reviewed, 1 unresolved discussion (waiting on @dan-starkware and @dorimedini-starkware)

___
*[`crates/papyrus_rpc/src/v0_8/transaction.rs` line 1240 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/1852#-OB-pSbb21U7357oqFHV:-OB4ptAU2_uol6y0grgK:bmb1z2z) ([raw file](https://github.com/starkware-libs/sequencer/blob/62c4ff87ff56d3fa60160a6fda472d33bc71ceab/crates/papyrus_rpc/src/v0_8/transaction.rs#L1240)):*

or reword the `expect` (usize should fit in u64)
</blockquote></details>

Whoops thanks.
Reworded the `expect`, would have loved to use the util but this hits a different issue:

a) this util is from the blockifier, which isn't a dependency here (and adding it just for that is unreasonable)
b) we can extract it from the blockifier (it isn't blockifier-specific), but we don't have a natural place to put this util ATM. There were talks about `utils` crate but nothing settled yet (I personally don't like a global `utils` crate, it's too loosely goosey and gets messy.
Maybe we can have a `stdx` crate (std extensions), which will have things that _could_ have been part of the std. rust-analyzer have this. Anyway, out of scope here probably.

dorimedini-starkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1852#-:-OB4uz6c3tasQazaIbI7:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dan-starkware)

PR #1936, opened 2024-11-11
refactor(mempool): change with_priority_queue input to Transaction iterator

noam-starkware:
In order to encapsulate `TransactionReference` from mempool tests, change the input of `with_priority_queue` from a `TransactionReference` iterator to a `&Transaction` iterator.

The execution of `with_priority_queue` maps each `&Transaction` to a `TransactionReference`.

Some tests still require cloning of transactions due to the fact that `with_pool` uses `Transaction` iterators instead of `&Transactions`. See comments for specific tests in the PR.

We propose modifying `with_pool` and `with_pending_queue` to accept  `&Transaction` as well. This makes sense for two reasons:
1. The `Transaction` objects passed to the functions `with_[pool/priority_queue/pending_queue]` are not modified.
2. Using references instead of objects will allow us to omit cloning of `Transaction` objects inside tests allowing for more readable code.

### DIFF
--- a/crates/mempool/src/mempool_test.rs
+++ b/crates/mempool/src/mempool_test.rs
@@ -78,11 +78,13 @@ impl MempoolContentBuilder {
         self
     }
 
-    fn with_priority_queue<Q>(mut self, queue_txs: Q) -> Self
+    fn with_priority_queue<'a, Q>(mut self, queue_txs: Q) -> Self
     where
-        Q: IntoIterator<Item = TransactionReference>,
+        Q: IntoIterator<Item = &'a Transaction>,
     {
-        self.tx_queue_content_builder = self.tx_queue_content_builder.with_priority(queue_txs);
+        self.tx_queue_content_builder = self
+            .tx_queue_content_builder
+            .with_priority(queue_txs.into_iter().map(TransactionReference::new));
         self
     }
 
@@ -189,7 +191,7 @@ fn test_get_txs_returns_by_priority_order(#[case] n_requested_txs: usize) {
 
     let mut mempool = MempoolContentBuilder::new()
         .with_pool(txs.iter().cloned())
-        .with_priority_queue(txs.iter().map(TransactionReference::new))
+        .with_priority_queue(&txs)
         .build_into_mempool();
 
     // Test.
@@ -201,9 +203,7 @@ fn test_get_txs_returns_by_priority_order(#[case] n_requested_txs: usize) {
     assert_eq!(fetched_txs, expected_fetched_txs);
 
     // Assert: non-returned transactions are still in the mempool.
-    let remaining_tx_references = remaining_txs.iter().map(TransactionReference::new);
-    let mempool_content =
-        MempoolContentBuilder::new().with_priority_queue(remaining_tx_references).build();
+    let mempool_content = MempoolContentBuilder::new().with_priority_queue(remaining_txs).build();
     mempool_content.assert_eq(&mempool);
 }
 
@@ -226,8 +226,8 @@ fn test_get_txs_does_not_remove_returned_txs_from_pool() {
     // Setup.
     let tx = tx!();
 
-    let queue_txs = [TransactionReference::new(&tx)];
-    let pool_txs = [tx];
+    let queue_txs = [&tx];
+    let pool_txs = [tx.clone()];
     let mut mempool = MempoolContentBuilder::new()
         .with_pool(pool_txs.clone())
         .with_priority_queue(queue_txs)
@@ -247,7 +247,7 @@ fn test_get_txs_replenishes_queue_only_between_chunks() {
     let tx_address_0_nonce_1 = tx!(tx_hash: 2, address: "0x0", tx_nonce: 1, tip: 20);
     let tx_address_1_nonce_0 = tx!(tx_hash: 3, address: "0x1", tx_nonce: 0, tip: 10);
 
-    let queue_txs = [&tx_address_0_nonce_0, &tx_address_1_nonce_0].map(TransactionReference::new);
+    let queue_txs = [&tx_address_0_nonce_0, &tx_address_1_nonce_0];
     let pool_txs =
         [&tx_address_0_nonce_0, &tx_address_0_nonce_1, &tx_address_1_nonce_0].map(|tx| tx.clone());
     let mut mempool = MempoolContentBuilder::new()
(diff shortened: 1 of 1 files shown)


noam-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1936)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ayeletstarkware)

___
*[`crates/mempool/src/mempool_test.rs` line 230 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/1936#-OBV2-aO-TY8mrjLoWLi:-OBV2-aO-TY8mrjLoWLj:b-2gnbyk) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6b5aaac2d94e12ac4993341222943f5158984d3/crates/mempool/src/mempool_test.rs#L230)):*
> ```XML
> 
>     let queue_txs = [&tx];
>     let pool_txs = [tx.clone()];
> ```

This `clone` is an example of one which can be removed once `with_pool` is defined to work with `&Transaction` iterator. It is currently required because of the borrow action in the previous line. Were this a reference as well, cloning would not be needed as multiple immutable borrows can exist at the same time.



_Code quote:_
```XML
let pool_txs = [tx.clone()];
```

ayeletstarkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1936#-:-OBVNGKeC9JPkeUMPUYN:bnfp4nl)* status: all files reviewed, 1 unresolved discussion

elintul (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1936)* status: all files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware and @noam-starkware)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/1936#-OBVR6NB2M0oIeVIx0oE:-OBVR6NCDqcnllZ660D6:b6gulad) (no related file):*
Any chance to have the additional `clone` in a single location? For example, implement `FromIterator<Transaction> for TransactionQueue`.

elintul (commented):
> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1936)* status: all files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware and @noam-starkware)

noam-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1936)* status: all files reviewed, 2 unresolved discussions (waiting on @ayeletstarkware and @elintul)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/1936#-OBVR6NB2M0oIeVIx0oE:-OBVdWAS3MjGciWhPmfY:b-gqv1w0) (no related file):*

Any chance to have the additional `clone` in a single location? For example, implement `FromIterator<Transaction> for TransactionQueue`.
</blockquote></details>

I'm not sure if I understand correctly, right now there are multiple clones because they are split among multiple tests. I'm not sure how we can technically group all these clones together, or have I misunderstood your question?

I just want to add that if we modify `with_pool` and `with_pending_queue` to work with `&Transaction` iterators similar to `with_priority_queue` then there's no need to perform any cloning inside the tests, and the only cloning would be inside `with_pool` :

```
    fn with_pool<'a, P>(mut self, pool_txs: P) -> Self
    where
        P: IntoIterator<Item = &'a Transaction>,
    {
        self.tx_pool = Some(pool_txs.into_iter().cloned().collect());
        self
    }
```

Is this what you were aiming for?

elintul (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1936)* status: all files reviewed, 3 unresolved discussions (waiting on @ayeletstarkware and @noam-starkware)

___
*[`crates/mempool/src/mempool_test.rs` line 83 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/1936#-OBZt_cJ0L63x2HXQeOA:-OBZt_cJ0L63x2HXQeOB:b-hhldve) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6b5aaac2d94e12ac4993341222943f5158984d3/crates/mempool/src/mempool_test.rs#L83)):*
> ```XML
>     fn with_priority_queue<'a, Q>(mut self, queue_txs: Q) -> Self
>     where
>         Q: IntoIterator<Item = &'a Transaction>,
> ```

Would this work?
Please read the chapter on iterators.



_Suggestion:_
```XML
Iterator<Item = Transaction>
```

noam-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/1936)* status: all files reviewed, 3 unresolved discussions (waiting on @ayeletstarkware and @elintul)

___
*[`crates/mempool/src/mempool_test.rs` line 83 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/1936#-OBZt_cJ0L63x2HXQeOA:-OB_RY1t9VlBwLSZ0FxK:bik780d) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6b5aaac2d94e12ac4993341222943f5158984d3/crates/mempool/src/mempool_test.rs#L83)):*

Would this work?
Please read the chapter on iterators.
</blockquote></details>

If I understand correctly, these are two independent changes: changing the trait from `IntoIterator` to `Iterator` and changing the iterated object to a `Transaction` instead of `&Transaction`.

If we implement these changes, whenever we call `with_priority_queue` we will need to supply it with an iterator (instead of an iterable object), for example:

```
let mut mempool = MempoolContentBuilder::new()
        .with_priority_queue(txs.into_iter())
        .build_into_mempool();
```

This, however will consume `txs` and so in tests where we need to keep it alive we will need to supply a cloned iterator, like this:

```
    let mut mempool = MempoolContentBuilder::new()
        .with_priority_queue(txs.clone().into_iter()) // into_iter is required to return owned values and thus consumes txs
        .build_into_mempool();
    ...
    foo(txs) // Alive, since we cloned txs in the previous call
```

And so this implementation will still require us to perform multiple clones inside tests.

Another option is to change the trait to `Iterator` but keep the iterated type `&Transaction`. In such a case we will still need to convert `txs` to an iterator before invoking the function but in this case it will not be consumed. 
This happens because Rust implements `[&T].into_iter()` via an iterator that borrows a reference to each object in the iterable and thus doesn't take ownership and consumes the underlying iterable:

```
    let mut mempool = MempoolContentBuilder::new()
        .with_pool(txs.clone())
        .with_priority_queue(txs.iter()) // iter returns immutable references and thus doesn't consume txs
        .build_into_mempool();
    foo(txs) // Alive, since txs was not consumed in the previous call
```

The Rust Book details the differences between `iter()` and `into_iter()` [here](https://doc.rust-lang.org/book/ch13-02-iterators.html#:~:text=Also%20note%20that%20the%20values,iter_mut%20instead%20of%20iter.).

PR #2026, opened 2024-11-13
chore: share the test util get_absolute_path

ArniStarkware:


Itay-Tsabary-Starkware (approved):
:lgtm:

> Reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2026#-:-OBe9AXJ3Ga2CZF64uEn:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ArniStarkware)

___
*[`crates/papyrus_config/Cargo.toml` line 26 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2026#-OBe5uBA5kFTlRZmyCFP:-OBe8OaQCZEPELSccJnd:b-gewq10) ([raw file](https://github.com/starkware-libs/sequencer/blob/88a733d7f2334ea91e0033ae0607770c11d42c8f/crates/papyrus_config/Cargo.toml#L26)):*
<details><summary><i>Previously, ArniStarkware (Arnon Hod) wrote…</i></summary><blockquote>

@Itay-Tsabary-Starkware - Correct me if I am wrong, but you are now in charge of this crate

As discussed, starknet_api should be at the bottom of crate hierarchies. The config crate should also be low in the hierarchy, which might be problematic if we have an inheritance here.
</blockquote></details>

Right. IMO this is accetpable for the time being, but the more thought I put into this the more it seems we need this in a desginated util crate with this function as production code. You can see there are plenty of instances where the code opens files using relative paths in an unchecked manner.

PR #2038, opened 2024-11-14
fix(blockifier_reexecution): add additonal expected rpc error

aner-starkware:


dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2038)* status: all files reviewed, 1 unresolved discussion (waiting on @aner-starkware and @AvivYossef-starkware)

___
*[`crates/blockifier_reexecution/src/state_reader/test_state_reader.rs` line 55 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2038#-OBdf7CEBJ_OOgbyNeDw:-OBdf7CEBJ_OOgbyNeDx:b-scrvcq) ([raw file](https://github.com/starkware-libs/sequencer/blob/e9b61cefcf10b1495c82ac4ee73e4fcdb550e117/crates/blockifier_reexecution/src/state_reader/test_state_reader.rs#L55)):*
> ```XML
> pub const DEFAULT_RETRY_WAIT_TIME: u64 = 10000;
> pub const DEFAULT_EXPECTED_ERROR_STRINGS: [&str; 4] =
>     ["Connection error", "RPCError", "429 Too Many Requests", "error decoding response body"];
> ```

no - this should not be whitelisted.
this was the same error I got when sending a get_contract_class request with a decimal class hash instead of a hex string.
retry shouldn't help in this case.
if you get this error, inject print statements in the gateway RPC client code to see the actual RPC response *before trying to parse it into an `RpcResponse` variant*



_Code quote:_
```XML
"error decoding response body"
```

aner-starkware:
*[`crates/blockifier_reexecution/src/state_reader/test_state_reader.rs` line 55 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2038#-OBdf7CEBJ_OOgbyNeDw:-OBdxJQNATnhkS2Z7IK7:b-kw4myp) ([raw file](https://github.com/starkware-libs/sequencer/blob/e9b61cefcf10b1495c82ac4ee73e4fcdb550e117/crates/blockifier_reexecution/src/state_reader/test_state_reader.rs#L55)):*

no - this should not be whitelisted.
this was the same error I got when sending a get_contract_class request with a decimal class hash instead of a hex string.
retry shouldn't help in this case.
if you get this error, inject print statements in the gateway RPC client code to see the actual RPC response *before trying to parse it into an `RpcResponse` variant*
</blockquote></details>

Why not retry? This error can happen randomly as well, as we saw, so without retrying, it will randomly fail some tests. The error is printed out as well, so we could still debug if the problem persists.

aner-starkware:
*[`crates/blockifier_reexecution/src/state_reader/test_state_reader.rs` line 55 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2038#-OBdf7CEBJ_OOgbyNeDw:-OBdyJJ7AcKCdtajT9As:b-n0mgor) ([raw file](https://github.com/starkware-libs/sequencer/blob/e9b61cefcf10b1495c82ac4ee73e4fcdb550e117/crates/blockifier_reexecution/src/state_reader/test_state_reader.rs#L55)):*

Why not retry? This error can happen randomly as well, as we saw, so without retrying, it will randomly fail some tests. The error is printed out as well, so we could still debug if the problem persists.
</blockquote></details>

The main thing is that it's not an error that we should return, so we should `panic` if we get it (that's what "expected" means here)

aner-starkware:
*[`crates/blockifier_reexecution/src/state_reader/test_state_reader.rs` line 55 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2038#-OBdf7CEBJ_OOgbyNeDw:-OBdyij8BLGfMqOBDuZc:b-n878a2) ([raw file](https://github.com/starkware-libs/sequencer/blob/e9b61cefcf10b1495c82ac4ee73e4fcdb550e117/crates/blockifier_reexecution/src/state_reader/test_state_reader.rs#L55)):*

The main thing is that it's not an error that we should return, so we should `panic` if we get it (that's what "expected" means here)
</blockquote></details>

When you think about it, the way it works now is actually more "black-listing" than "white-listing"  I think it's better to switch to white-listing (added a TODO)

dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2038)* status: all files reviewed, 1 unresolved discussion (waiting on @aner-starkware and @AvivYossef-starkware)

___
*[`crates/blockifier_reexecution/src/state_reader/test_state_reader.rs` line 55 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2038#-OBdf7CEBJ_OOgbyNeDw:-OBe9EctByiH1Am3Bf23:bs0u9u0) ([raw file](https://github.com/starkware-libs/sequencer/blob/e9b61cefcf10b1495c82ac4ee73e4fcdb550e117/crates/blockifier_reexecution/src/state_reader/test_state_reader.rs#L55)):*
> Why not retry? This error can happen randomly as well, as we saw, so without retrying

this needs to be reproduced with extra data from the gateway. I'll see if I can make it easier to see the source of these errors, will try to get to it today, but for now - inject prints before trying to call `serde::from::<RpcResponse>` (RPC client does this) to see what the RPC server is complaining about.
We do not expect a retry to help here - if it does, we need to find out why

> The error is printed out as well, so we could still debug if the problem persists.

I am not sure this is true - I only saw this uninformative string when fixing the decimal/hex issue. I had to inject prints to see the source of the error

> The main thing is that it's not an error that we should return, so we should `panic` if we get it (that's what "expected" means here)

we need to separate these two concepts:
1. errors we expect to sporadically encounter when communicating with a server (429, connection issues...)
2. errors we want to cause panic

since the source of the error isn't always printed out on this panic, I think we should treat this list as type (1), not type (2). We can add a `panic_on_recurring_expected_errors` boolean field to `RetryConfig` and set it to `true`, but if you want a separate list to panic for, that can also work.

(this also relates to your comment on whitelisting/blacklisting)

dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2038)* status: all files reviewed, 1 unresolved discussion (waiting on @aner-starkware and @AvivYossef-starkware)

___
*[`crates/blockifier_reexecution/src/state_reader/test_state_reader.rs` line 55 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2038#-OBdf7CEBJ_OOgbyNeDw:-OBeZOCc2bRd3X5AZs68:b3f0hj) ([raw file](https://github.com/starkware-libs/sequencer/blob/e9b61cefcf10b1495c82ac4ee73e4fcdb550e117/crates/blockifier_reexecution/src/state_reader/test_state_reader.rs#L55)):*

Regarding the printout - I added it in this PR; that's why you didn't see it (before, it printed only when it `panic!`ed, not on errors it returned).
I know retrying on this helps, I don't know why, but in any case it's better than the current situation where this error is just returned. 
If you want me to change to just `panic` without retrying, I will, but I don't understand why, since we already saw that retrying can help in this case.
</blockquote></details>

[here](https://reviewable.io/reviews/starkware-libs/sequencer/2062) is what I am proposing as a solution instead of this PR

PR #2145, opened 2024-11-18
chore(l1_provider): add validate state transition

giladchase:
Also minor fixes:
- also remove unnecessary `#[macro_export]` for local macro
- use longer path for errors instead of importing the variant.

### DIFF
--- a/crates/starknet_l1_provider/src/l1_provider_tests.rs
+++ b/crates/starknet_l1_provider/src/l1_provider_tests.rs
@@ -8,10 +8,8 @@ use starknet_api::test_utils::l1_handler::executable_l1_handler_tx;
 use starknet_api::transaction::TransactionHash;
 
 use crate::errors::L1ProviderError;
-use crate::errors::L1ProviderError::UnexpectedProviderStateTransition;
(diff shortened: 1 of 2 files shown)


giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2145)* status: all files reviewed, 1 unresolved discussion (waiting on @ayeletstarkware and @elintul)

___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 66 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2145#-OBzhGdPEHe3BPTB0EXW:-OC2U13-ECbIKei1RxXC:b-gatjdl) ([raw file](https://github.com/starkware-libs/sequencer/blob/79251600dad4ffebd5f871ef7ea98907c2c7cd48/crates/starknet_l1_provider/src/l1_provider_tests.rs#L66)):*

Plan is to transition to "pending" between each (validator/proposer) role?
</blockquote></details>

Yep.
For example, commit_block should trigger that automatically for now, until we get to the optimisitc-proposer flow which will be more delicate.

PR #2195, opened 2024-11-20
chore: remove redundant crate dependency

Itay-Tsabary-Starkware:


lev-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2195)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @Itay-Tsabary-Starkware)

___
*[`crates/starknet_monitoring_endpoint/src/monitoring_endpoint_test.rs` line 29 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2195#-OCS9p1C1W53JzRh9yOP:-OCS9p1C1W53JzRh9yOQ:bvxkav5) ([raw file](https://github.com/starkware-libs/sequencer/blob/183268b7e4249497649b847566e0a0302f46a1dc/crates/starknet_monitoring_endpoint/src/monitoring_endpoint_test.rs#L29)):*
> ```XML
> }
> 
> async fn request_app(app: Router, method: &str) -> Response {
> ```

Actually, the type axum::response::Response is by default Response<UnsyncBoxBody<Bytes, Error>>. So dependency is not removed but the axum brings it with it. 
I think that preferable to use the explicit type and not implicit by default.

Itay-Tsabary-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2195)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @lev-starkware)

___
*[`crates/starknet_monitoring_endpoint/src/monitoring_endpoint_test.rs` line 29 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2195#-OCS9p1C1W53JzRh9yOP:-OCT8e3xAgZNI78ia1J4:bierwhq) ([raw file](https://github.com/starkware-libs/sequencer/blob/183268b7e4249497649b847566e0a0302f46a1dc/crates/starknet_monitoring_endpoint/src/monitoring_endpoint_test.rs#L29)):*
<details><summary><i>Previously, lev-starkware wrote…</i></summary><blockquote>

Actually, the type axum::response::Response is by default Response<UnsyncBoxBody<Bytes, Error>>. So dependency is not removed but the axum brings it with it. 
I think that preferable to use the explicit type and not implicit by default.
</blockquote></details>

1) Could you please point me to the docs that state that?
2) I don't see the benefit of using these internal types: we get the same effect without them with the benefit of less code complexity + less direct imports. If internally they are translated to whatever data structure, then let that translated remain internal.

lev-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2195)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @Itay-Tsabary-Starkware)

___
*[`crates/starknet_monitoring_endpoint/src/monitoring_endpoint_test.rs` line 29 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2195#-OCS9p1C1W53JzRh9yOP:-OCTbPjs4ZrVpPYOB297:b-q603ef) ([raw file](https://github.com/starkware-libs/sequencer/blob/183268b7e4249497649b847566e0a0302f46a1dc/crates/starknet_monitoring_endpoint/src/monitoring_endpoint_test.rs#L29)):*

1) Could you please point me to the docs that state that?
2) I don't see the benefit of using these internal types: we get the same effect without them with the benefit of less code complexity + less direct imports. If internally they are translated to whatever data structure, then let that translated remain internal.
</blockquote></details>

1. See code snippet.
2. This is a philosophical debate: I like to see what I have in Response and you don't.
3. But in any case the http_body crate is compiled and linked with this crate.



_Code snippet:_
```XML
/// Type alias for [`http::Response`] whose body type defaults to [`BoxBody`], the most common body
/// type used with axum.
pub type Response<T = BoxBody> = http::Response<T>;
---
/// This is used in axum as the response body type for applications. It's
/// necessary to unify multiple response bodies types into one.
pub type BoxBody = http_body::combinators::UnsyncBoxBody<Bytes, Error>;
```

Itay-Tsabary-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2195)* status: 0 of 6 files reviewed, 1 unresolved discussion (waiting on @lev-starkware)

___
*[`crates/starknet_monitoring_endpoint/src/monitoring_endpoint_test.rs` line 29 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2195#-OCS9p1C1W53JzRh9yOP:-OCUBv5ZBciq9JobkrrH:bqi89er) ([raw file](https://github.com/starkware-libs/sequencer/blob/183268b7e4249497649b847566e0a0302f46a1dc/crates/starknet_monitoring_endpoint/src/monitoring_endpoint_test.rs#L29)):*
<details><summary><i>Previously, lev-starkware wrote…</i></summary><blockquote>

1. See code snippet.
2. This is a philosophical debate: I like to see what I have in Response and you don't.
3. But in any case the http_body crate is compiled and linked with this crate.
</blockquote></details>

Ack, thanks for the pointers. 

Summary afaiu:
1. The previous code state unnecessarily bundles our implementation with axum's internal implementation, i.e., tomorrow can there be a new default, we don't need to know about this, and tbh shouldn't care.
2. The suggested change removes this bundling by falling to use the implicit default values, which are the same, hence there is no actual functionality change.

Do you agree with the above?

IMO, decoupling is strictly preferred over explicitness here, and I'd like to proceed with this change. Thanks :pray:

lev-starkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2195#-:-OCbY5N87HugOoN8e0L9:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Itay-Tsabary-Starkware)

PR #2362, opened 2024-11-28
Save execution time in Call Info

JulianGCalderon:
Possible approach for benchmarking contract execution, without much overhead.


JulianGCalderon:
My mistake. I didn't meant to create this issue.

PR #2464, opened 2024-12-05
fix(cairo_native): remove native FC folder

meship-starkware:


meship-starkware (commented):
> Reviewed 4 of 7 files at r1.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: 4 of 7 files reviewed, all discussions resolved (waiting on @dorimedini-starkware and @Yoni-Starkware)

meship-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: 4 of 7 files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/test_utils/contracts.rs` line 118 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODKUx6F3kuvPrIE7Oqq:-ODKUx6F3kuvPrIE7Oqr:blk6gck) ([raw file](https://github.com/starkware-libs/sequencer/blob/3e035be9e34bb43f042eae2b157517a9c976bcc7/crates/blockifier/src/test_utils/contracts.rs#L118)):*
> ```XML
>             | Self::TestContract(version)
>             | Self::ERC20(version)
>             | Self::LegacyTestContract(version) => *version,
> ```

This has a problematic Meaning now. Up until now, if you had a version, you had all versions but the legacy contract only has Cairo1, and Native does not have Cairo0



_Code quote:_
```XML
Self::LegacyTestContract(version) =
```
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 160 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODKWq682_QMoSE-j9Bb:-ODKWq682_QMoSE-j9Bc:bkbyeyx) ([raw file](https://github.com/starkware-libs/sequencer/blob/3e035be9e34bb43f042eae2b157517a9c976bcc7/crates/blockifier/src/test_utils/contracts.rs#L160)):*
> ```XML
>             | Self::TestContract(v)
>             | Self::ERC20(v)
>             | Self::LegacyTestContract(v) => *v = version,
> ```

This is risky. We need to find a way to disallow the creation of Cairo0 LeagayContract. Which might be problematic for the all contacts method see below.   



_Code quote:_
```XML
           | Self::LegacyTestContract(v) => *v = version,
```
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 446 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODKVij23n7oXMSK4oU_:-ODKVij23n7oXMSK4oUa:b62o580) ([raw file](https://github.com/starkware-libs/sequencer/blob/3e035be9e34bb43f042eae2b157517a9c976bcc7/crates/blockifier/src/test_utils/contracts.rs#L446)):*
> ```XML
>                 let mut other_contract = contract;
>                 other_contract.set_cairo_version(contract.cairo_version().other());
>                 vec![other_contract].into_iter()
> ```

I can go the direct way here, check for the legacy contract, and remove the has only casm and native, but I prefer the generalization here. This function assumes that there is no special treatment for Native CairoVersion. 
Also, the fact that the default is Cairo0 makes us create a LegacyTestContract(Cairo0), which means nothing. 



_Code quote:_
```XML
            } else if contract.has_only_casm_and_native() {
                let mut other_contract = contract;
                other_contract.set_cairo_version(contract.cairo_version().other());
                vec![other_contract].into_iter()
```

Yoni-Starkware (changes requested):
> Reviewed 5 of 7 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: 5 of 7 files reviewed, 5 unresolved discussions (waiting on @dorimedini-starkware and @meship-starkware)

___
*[`crates/blockifier/src/test_utils/contracts.rs` line 450 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODL3maT6JTki0R7h-1K:-ODL3maT6JTki0R7h-1L:b-gzmyrx) ([raw file](https://github.com/starkware-libs/sequencer/blob/faa8438968d1926637800cb3e9ca7269be47e2e3/crates/blockifier/src/test_utils/contracts.rs#L450)):*
> ```XML
>             }
>         })
>     }
> ```

Let's make it simple. Remove all functions that are used only here, and implement `all_contracts` using a single `match` expression explicitly.
Remove the `.other()` function as well



_Code quote:_
```XML
    pub fn all_contracts() -> impl Iterator<Item = Self> {
        // EnumIter iterates over all variants with Default::default() as the cairo
        // version.
        Self::iter().flat_map(|contract| {
            if contract.has_all_versions() {
                let mut other_contract = contract;
                other_contract.set_cairo_version(contract.cairo_version().other());
                vec![contract, other_contract].into_iter()
            } else if contract.has_only_casm_and_native() {
                let mut other_contract = contract;
                other_contract.set_cairo_version(contract.cairo_version().other());
                vec![other_contract].into_iter()
            } else {
                vec![contract].into_iter()
            }
        })
    }
```

dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: all files reviewed, 7 unresolved discussions (waiting on @meship-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/test_utils/contracts.rs` line 118 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODKUx6F3kuvPrIE7Oqq:-ODLMVxi1E2slccB67n2:b-wwujog) ([raw file](https://github.com/starkware-libs/sequencer/blob/3e035be9e34bb43f042eae2b157517a9c976bcc7/crates/blockifier/src/test_utils/contracts.rs#L118)):*
<details><summary><i>Previously, meship-starkware (Meshi Peled) wrote…</i></summary><blockquote>

This has a problematic Meaning now. Up until now, if you had a version, you had all versions but the legacy contract only has Cairo1, and Native does not have Cairo0
</blockquote></details>

why does the legacy contract now have two cairo1 variants? why not only cairo1 non-native?
IIRC the legacy contract must be compiled with a legacy compiler --> too old for a native compilation; no?
IIUC we can't run native on old cairo1 contracts...
if I'm right, you can simply revert this added `version` data in the legacy variant and all is well again.
IF we will have a contract that doesn't have cairo0 but has two cairo1 versions, then you're right... we'll need the following refactor:
```
enum RunnableCairo1Version {
    Casm,
    Native,
}

enum RunnableVersion {
    Cairo0,
    Cairo1(RunnableCairo1Version),
}

enum FeatureContract {
    SecurityTests,
    TestContract(RunnableVersion),
    OnlyCairo1Contract(RunnableCairo1Version),
    ... etc
}
```
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 160 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODKWq682_QMoSE-j9Bb:-ODLNG4402Ac4d9TkF1n:b5x87a9) ([raw file](https://github.com/starkware-libs/sequencer/blob/3e035be9e34bb43f042eae2b157517a9c976bcc7/crates/blockifier/src/test_utils/contracts.rs#L160)):*
<details><summary><i>Previously, meship-starkware (Meshi Peled) wrote…</i></summary><blockquote>

This is risky. We need to find a way to disallow the creation of Cairo0 LeagayContract. Which might be problematic for the all contacts method see below.
</blockquote></details>

see above: I think we should disallow creation of both cairo0 and native, and then there's no problem
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 446 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODKVij23n7oXMSK4oU_:-ODLNlc63TVxptbL_4XC:bf5i1i1) ([raw file](https://github.com/starkware-libs/sequencer/blob/3e035be9e34bb43f042eae2b157517a9c976bcc7/crates/blockifier/src/test_utils/contracts.rs#L446)):*
<details><summary><i>Previously, meship-starkware (Meshi Peled) wrote…</i></summary><blockquote>

I can go the direct way here, check for the legacy contract, and remove the has only casm and native, but I prefer the generalization here. This function assumes that there is no special treatment for Native CairoVersion. 
Also, the fact that the default is Cairo0 makes us create a LegacyTestContract(Cairo0), which means nothing.
</blockquote></details>

see above (no native legacy): i think this can be simplified
```
contract.supported_versions().into_iter().map(|version| {
    let mut versioned_contract = contract.clone();
    versioned_contract.set_version(version);
    versioned_contract
})
```
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 450 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODL3maT6JTki0R7h-1K:-ODLONzxBAwFb54tdMSe:bq7isbr) ([raw file](https://github.com/starkware-libs/sequencer/blob/faa8438968d1926637800cb3e9ca7269be47e2e3/crates/blockifier/src/test_utils/contracts.rs#L450)):*

Let's make it simple. Remove all functions that are used only here, and implement `all_contracts` using a single `match` expression explicitly.
Remove the `.other()` function as well
</blockquote></details>

I suggest implementing a `supported_versions()` method and iterating over it.
___
*[`crates/blockifier/tests/feature_contracts_compatibility_test.rs` line 155 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODLOo1Y-fof9oyhfkVm:-ODLOo1Y-fof9oyhfkVn:b-59c8e4) ([raw file](https://github.com/starkware-libs/sequencer/blob/99ddcb085396e833625d0fdb499bbb462ade47ff/crates/blockifier/tests/feature_contracts_compatibility_test.rs#L155)):*
> ```XML
>         // TODO (Meshi 01/01/2025) Remove this once native folder is deleted.
>         #[cfg(feature = "cairo_native")]
>         CompilationArtifacts::Cairo1Native { .. } => {
> ```

do you even need this variant now that you aren't compiling sierra->native?



_Code quote:_
```XML
CompilationArtifacts::Cairo1Native
```
___
*[`crates/blockifier/tests/feature_contracts_compatibility_test.rs` line 293 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODLPHdc3dsZYsKgXGBH:-ODLPHdc3dsZYsKgXGBI:b-4f0ni1) ([raw file](https://github.com/starkware-libs/sequencer/blob/99ddcb085396e833625d0fdb499bbb462ade47ff/crates/blockifier/tests/feature_contracts_compatibility_test.rs#L293)):*
> ```XML
>         #[cfg(feature = "cairo_native")]
>         CairoVersion::Native => panic!("This test does not support native CairoVersion."),
>     }
> ```

cleaner, IMO (no `let mut x: T;` lines)



_Suggestion:_
```XML
    let compiled_paths_on_filesystem = match cairo_version {
        CairoVersion::Cairo0 => {
            verify_and_get_files(cairo_version)
                .into_iter()
                .map(|metadata| metadata.compiled_path().to_string())
                .collect()
        }
        CairoVersion::Cairo1 => {
            let (fetched_compiled_paths_on_filesystem, sierra_paths_on_filesystem) =
                verify_and_get_files(cairo_version)
                    .into_iter()
                    .map(|metadata| (metadata.compiled_path(), metadata.sierra_path()))
                    .collect();

            let mut sierra_paths_from_enum: Vec<String> = FeatureContract::all_feature_contracts()
                .filter(|contract| contract.cairo_version() == cairo_version)
                .map(|contract| contract.get_sierra_path())
                .collect();

            sierra_paths_from_enum.sort();
            sierra_paths_on_filesystem.sort();
            assert_eq!(sierra_paths_from_enum, sierra_paths_on_filesystem);
            fetched_compiled_paths_on_filesystem
        }

        #[cfg(feature = "cairo_native")]
        CairoVersion::Native => panic!("This test does not support native CairoVersion."),
    }
```
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 302 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODLNUfdB-yVZWbYNq1N:-ODLNUfdB-yVZWbYNq1O:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/99ddcb085396e833625d0fdb499bbb462ade47ff/crates/blockifier/src/test_utils/contracts.rs#L302)):*
> ```XML
>                     #[cfg(feature = "cairo_native")]
>                     CairoVersion::Native => "sierra",
>                 },
> ```





_Suggestion:_
```XML
                "feature_contracts/cairo{}/{}{}.json",
                match cairo_version {
                    CairoVersion::Cairo0 => "0/compiled",
                    CairoVersion::Cairo1 => "1/compiled",
                    #[cfg(feature = "cairo_native")]
                    CairoVersion::Native => "1/sierra",
                },
```

Yoni-Starkware (commented):
> Reviewed 1 of 3 files at r3.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: 6 of 8 files reviewed, 7 unresolved discussions (waiting on @dorimedini-starkware and @meship-starkware)

meship-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: 6 of 8 files reviewed, 7 unresolved discussions (waiting on @dorimedini-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/test_utils/contracts.rs` line 118 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODKUx6F3kuvPrIE7Oqq:-ODLUxvt9JSjNdh5ZpaJ:bbamws1) ([raw file](https://github.com/starkware-libs/sequencer/blob/3e035be9e34bb43f042eae2b157517a9c976bcc7/crates/blockifier/src/test_utils/contracts.rs#L118)):*

why does the legacy contract now have two cairo1 variants? why not only cairo1 non-native?
IIRC the legacy contract must be compiled with a legacy compiler --> too old for a native compilation; no?
IIUC we can't run native on old cairo1 contracts...
if I'm right, you can simply revert this added `version` data in the legacy variant and all is well again.
IF we will have a contract that doesn't have cairo0 but has two cairo1 versions, then you're right... we'll need the following refactor:
```
enum RunnableCairo1Version {
    Casm,
    Native,
}

enum RunnableVersion {
    Cairo0,
    Cairo1(RunnableCairo1Version),
}

enum FeatureContract {
    SecurityTests,
    TestContract(RunnableVersion),
    OnlyCairo1Contract(RunnableCairo1Version),
    ... etc
}
```
</blockquote></details>

NM created a new contract that is the same as the legacy contract and called it test_entry_point_v1. We decided it was unnecessary, and now every Cairo1 contract also has a Native version. You are right about the compiler, though. So we can either have two legacy Sierras, or we can get back the native entry point V1 contract. I think adding the NM contract is the better answer. WDYT @Yoni-Starkware?
___
*[`crates/blockifier/tests/feature_contracts_compatibility_test.rs` line 155 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODLOo1Y-fof9oyhfkVm:-ODLWasHB6GB1Wpweiz7:bxl3tzb) ([raw file](https://github.com/starkware-libs/sequencer/blob/99ddcb085396e833625d0fdb499bbb462ade47ff/crates/blockifier/tests/feature_contracts_compatibility_test.rs#L155)):*

do you even need this variant now that you aren't compiling sierra->native?
</blockquote></details>

The short answer is no, but this would mean that natives would not be treated differently in compile. This is fine for now, but if we would like to use the contracts compile method outside of the feature contracts test, it might be problematic.   
___
*[`crates/blockifier/tests/feature_contracts_compatibility_test.rs` line 293 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODLPHdc3dsZYsKgXGBH:-ODLZKyD0CfoVpDkKmWW:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/99ddcb085396e833625d0fdb499bbb462ade47ff/crates/blockifier/tests/feature_contracts_compatibility_test.rs#L293)):*

cleaner, IMO (no `let mut x: T;` lines)
</blockquote></details>

Done.
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 302 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODLNUfdB-yVZWbYNq1N:-ODLZc-e6gCzpOEPuNHp:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/99ddcb085396e833625d0fdb499bbb462ade47ff/crates/blockifier/src/test_utils/contracts.rs#L302)):*
> ```XML
>                     #[cfg(feature = "cairo_native")]
>                     CairoVersion::Native => "sierra",
>                 },
> ```

Done.

dorimedini-starkware (changes requested):
> Reviewed 3 of 3 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: all files reviewed, 5 unresolved discussions (waiting on @meship-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/test_utils/contracts.rs` line 118 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODKUx6F3kuvPrIE7Oqq:-ODLuxWaE3prbXYR8Wl5:b-pmwmp5) ([raw file](https://github.com/starkware-libs/sequencer/blob/3e035be9e34bb43f042eae2b157517a9c976bcc7/crates/blockifier/src/test_utils/contracts.rs#L118)):*
> I think adding the NM contract is the better answer

you mean bringing back the test_entry_point_v1 contract? sounds simplest to me too
___
*[`crates/blockifier/tests/feature_contracts_compatibility_test.rs` line 155 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODLOo1Y-fof9oyhfkVm:-ODLvA7x9rCfV3Gshwnz:b-x3lkt9) ([raw file](https://github.com/starkware-libs/sequencer/blob/99ddcb085396e833625d0fdb499bbb462ade47ff/crates/blockifier/tests/feature_contracts_compatibility_test.rs#L155)):*
<details><summary><i>Previously, meship-starkware (Meshi Peled) wrote…</i></summary><blockquote>

The short answer is no, but this would mean that natives would not be treated differently in compile. This is fine for now, but if we would like to use the contracts compile method outside of the feature contracts test, it might be problematic.
</blockquote></details>

As long as native compilation is done during runtime - I don't see a reason to implement any useless code in the feature contracts testing framework.
if we want to start precompiling, we'll refactor... but that can wait
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 22 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODLul9m71lOLE1DAPnT:-ODLul9m71lOLE1DAPnU:byafam) ([raw file](https://github.com/starkware-libs/sequencer/blob/99ddcb085396e833625d0fdb499bbb462ade47ff/crates/blockifier/src/test_utils/contracts.rs#L22)):*
> ```XML
> #[cfg(feature = "cairo_native")]
> use crate::execution::native::contract_class::NativeCompiledClassV1;
> #[cfg(feature = "cairo_native")]
> ```

do you still need this gate...? you deleted the relevant import



_Code quote:_
```XML
#[cfg(feature = "cairo_native")]
```

meship-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: all files reviewed, 5 unresolved discussions (waiting on @Yoni-Starkware)

___
*[`crates/blockifier/src/test_utils/contracts.rs` line 118 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODKUx6F3kuvPrIE7Oqq:-ODMEc1m2h3B9DH031ye:b-nc69f1) ([raw file](https://github.com/starkware-libs/sequencer/blob/3e035be9e34bb43f042eae2b157517a9c976bcc7/crates/blockifier/src/test_utils/contracts.rs#L118)):*

> I think adding the NM contract is the better answer

you mean bringing back the test_entry_point_v1 contract? sounds simplest to me too
</blockquote></details>

In this case, either the test_entry_point_v1 contract would have both Ciro1 and Native versions, and then it would be just like the legacy contract, or it would only have native, and then the feature contract test would also have to check native. I prefer the first case. I think it easier to assume that for all Native contracts, we have a casm one, especially when Native also needs the casm file

dorimedini-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: all files reviewed, 5 unresolved discussions (waiting on @meship-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/test_utils/contracts.rs` line 118 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODKUx6F3kuvPrIE7Oqq:-ODMLodb-mXZAQllUfFi:bl22p7x) ([raw file](https://github.com/starkware-libs/sequencer/blob/3e035be9e34bb43f042eae2b157517a9c976bcc7/crates/blockifier/src/test_utils/contracts.rs#L118)):*
> or it would only have native, and then the feature contract test would also have to check native

wdym "the feature contract test would also have to check native"?
you mean the recompile test?
can't the recompile test treat cairo1 casm and native the same way? it is not responsible for generating native artifacts; it just needs to create sierra + casm for all cairo1 sources...
I was thinking the recompilation logic would just treat both the legacy contract and the NM contract as "cairo1" for purpose of recompilation

meship-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: 0 of 38 files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/test_utils/contracts.rs` line 22 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODLul9m71lOLE1DAPnT:-ODN4UjV76SChwI5SHV0:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/99ddcb085396e833625d0fdb499bbb462ade47ff/crates/blockifier/src/test_utils/contracts.rs#L22)):*

do you still need this gate...? you deleted the relevant import
</blockquote></details>

Done.

Yoni-Starkware (commented):
> Reviewed 1 of 3 files at r5.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: 0 of 38 files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware)

dorimedini-starkware (changes requested):
> Reviewed 38 of 38 files at r7, 16 of 16 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: all files reviewed, 3 unresolved discussions (waiting on @meship-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/test_utils/contracts.rs` line 450 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODL3maT6JTki0R7h-1K:-OD_b_EWCCEZ_8Yafn4p:b-ocs7qr) ([raw file](https://github.com/starkware-libs/sequencer/blob/faa8438968d1926637800cb3e9ca7269be47e2e3/crates/blockifier/src/test_utils/contracts.rs#L450)):*

I suggest implementing a `supported_versions()` method and iterating over it.
</blockquote></details>

concrete suggestion in another comment
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 138 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-OD_ccCFB2T4nSg4empT:-OD_ccCFB2T4nSg4empU:bfi5umx) ([raw file](https://github.com/starkware-libs/sequencer/blob/69a8e319df8f8d4b38c1605c0cdb6a041a2f9afb/crates/blockifier/src/test_utils/contracts.rs#L138)):*
> ```XML
>             Self::SierraExecutionInfoV1Contract(rv) => {
>                 assert!(matches!(version, CairoVersion::Cairo1(_)));
>                 *rv = version.get_runnable_version()
> ```

... and then you won't need the `get_runnable_version` method on the CairoVersion enum



_Suggestion:_
```XML
                match version {
                    CairoVersion::Cairo0 => panic!("{self:?} has not Cairo0 variant."),
                    CairoVersion::Cairo1(runnable) => *rv = runnable,
                }
```
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 424 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-OD_aNSOEfRuN-0-s5j_:-OD_aNSOEfRuN-0-s5ja:b-2q216v) ([raw file](https://github.com/starkware-libs/sequencer/blob/a3dbc9fc77a694b5de96cf141faf570342327172/crates/blockifier/src/test_utils/contracts.rs#L424)):*
> ```XML
>     }
> 
>     pub fn all_contracts() -> impl Iterator<Item = Self> {
> ```

*EDIT*: I didn't see the `support_cairo_version` before this comment, but my suggestion
1. is still simpler IMO
2. supports native variants
3. doesn't rely on the developer to remember to update the function... as it is currently written, if i add another contract (with cairo0, for example), `support_cairo_version` won't return it in the vector, and the compiler won't complain. only the feature contracts test will fail, right?

how about:
```
pub fn all_contract_versions(&self) -> impl Iterator<Item = Self>
```
which returns 1, 2 or 3 contracts, depending on which versions are supported?
this will be a much simpler `match`:
```
match self {
    TestContract(_) | ..all contracts with all versions supported.. => {
        // create 2 or 3 variants, depending on native feature active or not
    }
    Legacy | ... all contracts with only one version.. => vec![self].into(),
    SpecialNativeContract => RunnableCairo1::iter().map(|v| ..create cairo1 variant..)
}
```
that way, in the `all_contracts` iterator can simply do
```
Self::iter().flat_map(|contract| contract.all_contract_versions())
```



_Code quote:_
```XML
pub fn all_contracts() -> impl Iterator<Item = Self>
```
___
*[`crates/blockifier/src/test_utils.rs` line 100 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-OD_btka6o87zqiQGX4M:-OD_btka6o87zqiQGX4N:b-ner19u) ([raw file](https://github.com/starkware-libs/sequencer/blob/69a8e319df8f8d4b38c1605c0cdb6a041a2f9afb/crates/blockifier/src/test_utils.rs#L100)):*
> ```XML
>     }
> 
>     pub fn get_runnable_version(&self) -> RunnableCairoVersion {
> ```

ah...
let's differentiate between the "cairo1 runtime" and the "general cairo runtime".
why do you need to convert a CairoVersion to a RunnableCairoVersion at all?
only the Cairo1 variant should have this conversion...
I would prefer a "runs_casm" boolean method, instead of abusing the runnable enum (which should be for cairo1 only)



_Code quote:_
```XML
pub fn get_runnable_version(&self) -> RunnableCairoVersion
```

meship-starkware (commented):
> Reviewed 1 of 38 files at r7, 2 of 16 files at r8.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: all files reviewed, 4 unresolved discussions (waiting on @dorimedini-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/feature_contracts/cairo1/test_contract_execution_info_v1.cairo` line 1 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-OD_gLnC7Q93-b6mXKVH:-OD_gLnC7Q93-b6mXKVI:buauqcn) ([raw file](https://github.com/starkware-libs/sequencer/blob/69a8e319df8f8d4b38c1605c0cdb6a041a2f9afb/crates/blockifier/feature_contracts/cairo1/test_contract_execution_info_v1.cairo#L1)):*
> ```Cairo
> #[starknet::contract]
> ```

This contract is unnecessary if we use the legacy contract and ignore the Sierra version constraint on the native compilation for the V1 execution info test. Still, I think I'd prefer to keep it for now and add a PR that will remove this contract later so we can Start adding more tests for native.

Yoni-Starkware (commented):
> Reviewed 1 of 16 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: all files reviewed, 4 unresolved discussions (waiting on @dorimedini-starkware)

meship-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: 25 of 40 files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/test_utils.rs` line 100 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-OD_btka6o87zqiQGX4M:-ODaLwRh5fSQC-v9v3zR:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/69a8e319df8f8d4b38c1605c0cdb6a041a2f9afb/crates/blockifier/src/test_utils.rs#L100)):*

ah...
let's differentiate between the "cairo1 runtime" and the "general cairo runtime".
why do you need to convert a CairoVersion to a RunnableCairoVersion at all?
only the Cairo1 variant should have this conversion...
I would prefer a "runs_casm" boolean method, instead of abusing the runnable enum (which should be for cairo1 only)
</blockquote></details>

Done.
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 450 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODL3maT6JTki0R7h-1K:-ODaV9SG6p0X-HP3opfJ:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/faa8438968d1926637800cb3e9ca7269be47e2e3/crates/blockifier/src/test_utils/contracts.rs#L450)):*

concrete suggestion in another comment
</blockquote></details>

Done.
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 424 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-OD_aNSOEfRuN-0-s5j_:-ODaUk-PBhKPNR5n_3tE:bhlramk) ([raw file](https://github.com/starkware-libs/sequencer/blob/a3dbc9fc77a694b5de96cf141faf570342327172/crates/blockifier/src/test_utils/contracts.rs#L424)):*

*EDIT*: I didn't see the `support_cairo_version` before this comment, but my suggestion
1. is still simpler IMO
2. supports native variants
3. doesn't rely on the developer to remember to update the function... as it is currently written, if i add another contract (with cairo0, for example), `support_cairo_version` won't return it in the vector, and the compiler won't complain. only the feature contracts test will fail, right?

how about:
```
pub fn all_contract_versions(&self) -> impl Iterator<Item = Self>
```
which returns 1, 2 or 3 contracts, depending on which versions are supported?
this will be a much simpler `match`:
```
match self {
    TestContract(_) | ..all contracts with all versions supported.. => {
        // create 2 or 3 variants, depending on native feature active or not
    }
    Legacy | ... all contracts with only one version.. => vec![self].into(),
    SpecialNativeContract => RunnableCairo1::iter().map(|v| ..create cairo1 variant..)
}
```
that way, in the `all_contracts` iterator can simply do
```
Self::iter().flat_map(|contract| contract.all_contract_versions())
```
</blockquote></details>

I don't understand why we need the native contract as well, but I agree that a compilation error when adding a contract is better than failing a test. LMK, if you feel it's too complicated @Yoni-Starkware @dorimedini-starkware  
Done.  
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 138 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-OD_ccCFB2T4nSg4empT:-ODaMGHp3wK3uvYDIQB4:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/69a8e319df8f8d4b38c1605c0cdb6a041a2f9afb/crates/blockifier/src/test_utils/contracts.rs#L138)):*

... and then you won't need the `get_runnable_version` method on the CairoVersion enum
</blockquote></details>

Done.

dorimedini-starkware (changes requested):
> Reviewed 15 of 15 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: all files reviewed, 3 unresolved discussions (waiting on @meship-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/test_utils/contracts.rs` line 424 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-OD_aNSOEfRuN-0-s5j_:-ODaZR-t8de9FcDFrpux:bapb0v1) ([raw file](https://github.com/starkware-libs/sequencer/blob/a3dbc9fc77a694b5de96cf141faf570342327172/crates/blockifier/src/test_utils/contracts.rs#L424)):*
<details><summary><i>Previously, meship-starkware (Meshi Peled) wrote…</i></summary><blockquote>

I don't understand why we need the native contract as well, but I agree that a compilation error when adding a contract is better than failing a test. LMK, if you feel it's too complicated @Yoni-Starkware @dorimedini-starkware  
Done.
</blockquote></details>

I think now is better :)
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 438 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODaY92e3yNueMbXtDV4:-ODaY92e3yNueMbXtDV5:bhl8hnq) ([raw file](https://github.com/starkware-libs/sequencer/blob/60b4d1d0c68101b53f1c4c68f70b72aa58c267c1/crates/blockifier/src/test_utils/contracts.rs#L438)):*
> ```XML
>                     ret_val.push(native_contract);
>                 }
>                 ret_val
> ```

how about this? you won't need the assert, and this function can now be called even when `self` isn't Cairo0



_Suggestion:_
```XML
                #[cfg(not(feature = "cairo_native"))]
                let versions = vec![
                    CairoVersion::Cairo0,
                    CairoVersion::Cairo1(RunnableCairo1::Casm)
                ];
                #[cfg(feature = "cairo_native")]
                let versions = vec![
                    CairoVersion::Cairo0,
                    CairoVersion::Cairo1(RunnableCairo1::Casm),
                    CairoVersion::Cairo1(RunnableCairo1::Native),
                ];
                versions.iter().map(|v| {
                    let versioned_contract = *self;
                    versioned_contract.set_cairo_version(v);
                    versioned_contract
                })
```
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 452 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODaYulmAmJN-BOgAloD:-ODaYulmAmJN-BOgAloE:bejm6ub) ([raw file](https://github.com/starkware-libs/sequencer/blob/60b4d1d0c68101b53f1c4c68f70b72aa58c267c1/crates/blockifier/src/test_utils/contracts.rs#L452)):*
> ```XML
>                     ret_val.push(native_contract);
>                 }
>                 ret_val
> ```

see above: conditionally define a `versions` vector by feature, and then map.
that way you won't need the assert.
consider, also, doing this (only compute the `versions` vector in the `match`, and share the `map` logic):
```
let versions = match self {
    ...
};
versions.iter().map(|v| {
    let versioned_contract = *self;
    versioned_contract.set_cairo_version(v);
    versioned_contract
})
```



_Code quote:_
```XML
                // This would only work if the default runnable cairo1 is Casm.
                assert!(self.cairo_version() == CairoVersion::Cairo1(RunnableCairo1::Casm));
                let mut ret_val = vec![*self];
                #[cfg(feature = "cairo_native")]
                {
                    let mut native_contract = *self;
                    native_contract
                        .set_cairo_version(CairoVersion::Cairo1(RunnableCairo1::Native));
                    ret_val.push(native_contract);
                }
                ret_val
```

meship-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: 39 of 40 files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware and @Yoni-Starkware)

___
*[`crates/blockifier/src/test_utils/contracts.rs` line 438 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODaY92e3yNueMbXtDV4:-ODampODAs67Vr2Pi2PG:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/01e3d804b604f077ead7e6b996824cd2a62cfcac/crates/blockifier/src/test_utils/contracts.rs#L438)):*

how about this? you won't need the assert, and this function can now be called even when `self` isn't Cairo0
</blockquote></details>

Done.
___
*[`crates/blockifier/src/test_utils/contracts.rs` line 452 at r9](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-ODaYulmAmJN-BOgAloD:-ODan5Px9Axx9pJO69Ac:bcnzevi) ([raw file](https://github.com/starkware-libs/sequencer/blob/01e3d804b604f077ead7e6b996824cd2a62cfcac/crates/blockifier/src/test_utils/contracts.rs#L452)):*

see above: conditionally define a `versions` vector by feature, and then map.
that way you won't need the assert.
consider, also, doing this (only compute the `versions` vector in the `match`, and share the `map` logic):
```
let versions = match self {
    ...
};
versions.iter().map(|v| {
    let versioned_contract = *self;
    versioned_contract.set_cairo_version(v);
    versioned_contract
})
```
</blockquote></details>

I think that the match should return the final vector. I made a helper function that shares the map logic. I can have the match return the version list, but then I would have to separate the only cairo1 from the only cairo0. If you think that is better, then I'll do it.

dorimedini-starkware (approved):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: all files reviewed, 1 unresolved discussion (waiting on @Yoni-Starkware)

Yoni-Starkware (approved):
:lgtm:

> Reviewed 4 of 38 files at r7, 1 of 16 files at r8, 5 of 15 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464#-:-ODaxbM920SnyGKNkZQM:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @meship-starkware)

dorimedini-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2464)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @meship-starkware)

PR #2500, opened 2024-12-05
feat: allow limiting system resources in compilation processes

avi-starkware:


ArniStarkware (changes requested):
> Reviewed 4 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: 4 of 6 files reviewed, 5 unresolved discussions (waiting on @avi-starkware, @liorgold2, @noaov1, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 0 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkpeG03zX_e86zaBct:-ODkpeG03zX_e86zaBcu:b-xsjvy9) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits_test.rs#L0)):*
Add a test case for a simple CI command, with all resources limited, but not blocking the command.
___
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 22 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkoyVc5lcOFp44S0Oz:-ODkoyVc5lcOFp44S0P-:b-ntzd2x) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits_test.rs#L22)):*
> ```XML
> #[rstest]
> fn test_memory_size_limit() {
>     let memory_size_limits = ResourcesLimits::new(None, None, Some(100000));
> ```

Will the test still be substantial that way? It would take less resources. 
Make sure it takes as few resources for CI.



_Suggestion:_
```XML
let memory_size_limits = ResourcesLimits::new(None, None, Some(100));
```
___
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 38 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkpEm8BD4uVhy6RvOR:-ODkpEm8BD4uVhy6RvOS:b-xs3hob) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits_test.rs#L38)):*
> ```XML
>     file_size_limits.apply(&mut command);
>     command.spawn().expect("Failed to start disk consuming process").wait().unwrap();
>     assert!(std::fs::metadata("/tmp/file.txt").unwrap().len() <= 100);
> ```

Make the test more consise, and take less resources.



_Suggestion:_
```XML
    let file_size_limit = 10;
    let file_size_limits = ResourcesLimits::new(None, Some(file_size_limit), None);

    let mut command = Command::new("bash");
    command.args(["-c", "echo 0 > /tmp/file.txt; while true; do echo 0 >> /tmp/file.txt; done;"]);
    file_size_limits.apply(&mut command);
    command.spawn().expect("Failed to start disk consuming process").wait().unwrap();
    assert_eq!(std::fs::metadata("/tmp/file.txt").unwrap().len(), file_size_limit); // Or whatever the resoult should be.
```
___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 11 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkl3IU6v5g5oOaVm5h:-ODkl3IU6v5g5oOaVm5i:bgvpfh8) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L11)):*
> ```XML
> pub mod test;
> 
> #[allow(dead_code)]
> ```

Can we set this flag on the file level?



_Code quote:_
```XML
#[allow(dead_code)]
```
___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 17 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODklM-JCQMl3jh87YuM:-ODklM-JCQMl3jh87YuN:bcs5vg9) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L17)):*
> ```XML
>     hard: u64,
>     units: String,
> }
> ```

Partway through writing this, I partially regretted that.  I am leaving the comment more as a discussion than a suggestion.

If I understand the use case correctly, the `Resource` is implied.
The method `set` can have two parameters. `&self, resource: Resource`.

You can even define an iterator over `ResourcesLimits` that returns tuples `(Resource, ResourceLimits)`.



_Suggestion:_
```XML
struct ResourceLimits {
    soft: u64,
    hard: u64,
    units: String,
}
```
___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 91 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkoJELB5GrXx-bl8nW:-ODkoJEM2tFd5SOuaOTI:buhhilg) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L91)):*
> ```XML
>             // 2. No heap allocations occur in the `set` method.
>             // 3. `setrlimit` is an async-signal-safe system call, which means it is safe to invoke
>             //   after `fork`.
> ```

Did this block of text pass through Grammerly/ ChatGPT? Looks good to me.



_Code quote:_
```XML
            // The `pre_exec` method runs a given closure after forking the parent process and
            // before the exec of the child process.
            //
            // This closure will be run in the context of the child process after a `fork`. This
            // primarily means that any modifications made to memory on behalf of this closure will
            // **not** be visible to the parent process. This is often a very constrained
            // environment where normal operations like `malloc`, accessing environment variables
            // through [`std::env`] or acquiring a mutex are not guaranteed to work (due to other
            // threads perhaps still running when the `fork` was run).
            //
            // The current closure is safe for the following reasons:
            // 1. The [`ResourcesLimits`] struct is fully constructed and moved into the closure.
            // 2. No heap allocations occur in the `set` method.
            // 3. `setrlimit` is an async-signal-safe system call, which means it is safe to invoke
            //   after `fork`.
```

avi-starkware:
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 17 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODklM-JCQMl3jh87YuM:-ODkurI18dhq1avJOqZP:b5jj1pn) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L17)):*
<details><summary><i>Previously, ArniStarkware (Arnon Hod) wrote…</i></summary><blockquote>

Partway through writing this, I partially regretted that.  I am leaving the comment more as a discussion than a suggestion.

If I understand the use case correctly, the `Resource` is implied.
The method `set` can have two parameters. `&self, resource: Resource`.

You can even define an iterator over `ResourcesLimits` that returns tuples `(Resource, ResourceLimits)`.
</blockquote></details>

I don't fully understand your suggestion. How can I limit a resource without knowing what resource it is? How would I know what units this resource is counted in?

The point of this struct is to hold all the information needed for using `setrlimit()` and for displaying a short log message indicating what happened.

ArniStarkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: 4 of 6 files reviewed, 5 unresolved discussions (waiting on @avi-starkware, @liorgold2, @noaov1, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 17 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODklM-JCQMl3jh87YuM:-ODkwFK10-ktJuGMWy_Z:b-2qhskj) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L17)):*
<details><summary><i>Previously, avi-starkware wrote…</i></summary><blockquote>

I don't fully understand your suggestion. How can I limit a resource without knowing what resource it is? How would I know what units this resource is counted in?

The point of this struct is to hold all the information needed for using `setrlimit()` and for displaying a short log message indicating what happened.
</blockquote></details>

The point is that in the context of the struct `ResourcesLimits` (Plural), it is clear from the context that `cpu_time.resource = Resource::CPU`. Thus, it is redundant as a member of `ResourceLimits` (Singular).
___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 0 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkxJyv8bRNDY68UXb6:-ODkxJyv8bRNDY68UXb7:berg3gj) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L0)):*
The names `ResoureceLimits` vs `ResourcesLimits` is confusing.

avi-starkware:
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 22 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkoyVc5lcOFp44S0Oz:-ODkwyiQ3Zx1zk1SbGzr:byxdhji) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits_test.rs#L22)):*
<details><summary><i>Previously, ArniStarkware (Arnon Hod) wrote…</i></summary><blockquote>

Will the test still be substantial that way? It would take less resources. 
Make sure it takes as few resources for CI.
</blockquote></details>

This gives the process 100 bytes of memory. The process can't start with this limitation in place. 1KB is also not enough. With 10KB it works locally for me, but fails in the CI.

avi-starkware:
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 91 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkoJELB5GrXx-bl8nW:-ODkw5y_76Ng11b6zmuX:b-pbxleu) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L91)):*
<details><summary><i>Previously, ArniStarkware (Arnon Hod) wrote…</i></summary><blockquote>

Did this block of text pass through Grammerly/ ChatGPT? Looks good to me.
</blockquote></details>

I think the more important question is whether my explanation is easy to understand and whether it justifies using an `unsafe` block.

ArniStarkware:
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 91 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkoJELB5GrXx-bl8nW:-ODkypMX4ycn75LMfNO-:bi42dot) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L91)):*
<details><summary><i>Previously, avi-starkware wrote…</i></summary><blockquote>

I think the more important question is whether my explanation is easy to understand and whether it justifies using an `unsafe` block.
</blockquote></details>

I agree the question you raise is more important. I asked another, less important question :)

ArniStarkware:
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 22 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkoyVc5lcOFp44S0Oz:-ODkzFsB8Zr004zE4ZVJ:b-4hvi0m) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits_test.rs#L22)):*
<details><summary><i>Previously, avi-starkware wrote…</i></summary><blockquote>

This gives the process 100 bytes of memory. The process can't start with this limitation in place. 1KB is also not enough. With 10KB it works locally for me, but fails in the CI.
</blockquote></details>

Got it.

ArniStarkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: 4 of 6 files reviewed, 4 unresolved discussions (waiting on @avi-starkware, @liorgold2, @noaov1, and @Yoni-Starkware)

avi-starkware:
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 17 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODklM-JCQMl3jh87YuM:-ODky8cX3zGReFd-jh1O:b3vw7xa) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L17)):*
<details><summary><i>Previously, ArniStarkware (Arnon Hod) wrote…</i></summary><blockquote>

The point is that in the context of the struct `ResourcesLimits` (Plural), it is clear from the context that `cpu_time.resource = Resource::CPU`. Thus, it is redundant as a member of `ResourceLimits` (Singular).
</blockquote></details>

The name cpu_time is the name of the field I made up. Resource::CPU is the enum value to be passed to `setrlimit()`. I can have a `match` to resolve the correct resource entry, or I can just remove the struct `ResourceLimits` entirely if you think it is not needed.

ArniStarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: 4 of 6 files reviewed, 4 unresolved discussions (waiting on @avi-starkware, @liorgold2, @noaov1, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 17 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODklM-JCQMl3jh87YuM:-ODl-0T9E26QaFOtfyFe:b-t9titv) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L17)):*
<details><summary><i>Previously, avi-starkware wrote…</i></summary><blockquote>

The name cpu_time is the name of the field I made up. Resource::CPU is the enum value to be passed to `setrlimit()`. I can have a `match` to resolve the correct resource entry, or I can just remove the struct `ResourceLimits` entirely if you think it is not needed.
</blockquote></details>

I have a suggestion in another thread.
___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 71 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODl10Yx5Y3VdXseIy8z:-ODl10Yx5Y3VdXseIy9-:b-ikeqs4) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L71)):*
> ```XML
>             .flatten()
>             .try_for_each(|resource_limit| resource_limit.set())
>     }
> ```

What do you think?



_Suggestion:_
```XML
#[allow(dead_code)]
struct ResourcesLimits {
    pub cpu_time: Option<u64>,
    pub file_size: Option<u64>,
    pub memory_size: Option<u64>,
}

impl ResourcesLimits {
    fn cpu_time_resource(&self) -> Option<ResourceLimits> {
        self.cpu_time.map(|cpu_time| ResourceLimits {
            resource: Resource::CPU,
            soft: cpu_time,
            hard: cpu_time,
            units: "seconds".to_string(),
        })
    }

    fn file_size_resource(&self) -> Option<ResourceLimits> {
        self.file_size.map(|file_size| ResourceLimits {
            resource: Resource::FSIZE,
            soft: file_size,
            hard: file_size,
            units: "bytes".to_string(),
        })
    }

    fn memory_size_resource(&self) -> Option<ResourceLimits> {
        self.memory_size.map(|memory_size| ResourceLimits {
            resource: Resource::AS,
            soft: memory_size,
            hard: memory_size,
            units: "bytes".to_string(),
        })
    }

    #[allow(dead_code)]
    fn set(&self) -> io::Result<()> {
        [self.cpu_time_resource(), self.file_size_resource(), self.memory_size_resource()]
            .iter()
            .flatten()
            .try_for_each(|resource_limit| resource_limit.set())
    }
```

ArniStarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: 4 of 6 files reviewed, 4 unresolved discussions (waiting on @avi-starkware, @liorgold2, @noaov1, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 71 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODl10Yx5Y3VdXseIy8z:-ODl1OBU3T0T-yimYvr4:bkr4zqf) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L71)):*
<details><summary><i>Previously, ArniStarkware (Arnon Hod) wrote…</i></summary><blockquote>

What do you think?
</blockquote></details>

Also - the struct `ResourcesLimits` and the fields should be documented.

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: 4 of 6 files reviewed, 4 unresolved discussions (waiting on @ArniStarkware, @liorgold2, @noaov1, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 91 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkoJELB5GrXx-bl8nW:-ODlgY453EiAv42sMCYj:bwjs8e3) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L91)):*
<details><summary><i>Previously, ArniStarkware (Arnon Hod) wrote…</i></summary><blockquote>

I agree the question you raise is more important. I asked another, less important question :)
</blockquote></details>

I consulted chatGPT and made some changes to the text 

TAL
___
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 38 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkpEm8BD4uVhy6RvOR:-ODlgi7L-4WfTpK4j6Q-:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits_test.rs#L38)):*
<details><summary><i>Previously, ArniStarkware (Arnon Hod) wrote…</i></summary><blockquote>

Make the test more consise, and take less resources.
</blockquote></details>

Done.

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: 4 of 6 files reviewed, 3 unresolved discussions (waiting on @ArniStarkware, @liorgold2, @noaov1, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 11 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkl3IU6v5g5oOaVm5h:-ODllEQHE13GN93jchqH:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L11)):*
<details><summary><i>Previously, ArniStarkware (Arnon Hod) wrote…</i></summary><blockquote>

Can we set this flag on the file level?
</blockquote></details>

Done.
___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 71 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODl10Yx5Y3VdXseIy8z:-ODlqFeiEu6Qcif0Vqf1:b-r4ey5m) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L71)):*
<details><summary><i>Previously, ArniStarkware (Arnon Hod) wrote…</i></summary><blockquote>

Also - the struct `ResourcesLimits` and the fields should be documented.
</blockquote></details>

Added documentation.

I prefer my approach for two reasons:
1. The logic running inside the `apply` command is minimal. All structs are already constructed and moved into the closure at the point when `set` is called. This is desirable for code running inside an `unsafe` block.
2. I think my approach has much fewer moving parts.  Each element does very few things and is easy to understand by itself.
___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line  at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkxJyv8bRNDY68UXb6:-ODllFo474Gaqoqz4sGk:b-a1k9sg) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits.rs#L)):*
<details><summary><i>Previously, ArniStarkware (Arnon Hod) wrote…</i></summary><blockquote>

The names `ResoureceLimits` vs `ResourcesLimits` is confusing.
</blockquote></details>

I changed it.
TAL

avi-starkware:
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line  at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-ODkpeG03zX_e86zaBct:-OEDtYm01I7scFLn0-ey:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/69aff3ef5292774b79f0e071d6a61d1a38f250d1/crates/starknet_sierra_compile/src/resource_limits_test.rs#L)):*
<details><summary><i>Previously, ArniStarkware (Arnon Hod) wrote…</i></summary><blockquote>

Add a test case for a simple CI command, with all resources limited, but not blocking the command.
</blockquote></details>

Done.

ArniStarkware (approved):
:lgtm:

> Reviewed 1 of 2 files at r4, 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-:-OEE9R2L7QnxaIA-lDwK:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @liorgold2, @noaov1, and @Yoni-Starkware)

noaov1 (commented):
> Reviewed 3 of 5 files at r2, 1 of 2 files at r4, 1 of 2 files at r5, 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: all files reviewed, 3 unresolved discussions (waiting on @avi-starkware, @liorgold2, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 29 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEk9ZC14lFhuKlnYlO2:-OEk9ZC14lFhuKlnYlO3:b-u17ziw) ([raw file](https://github.com/starkware-libs/sequencer/blob/43ee3d62b88cd879d4ea8dae9548bc7464d40bdd/crates/starknet_sierra_compile/src/resource_limits.rs#L29)):*
> ```XML
>     fn set(&self) -> io::Result<()> {
>         // Use `println!` and not a logger because this method is called in an unsafe block, and we
>         // don't want to risk unexpected behavior.
> ```

Can you please elaborate? 



_Code quote:_
```XML
        // Use `println!` and not a logger because this method is called in an unsafe block, and we
        // don't want to risk unexpected behavior.
```
___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 30 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEk9c3b4X_L4meuWbnI:-OEk9c3b4X_L4meuWbnJ:b-uq5b1u) ([raw file](https://github.com/starkware-libs/sequencer/blob/43ee3d62b88cd879d4ea8dae9548bc7464d40bdd/crates/starknet_sierra_compile/src/resource_limits.rs#L30)):*
> ```XML
>         // Use `println!` and not a logger because this method is called in an unsafe block, and we
>         // don't want to risk unexpected behavior.
>         println!("Setting {:?} limits to {} {}.", self.resource, self.soft_limit, self.units);
> ```

Consider adding the hard limit as well



_Code quote:_
```XML
        println!("Setting {:?} limits to {} {}.", self.resource, self.soft_limit, self.units);
```
___
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 23 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEkB_r5C4UsC_Lo_x25:-OEkB_r5C4UsC_Lo_x26:b-pyf44y) ([raw file](https://github.com/starkware-libs/sequencer/blob/43ee3d62b88cd879d4ea8dae9548bc7464d40bdd/crates/starknet_sierra_compile/src/resource_limits_test.rs#L23)):*
> ```XML
> 
> #[rstest]
> fn test_memory_size_limit() {
> ```

Can you please explain how do you assert the memory limit in this test?



_Code quote:_
```XML
fn test_memory_size_limit() {
```

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: 2 of 6 files reviewed, 2 unresolved discussions (waiting on @ArniStarkware, @liorgold2, @noaov1, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 29 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEk9ZC14lFhuKlnYlO2:-OEnTjcn3ameOGhf7nN1:bvhbgth) ([raw file](https://github.com/starkware-libs/sequencer/blob/43ee3d62b88cd879d4ea8dae9548bc7464d40bdd/crates/starknet_sierra_compile/src/resource_limits.rs#L29)):*

Can you please elaborate?
</blockquote></details>

Using a logger (e.g., with `log:info!`) allows injecting logic into the command that prints the log message. This logger can do things that would be problematic in our unsafe block. For example, it could take a lock on a mutex which might never gets release in the unsafe block.
___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 30 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEk9c3b4X_L4meuWbnI:-OEnoeBJ7WE5zkZd8EHt:bimwk30) ([raw file](https://github.com/starkware-libs/sequencer/blob/43ee3d62b88cd879d4ea8dae9548bc7464d40bdd/crates/starknet_sierra_compile/src/resource_limits.rs#L30)):*

Consider adding the hard limit as well
</blockquote></details>

Done.

I didn't add it because in the current implementation I always set `soft_limit == hard_limit`, but we might want to use this object differently in the future.
___
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 23 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEkB_r5C4UsC_Lo_x25:-OEnJQ1B3WFWdz2yvOQC:bdt1v2h) ([raw file](https://github.com/starkware-libs/sequencer/blob/43ee3d62b88cd879d4ea8dae9548bc7464d40bdd/crates/starknet_sierra_compile/src/resource_limits_test.rs#L23)):*

Can you please explain how do you assert the memory limit in this test?
</blockquote></details>

I couldn't find a way to check that the memory limit was indeed enforced. I just check that running an infinite loop fails at some point.

I could use an external crate called `sysinfo` that allows polling the memory usage of a running process. It only works while the process is running, so I would have to have a loop that checks the memory usage frequently until the process stops, and make sure that no sample exceeds the memory_limit. Would you like me to do that?

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: 1 of 6 files reviewed, 3 unresolved discussions (waiting on @ArniStarkware, @liorgold2, @noaov1, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 23 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEkB_r5C4UsC_Lo_x25:-OErnUqq8bqA6LOVUvTs:b-burnii) ([raw file](https://github.com/starkware-libs/sequencer/blob/43ee3d62b88cd879d4ea8dae9548bc7464d40bdd/crates/starknet_sierra_compile/src/resource_limits_test.rs#L23)):*
<details><summary><i>Previously, avi-starkware wrote…</i></summary><blockquote>

I couldn't find a way to check that the memory limit was indeed enforced. I just check that running an infinite loop fails at some point.

I could use an external crate called `sysinfo` that allows polling the memory usage of a running process. It only works while the process is running, so I would have to have a loop that checks the memory usage frequently until the process stops, and make sure that no sample exceeds the memory_limit. Would you like me to do that?
</blockquote></details>

I added polling the memory of the child process to check it does not exceed the memory limit.
The test now takes 1.1 seconds though, because if I give the process less memory than 4.5MB, it doesn't have enough memory to start the process.

Moreover, an external crate ([sysinfo](https://docs.rs/sysinfo/latest/sysinfo/)) is needed to check the child-process memory usage.

PTAL

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: 1 of 6 files reviewed, 3 unresolved discussions (waiting on @ArniStarkware, @liorgold2, @noaov1, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 23 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEkB_r5C4UsC_Lo_x25:-OErp6_VD4HC2xUs7nQP:b-n4biip) ([raw file](https://github.com/starkware-libs/sequencer/blob/43ee3d62b88cd879d4ea8dae9548bc7464d40bdd/crates/starknet_sierra_compile/src/resource_limits_test.rs#L23)):*
<details><summary><i>Previously, avi-starkware wrote…</i></summary><blockquote>

I added polling the memory of the child process to check it does not exceed the memory limit.
The test now takes 1.1 seconds though, because if I give the process less memory than 4.5MB, it doesn't have enough memory to start the process.

Moreover, an external crate ([sysinfo](https://docs.rs/sysinfo/latest/sysinfo/)) is needed to check the child-process memory usage.

PTAL
</blockquote></details>

The crate looks actively maintained:
- last commit was yesterday
- new version tag every month

noaov1 (approved):
> Reviewed 3 of 5 files at r10, 1 of 2 files at r11, 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: all files reviewed, 3 unresolved discussions (waiting on @avi-starkware, @liorgold2, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 23 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEkB_r5C4UsC_Lo_x25:-OEt742E2En-Fr_K30gW:b-p4n8i6) ([raw file](https://github.com/starkware-libs/sequencer/blob/43ee3d62b88cd879d4ea8dae9548bc7464d40bdd/crates/starknet_sierra_compile/src/resource_limits_test.rs#L23)):*
> I just check that running an infinite loop fails at some point.

Is there a way to check the failure reason of a spawned thread?
I think that using  sysinfo may be too much. 
___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 87 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEtFHHcEq22Bh4An3GO:-OEtFHHcEq22Bh4An3GP:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/17180b8c92a246c4bdd2d3eb12298b291ccdc59d/crates/starknet_sierra_compile/src/resource_limits.rs#L87)):*
> ```XML
>     /// Apply the resource limits to a given command. This moves the [`ResourceLimits`] struct into
>     /// a closure that is held by the given command. The closure is executed once the command is
>     /// executed, inside the spawned child process.
> ```





_Suggestion:_
```XML
    /// Apply the resource limits to a given command object (a representation of a child process).
    /// This moves the [`ResourceLimits`] struct into a closure that is held by the given command.
    /// The closure is executed right before the exec system call is invoked in the child process.
```
___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 100 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEtIa8e1CFpcV_gNWh7:-OEtIa8e1CFpcV_gNWh8:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/17180b8c92a246c4bdd2d3eb12298b291ccdc59d/crates/starknet_sierra_compile/src/resource_limits.rs#L100)):*
> ```XML
>             // as using `malloc`, accessing environment variables through [`std::env`] or acquiring
>             // a mutex--are not guaranteed to work, because other threads may still be running at
>             // the time of `fork`.
> ```





_Suggestion:_
```XML
            // a mutex--are not guaranteed to work, because after `fork`, only one thread exists in the child
            // process, while there might be a few in the parent process.
```

avi-starkware:
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 23 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEkB_r5C4UsC_Lo_x25:-OEuBmpO8jGepHNxEPGF:b-u9x62m) ([raw file](https://github.com/starkware-libs/sequencer/blob/43ee3d62b88cd879d4ea8dae9548bc7464d40bdd/crates/starknet_sierra_compile/src/resource_limits_test.rs#L23)):*

> I just check that running an infinite loop fails at some point.

Is there a way to check the failure reason of a spawned thread?
I think that using  sysinfo may be too much.
</blockquote></details>

I found a way to do it without using sysinfo.

avi-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: 2 of 7 files reviewed, 1 unresolved discussion (waiting on @liorgold2, @noaov1, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 87 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEtFHHcEq22Bh4An3GO:-OEuCOTU8y8KcRv1RJ4r:b-ipw62l) ([raw file](https://github.com/starkware-libs/sequencer/blob/17180b8c92a246c4bdd2d3eb12298b291ccdc59d/crates/starknet_sierra_compile/src/resource_limits.rs#L87)):*
> ```XML
>     /// Apply the resource limits to a given command. This moves the [`ResourceLimits`] struct into
>     /// a closure that is held by the given command. The closure is executed once the command is
>     /// executed, inside the spawned child process.
> ```

Done . (minus the parenthesis)
A command object is not a representation of a child process. It represents a command that once executed would generate a child process.
___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 100 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEtIa8e1CFpcV_gNWh7:-OEuBxiI-EOFu1_vv71z:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/17180b8c92a246c4bdd2d3eb12298b291ccdc59d/crates/starknet_sierra_compile/src/resource_limits.rs#L100)):*
> ```XML
>             // as using `malloc`, accessing environment variables through [`std::env`] or acquiring
>             // a mutex--are not guaranteed to work, because other threads may still be running at
>             // the time of `fork`.
> ```

Done.

noaov1 (approved):
> Reviewed 5 of 5 files at r13, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @liorgold2, and @Yoni-Starkware)

___
*[`crates/starknet_sierra_compile/src/resource_limits.rs` line 100 at r12](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEtIa8e1CFpcV_gNWh7:-OEwpazT3pRpTBYiozQG:b5i4brg) ([raw file](https://github.com/starkware-libs/sequencer/blob/17180b8c92a246c4bdd2d3eb12298b291ccdc59d/crates/starknet_sierra_compile/src/resource_limits.rs#L100)):*
<details><summary><i>Previously, avi-starkware wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Is it? :)
___
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 23 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEkB_r5C4UsC_Lo_x25:-OEwz5ffBuG_kUDkzgBP:b-83sjoh) ([raw file](https://github.com/starkware-libs/sequencer/blob/43ee3d62b88cd879d4ea8dae9548bc7464d40bdd/crates/starknet_sierra_compile/src/resource_limits_test.rs#L23)):*
<details><summary><i>Previously, avi-starkware wrote…</i></summary><blockquote>

I found a way to do it without using sysinfo.
</blockquote></details>

Nice!
___
*[`crates/starknet_sierra_compile/src/resource_limits_test.rs` line 47 at r13](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-OEwysHbAVY45oUHBkR4:-OEwysHbAVY45oUHBkR5:b-q8x9rn) ([raw file](https://github.com/starkware-libs/sequencer/blob/a72c1c14fef8c780856abe956fc76cc0f5ea1546/crates/starknet_sierra_compile/src/resource_limits_test.rs#L47)):*
> ```XML
>         println!("Child process is using {} KB of memory.", memory_usage_kb);
>         assert!(memory_usage_kb * 1024 < memory_limit);
>         thread::sleep(Duration::from_millis(250));
> ```

Consider verifying the error message instead of asserting that the memory usage is within the bounds. (non-blocking)



_Code quote:_
```XML
        thread::sleep(Duration::from_millis(250));
```

noaov1 (approved):
> Reviewed 3 of 3 files at r14, 2 of 2 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @liorgold2, and @Yoni-Starkware)

noaov1 (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500#-:-OEyOIML0pfXhzvFFviN:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @liorgold2, and @Yoni-Starkware)

ArniStarkware (commented):
> Reviewed 3 of 3 files at r8.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @liorgold2, and @Yoni-Starkware)

ArniStarkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: all files reviewed, 1 unresolved discussion (waiting on @avi-starkware, @liorgold2, and @Yoni-Starkware)

noaov1 (approved):
> Reviewed 1 of 1 files at r16, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2500)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @liorgold2 and @Yoni-Starkware)

PR #2730, opened 2024-12-17
fix(consensus): dont fail when stream handler inbound receiver drops

guy-starkware:
In the StreamHandler inbound case (listening to the network and sending internal messages into Consensus) there's a case where the Consensus drops the receiver. 
This is a receiver that the StreamHandler sends whenever it sees a new stream, and the receiver should stay open for as long as that stream is not done. 

However, the Consensus may choose to drop it (e.g., when moving to a new height) and that leaves the StreamHandler with a broken pipe, and also some messages that should go into that pipe. 

As a first step, I'm making this no longer a panic, but if the receiver is dropped, then the sender inside StreamHandler is dropped, too (along with the remaining StreamData, which includes the message id, cached messages, etc). 

This still doesn't address what happens to all the other messages that should come into StreamHandler. I'll handle that in the next PR.

### DIFF
--- a/crates/sequencing/papyrus_consensus/src/stream_handler.rs
+++ b/crates/sequencing/papyrus_consensus/src/stream_handler.rs
@@ -28,6 +28,10 @@ type StreamKey = (PeerId, StreamId);
 
 const CHANNEL_BUFFER_LENGTH: usize = 100;
 
+// Use this struct for each inbound stream.
+// Drop the struct when:
+// (1) receiver on the other end is dropped,
+// (2) fin message is received and all messages are sent.
 #[derive(Debug, Clone)]
 struct StreamData<
     T: Clone + Into<Vec<u8>> + TryFrom<Vec<u8>, Error = ProtobufConversionError> + 'static,
@@ -178,13 +182,38 @@ impl<T: Clone + Send + Into<Vec<u8>> + TryFrom<Vec<u8>, Error = ProtobufConversi
         }
     }
 
-    fn inbound_send(data: &mut StreamData<T>, message: StreamMessage<T>) {
+    // Returns true if the receiver for this stream is dropped.
+    fn inbound_send(data: &mut StreamData<T>, message: StreamMessage<T>) -> bool {
         // TODO(guyn): reconsider the "expect" here.
         let sender = &mut data.sender;
         if let StreamMessageBody::Content(content) = message.message {
-            sender.try_send(content).expect("Send should succeed");
+            match sender.try_send(content) {
+                Ok(_) => {}
+                Err(e) => {
+                    if e.is_disconnected() {
+                        warn!(
+                            "Sender is disconnected, dropping the message. StreamId: {}, \
+                             MessageId: {}",
+                            message.stream_id, message.message_id
+                        );
+                        return true;
+                    } else if e.is_full() {
+                        // TODO(guyn): replace panic with buffering of the message.
+                        panic!(
+                            "Sender is full, dropping the message. StreamId: {}, MessageId: {}",
+                            message.stream_id, message.message_id
+                        );
+                    } else {
+                        // TODO(guyn): replace panic with more graceful error handling
+                        panic!("Unexpected error: {:?}", e);
+                    }
+                }
+            };
             data.next_message_id += 1;
+            return false;
         }
+        // A Fin message is not sent. This is a no-op, can safely return true.
+        true
     }
 
     // Send the message to the network.
@@ -225,6 +254,7 @@ impl<T: Clone + Send + Into<Vec<u8>> + TryFrom<Vec<u8>, Error = ProtobufConversi
                 return;
             }
         };
+
         let peer_id = metadata.originator_id;
         let stream_id = message.stream_id;
         let key = (peer_id, stream_id);
@@ -280,10 +310,14 @@ impl<T: Clone + Send + Into<Vec<u8>> + TryFrom<Vec<u8>, Error = ProtobufConversi
         // This means we can just send the message without buffering it.
         match message_id.cmp(&data.next_message_id) {
             Ordering::Equal => {
-                Self::inbound_send(data, message);
-                Self::process_buffer(data);
+                let mut receiver_dropped = Self::inbound_send(data, message);
+                if !receiver_dropped {
+                    receiver_dropped = Self::process_buffer(data);
+                }
 
-                if data.message_buffer.is_empty() && data.fin_message_id.is_some() {
+                if data.message_buffer.is_empty() && data.fin_message_id.is_some()
+                    || receiver_dropped
+                {
                     data.sender.close_channel();
                     self.inbound_stream_data.remove(&key);
                 }
@@ -323,9 +357,13 @@ impl<T: Clone + Send + Into<Vec<u8>> + TryFrom<Vec<u8>, Error = ProtobufConversi
 
     // Tries to drain as many messages as possible from the buffer (in order),
     // DOES NOT guarantee that the buffer will be empty after calling this function.
-    fn process_buffer(data: &mut StreamData<T>) {
+    // Returns true if the receiver for this stream is dropped.
+    fn process_buffer(data: &mut StreamData<T>) -> bool {
         while let Some(message) = data.message_buffer.remove(&data.next_message_id) {
-            Self::inbound_send(data, message);
+            if Self::inbound_send(data, message) {
+                return true;
+            }
         }
+        false
     }
 }
--- a/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs
+++ b/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs
@@ -395,6 +395,58 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    async fn inbound_close_channel() {
+        let (mut stream_handler, mut network_sender, mut inbound_channel_receiver, metadata, _, _) =
+            setup_test();
+
+        let stream_id = 127;
+        // Send two messages, no Fin.
+        for i in 0..2 {
+            let message = make_test_message(stream_id, i, false);
+            send(&mut network_sender, &metadata, message).await;
+        }
+
+        // Allow the StreamHandler to process the messages.
+        let join_handle = tokio::spawn(async move {
+            let _ = tokio::time::timeout(TIMEOUT, stream_handler.run()).await;
+            stream_handler
+        });
+        let mut stream_handler = join_handle.await.expect("Task should succeed");
+
+        let mut receiver = inbound_channel_receiver.next().await.unwrap();
+        for _ in 0..2 {
+            let _ = receiver.next().await.unwrap();
+        }
+
+        // Check that the stream handler contains the StreamData.
+        assert_eq!(stream_handler.inbound_stream_data.len(), 1);
+        assert_eq!(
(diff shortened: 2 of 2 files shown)


guy-starkware:
* **#2730** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/2730?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/2730?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

matan-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2730)* status: all files reviewed, 5 unresolved discussions (waiting on @guy-starkware)

___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 412 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2730#-OEKsxKl3SloWZMVj4Ep:-OEKsxKl3SloWZMVj4Eq:btmygbz) ([raw file](https://github.com/starkware-libs/sequencer/blob/75eaa43f8feb5fc2cc8d7bcc1947d609a9e3fce3/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L412)):*
> ```XML
>         // Allow the StreamHandler to process the messages.
>         let join_handle = tokio::spawn(async move {
>             let _ = tokio::time::timeout(TIMEOUT, stream_handler.run()).await;
> ```

I think often it's better not to have timeouts like this in a test. Normally tests are run with a timeout on the whole test, and so this explicit timeout is just adding complexity to your test code.



_Code quote:_
```XML
            let _ = tokio::time::timeout(TIMEOUT, stream_handler.run()).await;
```
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 433 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2730#-OEKtKFQ5iasmVHXtIrK:-OEKtKFQ5iasmVHXtIrL:b-cdb4ef) ([raw file](https://github.com/starkware-libs/sequencer/blob/75eaa43f8feb5fc2cc8d7bcc1947d609a9e3fce3/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L433)):*
> ```XML
> 
>         // Send more messages.
>         // TODO(guyn): if we set this to 2..4 it fails... the last message opens a new StreamData!
> ```

this is the follow up "ignore late messages"?



_Code quote:_
```XML
        // TODO(guyn): if we set this to 2..4 it fails... the last message opens a new StreamData!
```
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler.rs` line 185 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2730#-OEKs1G48fg6atRTi4UJ:-OEKs1G48fg6atRTi4UK:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/75eaa43f8feb5fc2cc8d7bcc1947d609a9e3fce3/crates/sequencing/papyrus_consensus/src/stream_handler.rs#L185)):*
> ```XML
>     }
> 
>     fn inbound_send(data: &mut StreamData<T>, message: StreamMessage<T>) -> bool {
> ```





_Suggestion:_
```XML
    // Returns false if the receiver is no longer subscribed to this stream.
    fn inbound_send(data: &mut StreamData<T>, message: StreamMessage<T>) -> bool {
```
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler.rs` line 204 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2730#-OEKsCBe2zjX8sEdqNUc:-OEKsCBfBpsokbNieO4I:bqgmmg7) ([raw file](https://github.com/starkware-libs/sequencer/blob/75eaa43f8feb5fc2cc8d7bcc1947d609a9e3fce3/crates/sequencing/papyrus_consensus/src/stream_handler.rs#L204)):*
> ```XML
>                             "Sender is full, dropping the message. StreamId: {}, MessageId: {}",
>                             message.stream_id, message.message_id
>                         );
> ```

Good catch. I would want this to just stay in the internal buffer for this stream_id (where we keep the out of order messages). I don't see this as an error. That buffer though must be limited in size to prevent OOM



_Code quote:_
```XML
                    } else if e.is_full() {
                        // TODO(guyn): replace panic with more graceful error handling
                        panic!(
                            "Sender is full, dropping the message. StreamId: {}, MessageId: {}",
                            message.stream_id, message.message_id
                        );
```
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler.rs` line 313 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2730#-OEKsdur6ObQwastQSBX:-OEKsdur6ObQwastQSBY:b-7zi8ni) ([raw file](https://github.com/starkware-libs/sequencer/blob/75eaa43f8feb5fc2cc8d7bcc1947d609a9e3fce3/crates/sequencing/papyrus_consensus/src/stream_handler.rs#L313)):*
> ```XML
>             Ordering::Equal => {
>                 // success will be false if the receiver is closed.
>                 let mut success = Self::inbound_send(data, message);
> ```

WDYT?



_Suggestion:_
```XML
                let mut receiver_unsubscribed = Self::inbound_send(data, message);
```

guy-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2730)* status: all files reviewed, 5 unresolved discussions (waiting on @matan-starkware)

___
*[`crates/sequencing/papyrus_consensus/src/stream_handler.rs` line 204 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2730#-OEKsCBe2zjX8sEdqNUc:-OEL3Iy47ofkMOB9snca:bksmzzn) ([raw file](https://github.com/starkware-libs/sequencer/blob/75eaa43f8feb5fc2cc8d7bcc1947d609a9e3fce3/crates/sequencing/papyrus_consensus/src/stream_handler.rs#L204)):*

Good catch. I would want this to just stay in the internal buffer for this stream_id (where we keep the out of order messages). I don't see this as an error. That buffer though must be limited in size to prevent OOM
</blockquote></details>

Good idea! 
It looks like this may be a little more complicated because of the way the code is written right now. Could require a bit of a refactor to make this work. 
Also, I'm not sure what happens after we buffer that message. If no other messages come in, that buffer will never get triggered and you may end up with all messages buffered forever. 
I'm putting a TODO on this option and we'll do it on the next PR. 
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler.rs` line 313 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2730#-OEKsdur6ObQwastQSBX:-OEL4cva46EHwmfhQM7j:bwwec88) ([raw file](https://github.com/starkware-libs/sequencer/blob/75eaa43f8feb5fc2cc8d7bcc1947d609a9e3fce3/crates/sequencing/papyrus_consensus/src/stream_handler.rs#L313)):*

WDYT?
</blockquote></details>

yeah, I was feeling a bit uneasy about "success". I'll take your idea. Maybe "receiver_dropped" is more obvious? Because as far as I can tell, the only way to close the channel from that end is to drop it. 
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 412 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2730#-OEKsxKl3SloWZMVj4Ep:-OEL5IEH6GdISGA-7yCA:b-iq7gae) ([raw file](https://github.com/starkware-libs/sequencer/blob/75eaa43f8feb5fc2cc8d7bcc1947d609a9e3fce3/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L412)):*

I think often it's better not to have timeouts like this in a test. Normally tests are run with a timeout on the whole test, and so this explicit timeout is just adding complexity to your test code.
</blockquote></details>

I don't know how to make the stream handler's `run()` function quit gracefully, so instead I'm using a timeout. This is not "in case things run too long" this is "after messages are consumed, quit the run". 
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 433 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2730#-OEKtKFQ5iasmVHXtIrK:-OEL5Wtc5S6RhEMfwdeC:bcw54vt) ([raw file](https://github.com/starkware-libs/sequencer/blob/75eaa43f8feb5fc2cc8d7bcc1947d609a9e3fce3/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L433)):*

this is the follow up "ignore late messages"?
</blockquote></details>

Exactly. 
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler.rs` line 185 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2730#-OEKs1G48fg6atRTi4UJ:-OEL3B_m9z1z4QSurFet:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/75eaa43f8feb5fc2cc8d7bcc1947d609a9e3fce3/crates/sequencing/papyrus_consensus/src/stream_handler.rs#L185)):*
> ```XML
>     }
> 
>     fn inbound_send(data: &mut StreamData<T>, message: StreamMessage<T>) -> bool {
> ```

Done.

matan-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2730)* status: all files reviewed, 2 unresolved discussions (waiting on @guy-starkware)

___
*[`crates/sequencing/papyrus_consensus/src/stream_handler.rs` line 204 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2730#-OEKsCBe2zjX8sEdqNUc:-OEPFmE0B28FB8isCxwX:b-sr8zia) ([raw file](https://github.com/starkware-libs/sequencer/blob/75eaa43f8feb5fc2cc8d7bcc1947d609a9e3fce3/crates/sequencing/papyrus_consensus/src/stream_handler.rs#L204)):*
<details><summary><i>Previously, guy-starkware wrote…</i></summary><blockquote>

Good idea! 
It looks like this may be a little more complicated because of the way the code is written right now. Could require a bit of a refactor to make this work. 
Also, I'm not sure what happens after we buffer that message. If no other messages come in, that buffer will never get triggered and you may end up with all messages buffered forever. 
I'm putting a TODO on this option and we'll do it on the next PR.
</blockquote></details>

Yeah I more just mean the TODO should be updated to this instead of triggering an error. We can discuss design later.
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler.rs` line 314 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/2730#-OEPG1pG0TGnEB_ZG6TJ:-OEPG1pG0TGnEB_ZG6TK:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/691a239cb2900d74b4fe13a98cfe2914caf87ee2/crates/sequencing/papyrus_consensus/src/stream_handler.rs#L314)):*
> ```XML
>             Ordering::Equal => {
>                 // receiver_dropped will be true if the receiver for this stream is dropped.
>                 let mut receiver_dropped = Self::inbound_send(data, message);
> ```





_Suggestion:_
```XML
                let mut receiver_dropped = Self::inbound_send(data, message);
```

guy-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2730)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @guy-starkware)

PR #2736, opened 2024-12-17
feat(consensus): set active/obserever height based on batcher height

asmaastarkware:


matan-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2736)* status: all files reviewed, 2 unresolved discussions (waiting on @asmaastarkware)

___
*[`crates/starknet_consensus_manager/src/consensus_manager.rs` line 72 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2736#-OEKm6_S0xu2DkZSOqGI:-OEKm6_S0xu2DkZSOqGJ:bkbt9c5) ([raw file](https://github.com/starkware-libs/sequencer/blob/c591619aa903a0880702ea3c51833713c097c19f/crates/starknet_consensus_manager/src/consensus_manager.rs#L72)):*
> ```XML
>                 ConsensusError::InternalNetworkError(
>                     "Failed to get height from batcher".to_string(),
>                 )
> ```

This is not a network error. I'm fine with just defining an `Other(String)` error type. Another option is for us to define another error type in this crate and have it inherit from ConsensusError and then add BatcherError. That feels more "proper," but not sure how much I care. What do you prefer?



_Code quote:_
```XML
                ConsensusError::InternalNetworkError(
                    "Failed to get height from batcher".to_string(),
                )
```
___
*[`crates/starknet_consensus_manager/src/consensus_manager.rs` line 78 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2736#-OEKmbpE71Yo-M2O_JaW:-OEKmbpF8fJqUywaxzus:b45h) ([raw file](https://github.com/starkware-libs/sequencer/blob/c591619aa903a0880702ea3c51833713c097c19f/crates/starknet_consensus_manager/src/consensus_manager.rs#L78)):*
> ```XML
>         } else {
>             BlockNumber(observer_height.0 + 1)
>         };
> ```





_Suggestion:_
```XML
        let active_height = if self.config.consensus_config.start_height == observer_height {
            // Setting `start_height` is only used to enable consensus starting immediately without observing
            // the first height. This means consensus may return to a height it already voted in, risking
            // equivocation. This is only safe to do if we restart all nodes at this height.
            observer_height
        } else {
            BlockNumber(observer_height.0 + 1)
        };
```

asmaastarkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2736)* status: all files reviewed, 2 unresolved discussions (waiting on @matan-starkware)

___
*[`crates/starknet_consensus_manager/src/consensus_manager.rs` line 72 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2736#-OEKm6_S0xu2DkZSOqGI:-OENnMH31hQ2F2fvAhjU:b-mc2mm8) ([raw file](https://github.com/starkware-libs/sequencer/blob/c591619aa903a0880702ea3c51833713c097c19f/crates/starknet_consensus_manager/src/consensus_manager.rs#L72)):*

This is not a network error. I'm fine with just defining an `Other(String)` error type. Another option is for us to define another error type in this crate and have it inherit from ConsensusError and then add BatcherError. That feels more "proper," but not sure how much I care. What do you prefer?
</blockquote></details>

went with Other(String), but can revisit if needed..

matan-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2736)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @asmaastarkware)

PR #2779, opened 2024-12-18
refactor(ci): improve rustfmt perf and cleanup

giladchase:
- Install nightly through moonrepo, so it'll be cached.
- Always run rustfmt on --all for both push/pull. It doesn't compile anything and runs on `--all` in a few seconds.

Benchmark: seems to reduce time from ~15 seconds to ~5 seconds, and logs indeed show that nothing is installed, only rustfmt is run.

Unrelated: remove clippy/rustfmt from the components in rust-toolchain.toml, when using the `default` profile they are already included.


giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2779)* status: all files reviewed, 2 unresolved discussions (waiting on @dorimedini-starkware)

___
*[`.github/actions/install_rust/action.yml` line 12 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2779#-OEQFC0F3YmHvO6TeQW1:-OEQHy7u5lBwVAwhQ_5D:by9991n) ([raw file](https://github.com/starkware-libs/sequencer/blob/c460696ec1f0bd6c8858a339f10afe23e7fe871c/.github/actions/install_rust/action.yml#L12)):*

why do we use this specific rustfmt channel?
</blockquote></details>

Same on we had in rustfmt.sh.
Bumping it requires fixing stuff, out of scope I think?
___
*[`.github/workflows/main.yml` line 76 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2779#-OEQEkA19nP6Vlk91x4J:-OEQI8rJ12JMuy-5f5eH:b-oqbdoj) ([raw file](https://github.com/starkware-libs/sequencer/blob/c460696ec1f0bd6c8858a339f10afe23e7fe871c/.github/workflows/main.yml#L76)):*

I don't like how the channel is hard-coded here; can't we have a single point of truth? either in rust-toolchain.toml or in the moonrepo action?
</blockquote></details>

There are some hacks i can add for sharing stuff between workflows, I hope I'll be able to just move the setup-rust thingie into main.yml and this will make things much simpler.
I'll TAL tomorrow

dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2779)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

___
*[`.github/actions/install_rust/action.yml` line 12 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2779#-OEQFC0F3YmHvO6TeQW1:-OEQPKJX2F0cdi9d64OS:bv9qrs9) ([raw file](https://github.com/starkware-libs/sequencer/blob/c460696ec1f0bd6c8858a339f10afe23e7fe871c/.github/actions/install_rust/action.yml#L12)):*

Same on we had in rustfmt.sh.
Bumping it requires fixing stuff, out of scope I think?
</blockquote></details>

but now we have a double point of truth..
I'd say the "correct" order of things is first run rustfmt with our target version ("stable"? is there such a thing for rustfmt?), then make this current change

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2779)* status: all files reviewed, 1 unresolved discussion (waiting on @dorimedini-starkware)

___
*[`.github/actions/install_rust/action.yml` line 12 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2779#-OEQFC0F3YmHvO6TeQW1:-OES4ZpSAisrOvz58u8C:b-lj83gy) ([raw file](https://github.com/starkware-libs/sequencer/blob/c460696ec1f0bd6c8858a339f10afe23e7fe871c/.github/actions/install_rust/action.yml#L12)):*

but now we have a double point of truth..
I'd say the "correct" order of things is first run rustfmt with our target version ("stable"? is there such a thing for rustfmt?), then make this current change
</blockquote></details>

Ideally it should be stable ya, but we (and most other ppl) never run rustfmt on stable, only on nightly,.

This is since because development is super slow and cautious, they have a [_yearly_ release.](https://github.com/rust-lang/rustfmt/releases), and too many basic stuff have been labeled unstable for ages. 

I think it is unavoidable (and common) to install both stable and nightly, where the latter is only used for rustfmt.

Note: rustfmt doesn't compile _our_ code, so the nightly change is relevant only for rustfmt's code being compiled under nightly.
Note 2: I haven't changed our existing CI toolchain, we've always installed this nightly alongside stable, only up until now the installation was hidden in rustfmt.sh and uncached.

dorimedini-starkware (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2779#-:-OESzLba2xFxQ9_Ndiex:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2779)* status: 3 of 6 files reviewed, all discussions resolved (waiting on @dorimedini-starkware)

___
*[`.github/workflows/main.yml` line 76 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/2779#-OEQEkA19nP6Vlk91x4J:-OETstfv9F9ekv7H2Upf:bk7p714) ([raw file](https://github.com/starkware-libs/sequencer/blob/c460696ec1f0bd6c8858a339f10afe23e7fe871c/.github/workflows/main.yml#L76)):*

There are some hacks i can add for sharing stuff between workflows, I hope I'll be able to just move the setup-rust thingie into main.yml and this will make things much simpler.
I'll TAL tomorrow
</blockquote></details>

Added it as an input to the action.

Not loving it, but it's the way to go given our current setup.

Let's just make sure we don't start passing too many variables around for logic, only for infra purposes if all else fails and stuff

dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2779)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

___
*[`.github/actions/install_rust/action.yml` line 18 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/2779#-OEUAqox3Lfgqxfdnb_u:-OEUAqox3Lfgqxfdnb_v:bo5tvwx) ([raw file](https://github.com/starkware-libs/sequencer/blob/d7395945068a8581ee313ebeacd1ab8d8d379144/.github/actions/install_rust/action.yml#L18)):*
> ```YAML
>         bins: taplo-cli@0.9.0, cargo-machete
>         # Install additional non-default toolchains (for rustfmt for example), NOP if input omitted.
>         channel: ${{ inputs.extra_rust_toolchains }}
> ```

is it really a NOOP if the arg is empty? did you test what happens if no arg is provided?



_Code quote:_
```YAML
channel: ${{ inputs.extra_rust_toolchains }}
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2779)* status: all files reviewed, 1 unresolved discussion (waiting on @dorimedini-starkware)

___
*[`.github/actions/install_rust/action.yml` line 18 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/2779#-OEUAqox3Lfgqxfdnb_u:-OEUoJ_P4IfsXG3RpClG:br647s6) ([raw file](https://github.com/starkware-libs/sequencer/blob/d7395945068a8581ee313ebeacd1ab8d8d379144/.github/actions/install_rust/action.yml#L18)):*

is it really a NOOP if the arg is empty? did you test what happens if no arg is provided?
</blockquote></details>

Yep, in a separate PRand also in this PR: compare the bootstrap step in the `run-tests:` job vs the one in the `code_style:` job, in the latter moonrepo gets a a `channel` arg and installs it as well as 1.83, whereas in the former moonrepo doesn't get a channel arg and only 1.83 is installed.
(The moonrepo part inside bootstrap is folded, you'll need to open it to see the args)

dorimedini-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/2779)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

PR #1714, opened 2024-10-31
feat: allow StreamHandler to use a generic StreamId

guy-starkware:
The stream ID is used to uniquely separate streams coming from the same peer. In the use case of consensus, this can be the `(height, round)` we are in. 

In general, the `stream_id` should be convertible into a `Vec<u8>>` and should have some other traits like `Display` and `Clone`, etc.

PR #3052, opened 2025-01-01
feat(consensus): limit number of proposal streams per peer

guy-starkware:
Putting a limit on the number of streams that can be held open (for each peer individually) helps solve a lot of possible problems, such as dropped streams. 

This is the first in a series of improvements dealing with failure modes of StreamHandler.

### DIFF
--- a/crates/starknet_consensus/src/stream_handler.rs
+++ b/crates/starknet_consensus/src/stream_handler.rs
@@ -5,9 +5,11 @@ use std::collections::hash_map::Entry::{Occupied, Vacant};
 use std::collections::HashMap;
 use std::fmt::{Debug, Display};
 use std::hash::Hash;
+use std::num::NonZeroUsize;
 
 use futures::channel::mpsc;
 use futures::StreamExt;
+use lru::LruCache;
 use papyrus_network::network_manager::{
     BroadcastTopicClient,
     BroadcastTopicClientTrait,
@@ -26,7 +28,9 @@ mod stream_handler_test;
 type PeerId = OpaquePeerId;
 type MessageId = u64;
 
+// TODO(guyn): add all of these to the config
 const CHANNEL_BUFFER_LENGTH: usize = 100;
+const MAX_STREAMS_PER_PEER: NonZeroUsize = NonZeroUsize::new(10).unwrap();
 
 /// A combination of trait bounds needed for the content of the stream.
 pub trait StreamContentTrait:
@@ -105,10 +109,10 @@ pub struct StreamHandler<StreamContent: StreamContentTrait, StreamId: StreamIdTr
     inbound_channel_sender: mpsc::Sender<mpsc::Receiver<StreamContent>>,
     // This receives messages from the network.
     inbound_receiver: BroadcastTopicServer<StreamMessage<StreamContent, StreamId>>,
-    // A map from (peer_id, stream_id) to a struct that contains all the information
+    // A map from peer_id and stream_id to a struct that contains all the information
     // about the stream. This includes both the message buffer and some metadata
-    // (like the latest message ID).
-    inbound_stream_data: HashMap<(PeerId, StreamId), StreamData<StreamContent, StreamId>>,
+    // (like the latest message ID). The mapping is {peer_id: {stream_id: StreamData}}.
+    inbound_stream_data: HashMap<PeerId, LruCache<StreamId, StreamData<StreamContent, StreamId>>>,
     // Whenever application wants to start a new stream, it must send out a
     // (stream_id, Receiver) pair. Each receiver gets messages that should
     // be sent out to the network.
@@ -313,19 +317,34 @@ impl<StreamContent: StreamContentTrait, StreamId: StreamIdTrait>
 
         let peer_id = metadata.originator_id.clone();
         let stream_id = message.stream_id.clone();
-        let key = (peer_id, stream_id);
 
-        let data = match self.inbound_stream_data.entry(key.clone()) {
+        let data = match self.inbound_stream_data.entry(peer_id.clone()) {
             // If data exists, remove it (it will be returned to hash map at end of function).
-            Occupied(entry) => entry.remove_entry().1,
+            Occupied(mut entry) => {
+                // If we received a message for a stream_id that we have not seen before,
+                // we need to create a new receiver for it.
+                let data = entry.get_mut().pop(&stream_id).unwrap_or_else(|| StreamData::new());
+                // If the Lru cache is left empty, remove it.
+                if entry.get().is_empty() {
+                    entry.remove();
+                }
+                data
+            }
             Vacant(_) => {
-                // If we received a message for a stream that we have not seen before,
+                // If we received a message for a peer_id that we have not seen before,
                 // we need to create a new receiver for it.
                 StreamData::new()
             }
         };
         if let Some(data) = self.handle_message_inner(message, metadata, data) {
-            self.inbound_stream_data.insert(key, data);
+            let existing_data = self
+                .inbound_stream_data
+                .entry(peer_id)
+                .or_insert_with(|| LruCache::new(MAX_STREAMS_PER_PEER))
+                .put(stream_id, data);
+            if existing_data.is_some() {
+                panic!("This stream data should have been removed from the cache!");
+            }
         }
     }
 
--- a/crates/starknet_consensus/src/stream_handler_test.rs
+++ b/crates/starknet_consensus/src/stream_handler_test.rs
@@ -16,7 +16,7 @@ use papyrus_protobuf::converters::ProtobufConversionError;
 use papyrus_test_utils::{get_rng, GetTestInstance};
 use prost::DecodeError;
 
-use super::{MessageId, StreamHandler};
+use super::{MessageId, StreamHandler, MAX_STREAMS_PER_PEER};
 
 const TIMEOUT: Duration = Duration::from_millis(100);
 const CHANNEL_SIZE: usize = 100;
@@ -56,12 +56,13 @@ impl Ord for TestStreamId {
 
 impl Display for TestStreamId {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "TestStreamId({})", self.0)
+        self.0.fmt(f)
     }
 }
 
-#[cfg(test)]
 mod tests {
+    use std::collections::HashMap;
+
     use papyrus_protobuf::consensus::{IntoFromProto, ProposalInit, ProposalPart};
 
     use super::*;
@@ -74,6 +75,18 @@ mod tests {
         let content = match fin {
             true => StreamMessageBody::Fin,
             false => StreamMessageBody::Content(ProposalPart::Init(ProposalInit::default())),
+        };
+        StreamMessage { message: content, stream_id, message_id }
+    }
+
+    fn make_test_message_with_stream_id(
+        stream_id: TestStreamId,
+        message_id: MessageId,
+        fin: bool,
+    ) -> StreamMessage<TestStreamId, TestStreamId> {
+        let content = match fin {
+            true => StreamMessageBody::Fin,
+            false => StreamMessageBody::Content(stream_id),
         };
         StreamMessage { message: content, stream_id, message_id }
     }
@@ -225,21 +238,31 @@ mod tests {
 
         assert_eq!(stream_handler.inbound_stream_data.len(), 1);
         assert_eq!(
-            stream_handler.inbound_stream_data[&(peer_id.clone(), stream_id)].message_buffer.len(),
+            stream_handler.inbound_stream_data[&peer_id]
+                .peek(&stream_id)
+                .unwrap()
+                .message_buffer
+                .len(),
             5
         );
         // Still waiting for message 0.
         assert_eq!(
-            stream_handler.inbound_stream_data[&(peer_id.clone(), stream_id)].next_message_id,
+            stream_handler.inbound_stream_data[&peer_id].peek(&stream_id).unwrap().next_message_id,
             0
         );
         // Has a receiver, waiting to be sent when message 0 is received.
         assert!(
-            stream_handler.inbound_stream_data[&(peer_id.clone(), stream_id)].receiver.is_some()
+            stream_handler.inbound_stream_data[&peer_id]
+                .peek(&stream_id)
+                .unwrap()
+                .receiver
+                .is_some()
         );
 
         let range: Vec<u64> = (1..6).collect();
-        let keys: Vec<u64> = stream_handler.inbound_stream_data[&(peer_id, stream_id)]
+        let keys: Vec<u64> = stream_handler.inbound_stream_data[&peer_id]
+            .peek(&stream_id)
+            .unwrap()
             .message_buffer
             .keys()
             .copied()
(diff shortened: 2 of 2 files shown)


guy-starkware:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/3052?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3052** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/3052?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/3052?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#1714** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/1714?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

matan-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3052)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @guy-starkware)

matan-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3052)* status: all files reviewed, 1 unresolved discussion (waiting on @guy-starkware)

___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 496 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OGUcZnEAAlojCg2BVyZ:-OGUcZnEAAlojCg2BVy_:bc6y1mq) ([raw file](https://github.com/starkware-libs/sequencer/blob/da53ea2367965933f9896b394dc1fcb8befdae31/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L496)):*
> ```XML
> 
>         // The total number of streams should still be MAX_STREAMS_PER_PEER.
>         assert_eq!(stream_handler.inbound_stream_data[&peer_id].len(), MAX_STREAMS_PER_PEER);
> ```

Returning to our slack conversation about testing, this is not what we should do. We need to test visible behavior:
- Have a single peer_id send `MAX_STREAMS_PER_PEER+1` streams
- Check that only the first MAX_STREAMS_PER_PEER streams get output



_Code quote:_
```XML
        // The total number of streams should still be MAX_STREAMS_PER_PEER.
        assert_eq!(stream_handler.inbound_stream_data[&peer_id].len(), MAX_STREAMS_PER_PEER);
```

guy-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3052)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @matan-starkware)

___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 496 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OGUcZnEAAlojCg2BVyZ:-OGV8AXn-5j2s286waFX:b-8p019j) ([raw file](https://github.com/starkware-libs/sequencer/blob/da53ea2367965933f9896b394dc1fcb8befdae31/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L496)):*

Returning to our slack conversation about testing, this is not what we should do. We need to test visible behavior:
- Have a single peer_id send `MAX_STREAMS_PER_PEER+1` streams
- Check that only the first MAX_STREAMS_PER_PEER streams get output
</blockquote></details>

OK, did a major refactor of this test. It now works without inspecting the inner parts of StreamHandler. 

I will do another PR at the top of the stack to refactor the other tests. I may be able to get rid of the "run StreamHandler with a timeout" but that is not certain. We'll see how it goes.

asmaastarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3052)* status: all files reviewed, 3 unresolved discussions (waiting on @guy-starkware and @matan-starkware)

___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 477 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OGZE72a0fcWZUaiw26O:-OGZE72a0fcWZUaiw26P:bs0esvw) ([raw file](https://github.com/starkware-libs/sequencer/blob/a363f4408d3d8198936a1f7597c1d3347e3fe41d/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L477)):*
> ```XML
>         // let peer_id = inbound_metadata.originator_id.clone();
>         // let inbound_metadata2 = BroadcastedMessageMetadata::get_test_instance(&mut get_rng());
>         // let peer_id2 = inbound_metadata2.originator_id.clone();
> ```

please delete



_Code quote:_
```XML
        // let peer_id = inbound_metadata.originator_id.clone();
        // let inbound_metadata2 = BroadcastedMessageMetadata::get_test_instance(&mut get_rng());
        // let peer_id2 = inbound_metadata2.originator_id.clone();
```
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 485 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OGZECOs1KEjf6nsBhAx:-OGZECOs1KEjf6nsBhAy:bsg75fh) ([raw file](https://github.com/starkware-libs/sequencer/blob/a363f4408d3d8198936a1f7597c1d3347e3fe41d/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L485)):*
> ```XML
>         // Sending in reverse order of streams will mean none of the streams 1 to 10 will get
>         // dropped. When stream 0 gets message 0, it will be a new stream (dropping 10) but
>         // it will only have message 0, we will not get messages 1 and 2.
> ```

use `MAX_STREAMS_PER_PEER` instead of 10



_Code quote:_
```XML
        // Send too many streams from the same peer. Send messages 1 to 3 on all channels.
        // Note that message 3 is Fin and doesn't get to the receiver (it closes it!).
        // Channel 0 is the last one sent, so it should be dropped when reaching channel 10.
        // Then send 0 on all channels, in reverse order, which will release the buffered messages.
        // Sending in reverse order of streams will mean none of the streams 1 to 10 will get
        // dropped. When stream 0 gets message 0, it will be a new stream (dropping 10) but
        // it will only have message 0, we will not get messages 1 and 2.
```

guy-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3052)* status: all files reviewed, 3 unresolved discussions (waiting on @asmaastarkware and @matan-starkware)

___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 477 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OGZE72a0fcWZUaiw26O:-OGZIO0h9nKP5zMlkBw4:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/a363f4408d3d8198936a1f7597c1d3347e3fe41d/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L477)):*
<details><summary><i>Previously, asmaastarkware (asmaa-starkware) wrote…</i></summary><blockquote>

please delete
</blockquote></details>

Done.
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 485 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OGZECOs1KEjf6nsBhAx:-OGZIbR53k-TLKRiHdHm:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/a363f4408d3d8198936a1f7597c1d3347e3fe41d/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L485)):*
<details><summary><i>Previously, asmaastarkware (asmaa-starkware) wrote…</i></summary><blockquote>

use `MAX_STREAMS_PER_PEER` instead of 10
</blockquote></details>

Done.

matan-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3052)* status: all files reviewed, 2 unresolved discussions (waiting on @asmaastarkware)

matan-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3052)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @dan-starkware and @giladchase)

guy-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3052)* status: all files reviewed, 6 unresolved discussions (waiting on @dan-starkware and @giladchase)

___
*[`crates/sequencing/papyrus_consensus/src/stream_handler.rs` line 306 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OH7b-xAA2oJrYh0tBHT:-OHCK50V3hrkNQLlRC7X:b-15ddq1) ([raw file](https://github.com/starkware-libs/sequencer/blob/034a3d069f0636aa289c61ab6157c72d363dc9b8/crates/sequencing/papyrus_consensus/src/stream_handler.rs#L306)):*

Out of scope questions, non-blocking:

1.  can `StreamMessage<StreamContent, StreamId>` be `type`defed?  It seems like it's always the same 2 generic args in this file.
2. Are the metadata and message instances always shared together? Or is there a usecase where only one is created and not the other. In particular, what happens if the streammessage is an Error, will the metadata be well-defined? (Asking cause it might make sense to make convert the tuple into a struct and formalize the marriage).
</blockquote></details>

1. You mean doing something like `type MyStreamMessage = StreamMessage<ProposalPart, HeightAndRound>`?
2. I think the `(message, metadata)` tuple is defined by the network crate. I'm not sure if we can change it. 
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler.rs` line 321 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OH7buWK7lPXCv4LJHo3:-OHCSL5CCRoYzXsjWeoy:bspq32d) ([raw file](https://github.com/starkware-libs/sequencer/blob/865bed0ef7363668d2d572be5909b51a01215152/crates/sequencing/papyrus_consensus/src/stream_handler.rs#L321)):*

`peer_id` should be `Copy`, its a newtype around a `Copy`.
Make sure though that the Copy is sensible, and not some huge struct that just happens to qualify for `Copy` status)
</blockquote></details>

The PeerId is an alias for OpaquePeerId which is defined in the network crate. I'm not sure if I should change it there or somehow wrap around it with something that is "Copy"? 
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler.rs` line 348 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OH7fFTu69rWkRkA7Xck:-OHClxiu-pba6Nl5Lkea:b-hnom3s) ([raw file](https://github.com/starkware-libs/sequencer/blob/865bed0ef7363668d2d572be5909b51a01215152/crates/sequencing/papyrus_consensus/src/stream_handler.rs#L348)):*

Is the success of this op significant? (asking cause we're not checking the return value)
</blockquote></details>

I added a panic in case the key already exists (it should never happen). 
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 55 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OH7hGF90NTxFHgWMltE:-OHCm7_h0yp4KJMlN7ph:b-vm2pbm) ([raw file](https://github.com/starkware-libs/sequencer/blob/034a3d069f0636aa289c61ab6157c72d363dc9b8/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L55)):*

Out of scope, nonblocking: Is this identical to the `derive` version?
</blockquote></details>

I seem to recall that you have to specify how the comparison is done, because there is no way to derive this without knowing which fields should be used (or in what order). I am not sure if that's what you are asking... 
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 64 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OH7fnk1DkOJgfJ0ZDwe:-OHCnR6fEfC2sTn921T8:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/034a3d069f0636aa289c61ab6157c72d363dc9b8/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L64)):*

I don't think this is necessary, this whole file is already `cfg(test)`'d.
When I saw it I felt I had to double-check that this was the case, which I suspect other readers might as well.
This is because this pattern is mostly common (AFAIK) for production code (though we don't use it either :sweat_smile:  cause it recompiles the whole code in the file even if changes are only in the test.
</blockquote></details>

Done.
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 93 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OH7jBVw973otLclbSHN:-OHCnzQR18ILtRwV6Acz:b-omw4zl) ([raw file](https://github.com/starkware-libs/sequencer/blob/865bed0ef7363668d2d572be5909b51a01215152/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L93)):*

Naming is a bit confusing, since both constructors take stream_id as an arg.

This is because`with_X` typically implies an arg X is passed, due to the prevalence of the Builder pattern.
BTW, builder might fit here, not sure yet: you have two flavors of messages (?), and two types of message states (fin or not-fin). 

Also looks like the Fin message is the same in both flavors, so I think it can just be its own function `make_fin`.
</blockquote></details>

Should I rename it `make_test_message_stream_id_content` to make it more clear? I tried to leave the old function so I wouldn't have to change the existing tests. 

Also, I'm not sure what I would gain from `make_fin`? I use this function often inside a loop where I can say `i==10` sends a fin and all other values send a regular message. Having a `make_fin` will make me have to add a conditional to all those loops. 
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 105 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OH7liuO8TcsOHN5oeJO:-OHCoRtF4S9s4GnU4Q1J:b-4h0eta) ([raw file](https://github.com/starkware-libs/sequencer/blob/034a3d069f0636aa289c61ab6157c72d363dc9b8/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L105)):*

Out of scope, nonblocking: this is a lot of overhead for an assert. Also callsites seem to create vectors already, which then get dereffed into slices here, only to be re-allocated as vectors for the sort.
</blockquote></details>

I think I've already had this discussion on a higher level of this stack... but it doesn't matter because this whole thing will go away towards the end of the stack. 
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 124 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OH7mjuzB-x_NiygxiX7:-OHCotxX0yfCZIgncjry:bxrfh7v) ([raw file](https://github.com/starkware-libs/sequencer/blob/865bed0ef7363668d2d572be5909b51a01215152/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L124)):*

Out of scope, nonblocking: this should be a struct + remove the clippy warning.
</blockquote></details>

So something like `InputParameters { ... }`? I had the impression that in tests it was ok to leave complex inputs, no? 
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 129 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OH7t65198uTFzD_aQfX:-OHCpQeq6elj5j8BUlXl:b-oh37ig) ([raw file](https://github.com/starkware-libs/sequencer/blob/865bed0ef7363668d2d572be5909b51a01215152/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L129)):*

can this be `T: StreamContentTrait`? Ditto for other place with this bound
</blockquote></details>

I will use `IntoFromProto` from another crate that I added a few PRs ago exactly for this sort of thing. 
___
*[`crates/sequencing/papyrus_consensus/src/stream_handler_test.rs` line 566 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3052#-OH7ns5z7FFytk5XxnkO:-OHCqWNTBveW5gLdXmDJ:b-78lqla) ([raw file](https://github.com/starkware-libs/sequencer/blob/865bed0ef7363668d2d572be5909b51a01215152/crates/sequencing/papyrus_consensus/src/stream_handler_test.rs#L566)):*

I think this will only show `got false expected true` on fail, which makes it harder to debug.
See if you can make this an assert_eq somehow,  since that shows both types on failure.
</blockquote></details>

Just theoretically, it would also be possible to just add an error message, right? Something like
`assert!(...containts(...), "inbound_stream_data does not contain {}", stream_id)`? 

I say theoretically because this assert is also going away in one of the next PRs.

PR #3098, opened 2025-01-03
dev: Minor SNOS changes

notlesh:
This is a small set of changes we have been using on [SNOS](https://github.com/keep-starknet-strange/snos/) for quite some time.

As they are probably only useful for us, we have maintained our own fork and ported them during upgrades. However, for projects which use SNOS as a dependency, this inevitably creates duplicated versions of `blockifier` in the dependency tree, so it would be great to get them included in the main repo.

### Backporting

While this PR proposes these changes for `main`, it would help more immediately to backport these into the current `v0.8.0-rc.3` release and roll a new release on `crates.io`. I am not sure what the best way to approach this would be, so please provide some feedback :) 

The "compare" here shows what is intended: https://github.com/starkware-libs/sequencer/compare/blockifier-v0.8.0-rc.3...Moonsong-Labs:sequencer:notlesh/blockifier-v0.8.0-rc.3-snos

Note that it includes one additional commit (`5ab9ae7`) which is not included in this PR because it is far more complicated against `main` and we may be able to work around it on SNOS anyway.

### DIFF
--- a/crates/blockifier/src/execution/syscalls/mod.rs
+++ b/crates/blockifier/src/execution/syscalls/mod.rs
@@ -34,7 +34,7 @@ use crate::execution::syscalls::syscall_base::SyscallResult;
 use crate::versioned_constants::{EventLimits, VersionedConstants};
 
 pub mod hint_processor;
-mod secp;
+pub mod secp;
 pub mod syscall_base;
 
 #[cfg(test)]
--- a/crates/blockifier/src/execution/syscalls/secp.rs
+++ b/crates/blockifier/src/execution/syscalls/secp.rs
@@ -72,7 +72,7 @@ where
         id
     }
 
-    fn get_point_by_id(
+    pub fn get_point_by_id(
         &self,
         ec_point_id: Felt,
     ) -> SyscallResult<&short_weierstrass::Affine<Curve>> {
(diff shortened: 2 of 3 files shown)


dorimedini-starkware (changes requested):
I think backporting to 0.8.0 (blockifier for starknet 0.13.2 and 0.13.3) will not be feasible; let's discuss offline the problem(s) this causes

> Reviewed 1 of 2 files at r1, 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3098#-:-OGOofIj7PqNO-_m7Ew1:b-dozxd1)* status: all files reviewed, 1 unresolved discussion (waiting on @notlesh)

___
*[`crates/blockifier/src/execution/syscalls/secp.rs` line 175 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3098#-OGOmzzp-ztFuSdHYK2Z:-OGOmzzp-ztFuSdHYK2_:b3d6pqn) ([raw file](https://github.com/starkware-libs/sequencer/blob/3ef812fd3e26a2ef4511fe726a2b49e7f1c88d9c/crates/blockifier/src/execution/syscalls/secp.rs#L175)):*
> ```XML
>     // The parity of the y coordinate, assuming a point with the given x coordinate exists.
>     // True means the y coordinate is odd.
>     pub y_parity: bool,
> ```

please revert and add getters instead, we don't want devs to be able to create instances manually.
if this is a requirement - add a `new_unchecked` associated function behind `cfg(test)` or `cfg(any(test, feature = "testing"))`



_Code quote:_
```XML
    pub x: BigUint,
    // The parity of the y coordinate, assuming a point with the given x coordinate exists.
    // True means the y coordinate is odd.
    pub y_parity: bool,
```

PR #3156, opened 2025-01-07
chore(ci): use cargo release cache when running with the release flag

adi-yakovian-starkware:


adi-yakovian-starkware:
FYI - @RoeiE-starkware

PR #3208, opened 2025-01-09
feat(starknet_l1_provider): add `SoftDeleteIndexMap`

giladchase:
Previous implementation inside transaction manager didn't account for
 validate, the current implementation is more general, and can in the
 past perhaps be reused elsewhere after some changes are made.

### DIFF
--- a/crates/starknet_l1_provider/src/l1_provider_tests.rs
+++ b/crates/starknet_l1_provider/src/l1_provider_tests.rs
@@ -62,7 +62,7 @@ fn validate_happy_flow() {
     // Transaction wasn't deleted after the validation.
     assert_eq!(
         l1_provider.validate(tx_hash!(1), BlockNumber(1)).unwrap(),
-        ValidationStatus::Validated
+        ValidationStatus::AlreadyIncludedInPropsedBlock
     );
 }
 
--- a/crates/starknet_l1_provider/src/lib.rs
+++ b/crates/starknet_l1_provider/src/lib.rs
@@ -4,6 +4,8 @@ pub mod l1_provider;
 pub mod l1_scraper;
 pub(crate) mod transaction_manager;
 
+mod soft_delete_index_map;
+
 #[cfg(test)]
 pub mod test_utils;
 
--- a/crates/starknet_l1_provider/src/soft_delete_index_map.rs
+++ b/crates/starknet_l1_provider/src/soft_delete_index_map.rs
@@ -1,0 +1,135 @@
+use std::collections::HashSet;
+
+use indexmap::map::Entry;
+use indexmap::IndexMap;
+use starknet_api::executable_transaction::L1HandlerTransaction;
+use starknet_api::transaction::TransactionHash;
+
+/// An IndexMap that supports soft deletion of entries.
+/// Entries marked as deleted remain hidden in the map, allowing for potential recovery,
+/// selective permanent deletion, or rollback before being purged.
+// TODO: replace with a fully generic struct if there's a need for it.
+// TODO: replace with a BTreeIndexMap if commit performance becomes an issue, see note in commit.
+#[derive(Clone, Debug, Default)]
+pub struct SoftDeleteIndexMap {
+    txs: IndexMap<TransactionHash, TransactionEntry>,
+    staged_txs: HashSet<TransactionHash>,
+}
+
+impl SoftDeleteIndexMap {
+    pub fn _new() -> Self {
+        Self::default()
+    }
+
+    /// Inserts a transaction into the map, returning the previous transaction if it existed.
+    pub fn _insert(&mut self, tx: L1HandlerTransaction) -> bool {
+        let tx_hash = tx.tx_hash;
+        match self.txs.entry(tx_hash) {
+            Entry::Occupied(entry) => {
+                assert_eq!(entry.get().transaction, tx);
+                false
+            }
+            Entry::Vacant(entry) => {
+                entry.insert(TransactionEntry::new(tx));
+                true
+            }
+        }
+    }
+
+    /// Soft delete and return a reference to the first unstaged transaction, by insertion order.
+    pub fn soft_pop_front(&mut self) -> Option<&L1HandlerTransaction> {
+        let entry = self.txs.iter().find(|(_, tx)| tx.is_available());
+        let (&tx_hash, _) = entry?;
+        self.soft_remove(tx_hash)
+    }
+
+    /// Stages the given transaction with the given hash if it exists and is not already staged, and
+    /// returns a reference to it.
+    pub fn soft_remove(&mut self, tx_hash: TransactionHash) -> Option<&L1HandlerTransaction> {
+        let entry = self.txs.get_mut(&tx_hash)?;
+
+        if !entry.is_available() {
+            return None;
+        }
+
+        assert_eq!(self.staged_txs.get(&tx_hash), None);
+        entry.set_state(TxState::Staged);
+        self.staged_txs.insert(tx_hash);
+
+        Some(&entry.transaction)
+    }
+
+    /// Commits given transactions by removing them entirely and returning the removed transactions.
+    /// Uncommitted staged transactions are rolled back to unstaged first.
+    // Performance note: This operation is linear time with both the number
+    // of known transactions and the number of committed transactions. This is assumed to be
+    // good enough while l1-handler numbers remain low, but if this changes and we need log(n)
+    // removals (amortized), replace indexmap with this (basically a BTreeIndexMap):
+    // BTreeMap<u32, TransactionEntry>, Hashmap<TransactionHash, u32> and a counter: u32, such that
+    // every new tx is inserted to the map with key counter++ and the counter is not reduced
+    // when removing entries. Once the counter reaches u32::MAX/2 we recreate the DS in Theta(n).
+    pub fn _commit(&mut self, tx_hashes: &[TransactionHash]) -> Vec<L1HandlerTransaction> {
+        self._rollback_staging();
+        let tx_hashes: HashSet<_> = tx_hashes.iter().copied().collect();
+        if tx_hashes.is_empty() {
+            return Vec::new();
+        }
+
+        // NOTE: this takes Theta(|self.txs|), see docstring.
+        let (committed, not_committed): (Vec<_>, Vec<_>) =
+            self.txs.drain(..).partition(|(hash, _)| tx_hashes.contains(hash));
+        self.txs.extend(not_committed);
+
+        committed.into_iter().map(|(_, entry)| entry.transaction).collect()
+    }
+
+    /// Rolls back all staged transactions, converting them to unstaged.
+    pub fn _rollback_staging(&mut self) {
+        for tx_hash in self.staged_txs.drain() {
+            self.txs.entry(tx_hash).and_modify(|entry| entry.set_state(TxState::Unstaged));
+        }
+    }
+
+    pub fn is_staged(&self, tx_hash: &TransactionHash) -> bool {
+        self.staged_txs.contains(tx_hash)
+    }
+}
+
+impl From<Vec<L1HandlerTransaction>> for SoftDeleteIndexMap {
+    fn from(txs: Vec<L1HandlerTransaction>) -> Self {
+        let txs = txs.into_iter().map(|tx| (tx.tx_hash, TransactionEntry::new(tx))).collect();
+        SoftDeleteIndexMap { txs, ..Default::default() }
+    }
+}
+
+/// Indicates whether a transaction is unstaged or staged.
+#[derive(Debug, Clone)]
+enum TxState {
+    Unstaged,
+    Staged,
+}
+
+/// Wraps an L1HandlerTransaction along with its current TxState,
+/// and provides convenience methods for stage/unstage.
+#[derive(Debug, Clone)]
+struct TransactionEntry {
+    pub transaction: L1HandlerTransaction,
+    pub state: TxState,
+}
+
+impl TransactionEntry {
+    pub fn new(transaction: L1HandlerTransaction) -> Self {
+        Self { transaction, state: TxState::Unstaged }
+    }
+
+    pub fn set_state(&mut self, state: TxState) {
+        self.state = state
+    }
+
+    pub fn is_available(&self) -> bool {
+        match self.state {
+            TxState::Unstaged => true,
+            TxState::Staged => false,
+        }
+    }
+}
--- a/crates/starknet_l1_provider/src/test_utils.rs
+++ b/crates/starknet_l1_provider/src/test_utils.rs
@@ -1,8 +1,8 @@
+use std::collections::HashSet;
 use std::mem;
 use std::sync::Mutex;
 
 use async_trait::async_trait;
-use indexmap::{IndexMap, IndexSet};
 use pretty_assertions::assert_eq;
 use starknet_api::block::BlockNumber;
 use starknet_api::executable_transaction::{
@@ -18,8 +18,10 @@ use starknet_l1_provider_types::{
 };
 
 use crate::l1_provider::L1Provider;
+use crate::soft_delete_index_map::SoftDeleteIndexMap;
 use crate::transaction_manager::TransactionManager;
 use crate::ProviderState;
+
 // Represents the internal content of the L1 provider for testing.
 // Enables customized (and potentially inconsistent) creation for unit testing.
 #[derive(Debug, Default)]
@@ -32,10 +34,7 @@ pub struct L1ProviderContent {
 impl From<L1ProviderContent> for L1Provider {
     fn from(content: L1ProviderContent) -> L1Provider {
         L1Provider {
-            tx_manager: content
-                .tx_manager_content
-                .map(|tm_content| tm_content.complete_to_tx_manager())
-                .unwrap_or_default(),
+            tx_manager: content.tx_manager_content.map(Into::into).unwrap_or_default(),
             state: content.state.unwrap_or_default(),
             current_height: content.current_height,
         }
@@ -85,29 +84,34 @@ impl L1ProviderContentBuilder {
 // Enables customized (and potentially inconsistent) creation for unit testing.
 #[derive(Debug, Default)]
 struct TransactionManagerContent {
-    txs: Option<IndexMap<TransactionHash, L1HandlerTransaction>>,
-    committed: Option<IndexSet<TransactionHash>>,
+    txs: Option<Vec<L1HandlerTransaction>>,
+    committed: Option<HashSet<TransactionHash>>,
 }
 
-impl TransactionManagerContent {
-    fn complete_to_tx_manager(self) -> TransactionManager {
+impl From<TransactionManagerContent> for TransactionManager {
+    fn from(mut content: TransactionManagerContent) -> TransactionManager {
+        let txs: Vec<_> = mem::take(&mut content.txs).unwrap();
         TransactionManager {
-            txs: self.txs.unwrap_or_default(),
-            committed: self.committed.unwrap_or_default(),
-            ..Default::default()
+            txs: SoftDeleteIndexMap::from(txs),
+            committed: content
+                .committed
+                .unwrap_or_default()
+                .into_iter()
+                .map(|tx_hash| (tx_hash, None))
+                .collect(),
         }
     }
 }
 
 #[derive(Debug, Default)]
 struct TransactionManagerContentBuilder {
-    txs: Option<IndexMap<TransactionHash, L1HandlerTransaction>>,
-    committed: Option<IndexSet<TransactionHash>>,
+    txs: Option<Vec<L1HandlerTransaction>>,
+    committed: Option<HashSet<TransactionHash>>,
 }
 
 impl TransactionManagerContentBuilder {
     fn with_txs(mut self, txs: impl IntoIterator<Item = L1HandlerTransaction>) -> Self {
-        self.txs = Some(txs.into_iter().map(|tx| (tx.tx_hash, tx)).collect());
+        self.txs = Some(txs.into_iter().collect());
         self
     }
 
--- a/crates/starknet_l1_provider/src/transaction_manager.rs
+++ b/crates/starknet_l1_provider/src/transaction_manager.rs
@@ -1,34 +1,38 @@
-use indexmap::{IndexMap, IndexSet};
+use indexmap::IndexMap;
 use starknet_api::executable_transaction::L1HandlerTransaction;
 use starknet_api::transaction::TransactionHash;
 use starknet_l1_provider_types::ValidationStatus;
 
+use crate::soft_delete_index_map::SoftDeleteIndexMap;
+
 #[derive(Debug, Default)]
 pub struct TransactionManager {
-    pub txs: IndexMap<TransactionHash, L1HandlerTransaction>,
-    pub staged: IndexSet<TransactionHash>,
-    pub committed: IndexSet<TransactionHash>,
+    pub txs: SoftDeleteIndexMap,
+    pub committed: IndexMap<TransactionHash, Option<L1HandlerTransaction>>,
 }
 
 impl TransactionManager {
     pub fn get_txs(&mut self, n_txs: usize) -> Vec<L1HandlerTransaction> {
-        let (tx_hashes, txs): (Vec<_>, Vec<_>) = self
-            .txs
-            .iter()
-            .skip(self.staged.len()) // Transactions are proposed FIFO.
-            .take(n_txs)
-            .map(|(&hash, tx)| (hash, tx.clone()))
-            .unzip();
+        let mut txs = Vec::with_capacity(n_txs);
 
-        self.staged.extend(tx_hashes);
+        for _ in 0..n_txs {
+            match self.txs.soft_pop_front().cloned() {
+                Some(tx) => txs.push(tx),
+                None => break,
+            }
+        }
         txs
     }
 
-    pub fn validate_tx(&self, tx_hash: TransactionHash) -> ValidationStatus {
-        if self.txs.contains_key(&tx_hash) {
+    pub fn validate_tx(&mut self, tx_hash: TransactionHash) -> ValidationStatus {
+        if self.committed.contains_key(&tx_hash) {
+            return ValidationStatus::AlreadyIncludedOnL2;
+        }
+
+        if self.txs.soft_remove(tx_hash).is_some() {
             ValidationStatus::Validated
-        } else if self.committed.contains(&tx_hash) {
-            ValidationStatus::AlreadyIncludedOnL2
+        } else if self.txs.is_staged(&tx_hash) {
+            ValidationStatus::AlreadyIncludedInPropsedBlock
         } else {
             ValidationStatus::ConsumedOnL1OrUnknown
         }
--- a/crates/starknet_l1_provider_types/src/lib.rs
+++ b/crates/starknet_l1_provider_types/src/lib.rs
@@ -23,6 +23,7 @@ pub type SharedL1ProviderClient = Arc<dyn L1ProviderClient>;
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub enum ValidationStatus {
+    AlreadyIncludedInPropsedBlock,
     AlreadyIncludedOnL2,
     ConsumedOnL1OrUnknown,
     Validated,


matan-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3208)* status: all files reviewed, 11 unresolved discussions (waiting on @giladchase)

___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9uXaE2U3xM4t7vTaM:-OG9uXaE2U3xM4t7vTaN:b3kre):*
?



_Suggestion:_
```
   future perhaps be reused elsewhere after some changes are made.
```
___
*[`crates/starknet_l1_provider/src/test_utils.rs` line 86 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA9dtKASoRtaArIB-n:-OGA9dtKASoRtaArIB-o:b-1set4n) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/test_utils.rs#L86)):*
> ```XML
> 
> impl TransactionManagerContent {
>     fn complete_to_tx_manager(mut self) -> TransactionManager {
> ```

TryInto?



_Code quote:_
```XML
    fn complete_to_tx_manager(mut self) -> TransactionManager {
```
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 15 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA8luLDXBX_WlY_dOW:-OGA8luMBeL1JhjJ1YR8:b-ppypef) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L15)):*
> ```XML
> #[derive(Clone, Debug, Default)]
> pub struct SoftDeleteIndexMap {
>     txs: IndexMap<TransactionHash, TransactionEntry>,
> ```

In principle what you are trying to achieve is a priority queue? It just so happens that our current priority is first come first serve which IndexMap gives us?

Is there any plan to move into a different form of selection?



_Code quote:_
```XML
    txs: IndexMap<TransactionHash, TransactionEntry>,
```
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 21 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9uqml7ypptyDQh-oK:-OG9uqmm6nGsLTe2POwh:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L21)):*
> ```XML
> impl SoftDeleteIndexMap {
>     /// Inserts a transaction into the map, returning the previous transaction if it existed.
>     pub fn _insert(&mut self, tx: L1HandlerTransaction) -> Option<L1HandlerTransaction> {
> ```





_Suggestion:_
```XML
    pub fn insert(&mut self, tx: L1HandlerTransaction) -> Option<L1HandlerTransaction> {
```
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 23 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9vFRy1CUvQJe11NYt:-OG9vFRy1CUvQJe11NYu:bkg9hik) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L23)):*
> ```XML
>     pub fn _insert(&mut self, tx: L1HandlerTransaction) -> Option<L1HandlerTransaction> {
>         match self.txs.entry(tx.tx_hash) {
>             Entry::Occupied(entry) => Some(entry.get().transaction.clone()),
> ```

Why don't we just call `self.txs.insert`?


IIUC this like does not actually insert the new tx into the map. Rather we assume that the `tx_hash` indicates the contents are the same and we we clone the contents of the old entry.

We intentionally ignore the TxState of the existing entry?



_Code quote:_
```XML
            Entry::Occupied(entry) => Some(entry.get().transaction.clone()),
```
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 32 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA0sPmDPWmSSVvjnbX:-OGA0sPmDPWmSSVvjnbY:b-yuywfx) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L32)):*
> ```XML
> 
>     /// Soft delete and return a reference to the first unstaged transaction, by insertion order.
>     pub fn soft_pop_back(&mut self) -> Option<&L1HandlerTransaction> {
> ```

Impl looks to me like we remove the first available element, not the last



_Suggestion:_
```XML
    pub fn soft_pop_front(&mut self) -> Option<&L1HandlerTransaction> {
```
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 65 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA7ywo9q7MQMTi7VVd:-OGA7ywo9q7MQMTi7VVe:binv5up) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L65)):*
> ```XML
>     /// BTreeMap<u32, TransactionEntry>, Hashmap<TransactionHash, u32> and a counter: u32, such that
>     /// every new tx is inserted to the map with key counter++ and the counter is not reduced
>     /// when removing entries. Once the counter reaches u32::MAX/2 we recreate the DS in Theta(n).
> ```

This should not be in the public doc, switch to `//`

So my understanding:
- N = tx_hashes.len()
- M = self.txs.len()

1. Build tx_hash set - O(N)
2. Partition txs - O(M) since each contains check is O(1)
3. rebuild txs - O(M-N)
4. build committed - O(N)

Let's simplify this and say "Operation is linear time with both the number of known transactions and the number of committed transactions"



_Code quote:_
```XML
    /// Performance note: this method runs in Theta(n), since removing elements from indexmap
    /// requires shifting elements in O(n). Since we are removing multiple elements, recreating
    /// the indexmap is faster than removing each element individually.
    /// This is assumed to be good enough while l1-handler numbers remain low, but if this changes
    /// and we need log(n) removals (amortized), replace indexmap with this (basically a
    /// BTreeIndexMap):
    /// BTreeMap<u32, TransactionEntry>, Hashmap<TransactionHash, u32> and a counter: u32, such that
    /// every new tx is inserted to the map with key counter++ and the counter is not reduced
    /// when removing entries. Once the counter reaches u32::MAX/2 we recreate the DS in Theta(n).
```
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 82 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA2SxbC0dboUhTCqtl:-OGA2SxbC0dboUhTCqtm:b7esz6m) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L82)):*
> ```XML
> 
>     /// Rolls back all staged transactions, converting them to unstaged.
>     pub fn _rollback(&mut self) {
> ```

Why are pub functions underscored? That's a python thing I thought.

I think of "rollback" being for a commit, not staging. I prefer either unstage or reset_staging, or rollback_staging.



_Suggestion:_
```XML
    pub fn unstage(&mut self) {
```
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 98 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA96ouA83MhtrAgm9s:-OGA96ouA83MhtrAgm9t:b-swg57h) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L98)):*
> ```XML
>         SoftDeleteIndexMap { txs, ..Default::default() }
>     }
> }
> ```

We don't also want a `new()` fn?



_Code quote:_
```XML
impl From<Vec<L1HandlerTransaction>> for SoftDeleteIndexMap {
    fn from(txs: Vec<L1HandlerTransaction>) -> Self {
        let txs = txs.into_iter().map(|tx| (tx.tx_hash, TransactionEntry::new(tx))).collect();
        SoftDeleteIndexMap { txs, ..Default::default() }
    }
}
```
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 124 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA1E5R6OpNbeQAZpPh:-OGA1E5R6OpNbeQAZpPi:bp0xxnr) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L124)):*
> ```XML
>             self.state = TxState::Staged;
>         }
>     }
> ```

What about removing this and unstage and having a function `set_state(TxState)`?



_Suggestion:_
```XML
    pub fn stage(&mut self) {
        self.state = TxState::Staged;
    }
```
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 127 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9wCxp-C8H6sw3LZOj:-OG9wCxp-C8H6sw3LZOk:b-9hbzv9) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L127)):*
> ```XML
> 
>     pub fn is_available(&self) -> bool {
>         matches!(self.state, TxState::Unstaged)
> ```

Not blocking, I just like to avoid macros when easy to.



_Suggestion:_
```XML
        match self.state {
            TxState::Unstaged => true,
            TxState::Staged => false,
        }
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3208)* status: 3 of 6 files reviewed, 9 unresolved discussions (waiting on @matan-starkware)

___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9uXaE2U3xM4t7vTaM:-OGAFusXC3mkkMa5wDRq:b-a9ugap):*

?
</blockquote></details>

Good lord :see_no_evil: tnx
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 15 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA8luLDXBX_WlY_dOW:-OGAG2HEB1Sh8ntbESXO:bm4y34f) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L15)):*
>In principle what you are trying to achieve is a priority queue? It just so happens that our current priority is first come first serve which IndexMap gives us?

Ya for now first come first serve, don't know of any current plans of this changing any time soon.
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 23 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9vFRy1CUvQJe11NYt:-OGAHPZ02Gdufv8_IR5_:bx6h2r9) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L23)):*
>IIUC this like does not actually insert the new tx into the map. Rather we assume that the `tx_hash` indicates the contents are the same and we we clone the contents of the old entry.

Yes I was assuming that tx_hash uniquely describes the L1TransactionHandler inside, since it is computed by having it's fields. In other words, unless someone adds a bug somewhere, the given `tx` should be identical to the stored one.

Added an assert to verify this.

>We intentionally ignore the TxState of the existing entry?

Ya it shouldn't matter for insertion. In fact, this is one of those cases that shouldn't happen anyway, since our scraper never backtracks and there are no dups on L1, but better be safe and check it anyway.
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 32 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA0sPmDPWmSSVvjnbX:-OGAIMWO1c7FHgpw0GJf:b-kroo92) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L32)):*

Impl looks to me like we remove the first available element, not the last
</blockquote></details>

Arrrrg i also get those mixed up, thanks!
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 65 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA7ywo9q7MQMTi7VVd:-OGAI_J2B4peWw5q0jP8:bnw1ub5) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L65)):*
>This should not be in the public doc, switch to `//`

Right, fixed.

>Let's simplify this and say "Operation is linear time with both the number of known transactions and the number of committed transactions"

Done (if I understand correctly which part you wanted changed?)
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 82 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA2SxbC0dboUhTCqtl:-OGAJ46S1KpS-0xcWKSb:barpu3g) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L82)):*
> Why are pub functions underscored? That's a python thing I thought.

Not related to the python privacy thing, even if it's pub it's still subject to the `unused` lint because the module is `pub(crate)` and not `pub`.
The `_`is to quiet it down without haveing to add a ignore lint thing

>I think of "rollback" being for a commit, not staging. I prefer either unstage or reset\_staging, or rollback\_staging.

Changed to `rollback_staging`. WDYM by rollback being for a commit? I picked rollback since this staging behavior seems almost exactly like db transactions, where `rollback` behaves like reset_staging, [ as the opposite of commit](https://en.wikipedia.org/wiki/Commit_(data_management)), in which all operations in the session are undo'ed (only if commit has been called previously)
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 98 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA96ouA83MhtrAgm9s:-OGAJZbvCjjhkl9R2B0d:b-br7ohh) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L98)):*

We don't also want a `new()` fn?
</blockquote></details>

Sure, added (also with `_` because not used yet).
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 124 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA1E5R6OpNbeQAZpPh:-OGAK_VP9pdel-IYdS1G:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L124)):*

What about removing this and unstage and having a function `set_state(TxState)`?
</blockquote></details>

Done
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 127 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9wCxp-C8H6sw3LZOj:-OGAN4FK2dqQ9-CeGpkR:b3q7uor) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L127)):*

Not blocking, I just like to avoid macros when easy to.
</blockquote></details>

death to macros
___
*[`crates/starknet_l1_provider/src/test_utils.rs` line 86 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA9dtKASoRtaArIB-n:-OGANVSZ4Bc9mk1oZ2jl:bjh5zy7) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/test_utils.rs#L86)):*

TryInto?
</blockquote></details>

Done (made it an Into since it can't fail)
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 21 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9uqml7ypptyDQh-oK:-OGAGy_RDcXsp9DojXdE:btltxct) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L21)):*
> ```XML
> impl SoftDeleteIndexMap {
>     /// Inserts a transaction into the map, returning the previous transaction if it existed.
>     pub fn _insert(&mut self, tx: L1HandlerTransaction) -> Option<L1HandlerTransaction> {
> ```

Not used in this PR yet, will be used in subsequent commits I have in store, wanted to shave things off of this PR as much as possible.

I could have added a `ignore(unused)` in the file, but those tend to get forgotten, and the `_` prefix is impossible to miss.
I can remove and add the ignore if you prefer though.

matan-starkware (changes requested):
> Reviewed 9 of 9 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3208)* status: all files reviewed, 5 unresolved discussions (waiting on @giladchase)

___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 21 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9uqml7ypptyDQh-oK:-OGYnIKYBud9xcoZSLhk:b-4if4y2) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L21)):*

Not used in this PR yet, will be used in subsequent commits I have in store, wanted to shave things off of this PR as much as possible.

I could have added a `ignore(unused)` in the file, but those tend to get forgotten, and the `_` prefix is impossible to miss.
I can remove and add the ignore if you prefer though.
</blockquote></details>

Ok I missed that fact, yeah underscore is nicer. 

Because this is already under review let's just move forward as is, but WDYT of in the future:
1. You write the whole thing; full struct + tests + usage (I'm assuming you already did that)
2. Sit with me to explain it - so I get the full context
3. Break into PRs to avoid these unused function
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 23 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9vFRy1CUvQJe11NYt:-OGYo8yQCDKv93Lg1BMp:b2nhlkv) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L23)):*

>IIUC this like does not actually insert the new tx into the map. Rather we assume that the `tx_hash` indicates the contents are the same and we we clone the contents of the old entry.

Yes I was assuming that tx_hash uniquely describes the L1TransactionHandler inside, since it is computed by having it's fields. In other words, unless someone adds a bug somewhere, the given `tx` should be identical to the stored one.

Added an assert to verify this.

>We intentionally ignore the TxState of the existing entry?

Ya it shouldn't matter for insertion. In fact, this is one of those cases that shouldn't happen anyway, since our scraper never backtracks and there are no dups on L1, but better be safe and check it anyway.
</blockquote></details>

Ok 
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 82 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGA2SxbC0dboUhTCqtl:-OGYr5cVDKM59pLHEdWm:b3srq7y) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L82)):*

> Why are pub functions underscored? That's a python thing I thought.

Not related to the python privacy thing, even if it's pub it's still subject to the `unused` lint because the module is `pub(crate)` and not `pub`.
The `_`is to quiet it down without haveing to add a ignore lint thing

>I think of "rollback" being for a commit, not staging. I prefer either unstage or reset\_staging, or rollback\_staging.

Changed to `rollback_staging`. WDYM by rollback being for a commit? I picked rollback since this staging behavior seems almost exactly like db transactions, where `rollback` behaves like reset_staging, [ as the opposite of commit](https://en.wikipedia.org/wiki/Commit_(data_management)), in which all operations in the session are undo'ed (only if commit has been called previously)
</blockquote></details>

In my mind "rollback a commit" is a phrase so I like mentioning staging explicitly.
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 127 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9wCxp-C8H6sw3LZOj:-OGYyTjsB-LiKwZPRSMB:b-7qtjwr) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L127)):*

death to macros
</blockquote></details>

Hehe

I will admit that I love the `derive` macro
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 28 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGYoMid89KXG9S1UyPR:-OGYoMid89KXG9S1UyPS:b5qr7dl) ([raw file](https://github.com/starkware-libs/sequencer/blob/4d2d043bbeca4f3b887c24cdcc84bd20a7809507/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L28)):*
> ```XML
>         match self.txs.entry(tx.tx_hash) {
>             Entry::Occupied(entry) => {
>                 debug_assert_eq!(entry.get().transaction, tx);
> ```

I'd hold off on making this debug until:
1. We have spent some time with the code and are confident there is no such bug
2. We are worried about optimizing this path



_Suggestion:_
```XML
                assert_eq!(entry.get().transaction, tx);
```
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 29 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGZ-Ijw5IcsAlXpKV6E:-OGZ-Ijw5IcsAlXpKV6F:b-v76jfy) ([raw file](https://github.com/starkware-libs/sequencer/blob/4d2d043bbeca4f3b887c24cdcc84bd20a7809507/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L29)):*
> ```XML
>             Entry::Occupied(entry) => {
>                 debug_assert_eq!(entry.get().transaction, tx);
>                 Some(entry.get().transaction.clone())
> ```

avoid clone?



_Suggestion:_
```XML
                Some(tx)
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3208)* status: 4 of 9 files reviewed, 2 unresolved discussions (waiting on @matan-starkware)

___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 21 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9uqml7ypptyDQh-oK:-OGZ11lAAY9Hgc1B6Laa:b-4oj1qs) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L21)):*

Ok I missed that fact, yeah underscore is nicer. 

Because this is already under review let's just move forward as is, but WDYT of in the future:
1. You write the whole thing; full struct + tests + usage (I'm assuming you already did that)
2. Sit with me to explain it - so I get the full context
3. Break into PRs to avoid these unused function
</blockquote></details>

Sure.

Though in order to completely it'll mean PRs will get somewhat large, because whether I go bottom-up or top-down, you'll always hit unused issues unless the whole flow is added at once.

Another option which might make things simpler is to make everything `pub` while it's WIP, and after everything's merged we can do a really quick one-pass and add pub(crate)/remove-pub on everything, once all the usages are in.
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 23 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9vFRy1CUvQJe11NYt:-OGZ29qC3B41O_3sjDsb:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L23)):*

Ok
</blockquote></details>

Done.
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 127 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9wCxp-C8H6sw3LZOj:-OGZ06ig2FH1Wa9Er2__:b-malo4z) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L127)):*

Hehe

I will admit that I love the `derive` macro
</blockquote></details>

Note that for `matches!` we have to fight clippy a bit, cause if I do something like:
```rust
match foo_out_of_many {
  Foo => true
  _ => false
}
```
clippy will tell me to use matches!.
You'll see it in a later PR when i match on ProviderState, I'll have to do:
```rust
match state {
   ProviderState::Uninitialized => true
   ProviderState::Pending | ProviderState::Validate | ProviderState::Proposer => false 
}
```
which is getting a bit noisy

Also ya derive's cool (except for the part about forcing derives on generic variables boooo)
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 29 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGZ-Ijw5IcsAlXpKV6E:-OGZN8QX5ik0uwX3pxij:b-5o3tz0) ([raw file](https://github.com/starkware-libs/sequencer/blob/4d2d043bbeca4f3b887c24cdcc84bd20a7809507/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L29)):*

avoid clone?
</blockquote></details>

Done (had to do an extra hash query to please the borrow checker)

matan-starkware (approved):
> Reviewed 5 of 5 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3208)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 21 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9uqml7ypptyDQh-oK:-OGZlgd8-G3mrW9HEX5e:blh2iyl) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L21)):*

Sure.

Though in order to completely it'll mean PRs will get somewhat large, because whether I go bottom-up or top-down, you'll always hit unused issues unless the whole flow is added at once.

Another option which might make things simpler is to make everything `pub` while it's WIP, and after everything's merged we can do a really quick one-pass and add pub(crate)/remove-pub on everything, once all the usages are in.
</blockquote></details>

Right, we don't want the entire project to be one big PR and block you sending. I was thinking more like this struct could be 1 big PR that you get right before breaking into many PRs for me to review. That scale make sense?
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 127 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9wCxp-C8H6sw3LZOj:-OGZm1uH9s5hmaQT6AI6:b-3nbs78) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L127)):*

Note that for `matches!` we have to fight clippy a bit, cause if I do something like:
```rust
match foo_out_of_many {
  Foo => true
  _ => false
}
```
clippy will tell me to use matches!.
You'll see it in a later PR when i match on ProviderState, I'll have to do:
```rust
match state {
   ProviderState::Uninitialized => true
   ProviderState::Pending | ProviderState::Validate | ProviderState::Proposer => false 
}
```
which is getting a bit noisy

Also ya derive's cool (except for the part about forcing derives on generic variables boooo)
</blockquote></details>

Ok I am open to that latter one using `matches!` since that's verbose, but can we just do:

```
if state == ProviderState::Uninitialized {
    true
} else {
    false
}
```

I thought that when you match only 1 variant clippy recommends using an `if` not `matches!`
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 29 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGZ-Ijw5IcsAlXpKV6E:-OGZnRibCJyQRqztF9Us:b-jmedxp) ([raw file](https://github.com/starkware-libs/sequencer/blob/4d2d043bbeca4f3b887c24cdcc84bd20a7809507/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L29)):*

Done (had to do an extra hash query to please the borrow checker)
</blockquote></details>

If we assert that the new entry has the same set of transactions as the old entry, why do we bother returning it? The std HashMap returns the old entry because it may have different contents than the new one. Should we just return a bool instead to specify if this entry already exists?

If you are really set on returning the old entry how about this?
```
    pub fn _insert(&mut self, tx: L1HandlerTransaction) -> Option<L1HandlerTransaction> {
        match self.txs.entry(tx.tx_hash) {
            Entry::Occupied(entry) => {
                assert_eq!(entry.get().transaction, tx);
                Some(tx)
            }
            Entry::Vacant(entry) => {
                entry.insert(TransactionEntry::new(tx));
                None
            }
        }
    }
```

Returning an owned value is much nice. Since we just checked equality we may as well just return the "new" one cuz it's easy. If you want to be pedantic then:

```
    pub fn _insert(&mut self, tx: L1HandlerTransaction) -> Option<L1HandlerTransaction> {
        match self.txs.entry(tx.tx_hash) {
            Entry::Occupied(mut entry) => {
                assert_eq!(entry.get().transaction, tx);
                Some(std::mem::replace(&mut entry.get_mut().transaction, tx))
            }
            Entry::Vacant(entry) => {
                entry.insert(TransactionEntry::new(tx));
                None
            }
        }
    }
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3208)* status: 8 of 10 files reviewed, 1 unresolved discussion (waiting on @matan-starkware)

___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 21 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9uqml7ypptyDQh-oK:-OGaDGld5AsFCRm2cAsd:b-sobdj5) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L21)):*

Right, we don't want the entire project to be one big PR and block you sending. I was thinking more like this struct could be 1 big PR that you get right before breaking into many PRs for me to review. That scale make sense?
</blockquote></details>

Sure, i think this will be seamless: I'll open a PR with everything split into commits, once you TAL I can close the PR then SPR it, and immediately the commits will be converted to a stack.
Or alternatively we can just use per-commit view on reviewable and review each commit in order.
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 127 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OG9wCxp-C8H6sw3LZOj:-OGaDjN_3Kw1cVz_u9Jh:b-p677lt) ([raw file](https://github.com/starkware-libs/sequencer/blob/b839e92ecb18f788a276417a39e4d39d1a12e2fd/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L127)):*

Ok I am open to that latter one using `matches!` since that's verbose, but can we just do:

```
if state == ProviderState::Uninitialized {
    true
} else {
    false
}
```

I thought that when you match only 1 variant clippy recommends using an `if` not `matches!`
</blockquote></details>

Got it. RE 1 variant, i think that's only for `if let` which will get recommended if the `_` branch is `()`. But if the result is a bool then it recommends matches.
___
*[`crates/starknet_l1_provider/src/soft_delete_index_map.rs` line 29 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/3208#-OGZ-Ijw5IcsAlXpKV6E:-OGcbl-Z8ke1rzcWalgI:bro5g2q) ([raw file](https://github.com/starkware-libs/sequencer/blob/4d2d043bbeca4f3b887c24cdcc84bd20a7809507/crates/starknet_l1_provider/src/soft_delete_index_map.rs#L29)):*

If we assert that the new entry has the same set of transactions as the old entry, why do we bother returning it? The std HashMap returns the old entry because it may have different contents than the new one. Should we just return a bool instead to specify if this entry already exists?

If you are really set on returning the old entry how about this?
```
    pub fn _insert(&mut self, tx: L1HandlerTransaction) -> Option<L1HandlerTransaction> {
        match self.txs.entry(tx.tx_hash) {
            Entry::Occupied(entry) => {
                assert_eq!(entry.get().transaction, tx);
                Some(tx)
            }
            Entry::Vacant(entry) => {
                entry.insert(TransactionEntry::new(tx));
                None
            }
        }
    }
```

Returning an owned value is much nice. Since we just checked equality we may as well just return the "new" one cuz it's easy. If you want to be pedantic then:

```
    pub fn _insert(&mut self, tx: L1HandlerTransaction) -> Option<L1HandlerTransaction> {
        match self.txs.entry(tx.tx_hash) {
            Entry::Occupied(mut entry) => {
                assert_eq!(entry.get().transaction, tx);
                Some(std::mem::replace(&mut entry.get_mut().transaction, tx))
            }
            Entry::Vacant(entry) => {
                entry.insert(TransactionEntry::new(tx));
                None
            }
        }
    }
```
</blockquote></details>

Good point, bool is fine, added.

PR #3329, opened 2025-01-15
fix(starknet_l1_provider): provider height should match valid API height

giladchase:


matan-starkware (changes requested):
> Reviewed 1 of 7 files at r1, 1 of 4 files at r3, 6 of 7 files at r6, 2 of 7 files at r7, 5 of 5 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3329)* status: all files reviewed, 16 unresolved discussions (waiting on @giladchase)

___
*[`crates/starknet_l1_provider/src/bootstrapper.rs` line 8 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxZozn5Y_cWPWhYY_8:-OGxZozn5Y_cWPWhYY_9:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/bootstrapper.rs#L8)):*
> ```XML
>     pub catch_up_height: BlockNumber,
>     pub commit_block_backlog: Vec<CommitBlockBacklog>,
> }
> ```





_Suggestion:_
```XML
/// Cache's commits to be applied later. This flow is only relevant while the node is starting up.
#[derive(Clone, Debug, Default, Eq, PartialEq)]
pub struct Bootstrapper {
    pub catch_up_height: BlockNumber,
    pub commit_block_backlog: Vec<CommitBlockBacklog>,
}
```
___
*[`crates/starknet_l1_provider/src/bootstrapper.rs` line 11 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxZj2v9zE7knpsNbZq:-OGxZj2v9zE7knpsNbZr:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/bootstrapper.rs#L11)):*
> ```XML
> 
> impl Bootstrapper {
>     pub fn is_caught_up(&self, current_provider_height: BlockNumber) -> bool {
> ```





_Suggestion:_
```XML
    /// Check if the caller has caught up with the bootstrapper.
    pub fn is_caught_up(&self, current_provider_height: BlockNumber) -> bool {
```
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 27 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxeScaABgW6RzzJ9PZ:-OGxeScb3_p07BCZoPqx:b-8tax) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L27)):*
> ```XML
>         )
>     }};
> }
> ```

There were a couple imports but I was able to replace all calls to the macro with this



_Suggestion:_
```XML
fn l1_handler(tx_hash: usize) -> L1HandlerTransaction {
    let tx_hash = TransactionHash(StarkHash::from(tx_hash));
    executable_l1_handler_tx(L1HandlerTxArgs { tx_hash, ..Default::default() })
}
```
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 41 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxgHgUEKlaSV5yLT3E:-OGxgHgUEKlaSV5yLT3F:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L41)):*
> ```XML
>         }
>     }};
> }
> ```





_Suggestion:_
```XML
fn make_bootstrapper(catch_up_height: u64, backlog: &[(u64, &[usize])]) -> Bootstrapper {
    let commit_block_backlog = backlog
        .into_iter()
        .map(|(height, txs)| CommitBlockBacklog {
            height: BlockNumber(*height),
            committed_txs: txs
                .into_iter()
                .map(|hash| TransactionHash(StarkHash::from(*hash)))
                .collect(),
        })
        .collect();
    Bootstrapper { catch_up_height: BlockNumber(catch_up_height), commit_block_backlog }
}
```
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 123 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxj-AeDJQupumTQg4y:-OGxj-AeDJQupumTQg4z:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L123)):*
> ```XML
> 
> #[test]
> fn test_commit_block_empty_block() {
> ```





_Suggestion:_
```XML
fn commit_empty_block() {
```
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 143 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxjpl-BcumDgUUqu0k:-OGxjpl-BcumDgUUqu0l:bwpzw6y) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L143)):*
> ```XML
> 
> #[test]
> fn test_commit_block_during_proposal() {
> ```

Didn't the previous test do this? Looks like we're committing across many different states.



_Code quote:_
```XML
fn test_commit_block_during_proposal() {
```
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 177 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxrYZv48zI3qZNkcdM:-OGxrYZv48zI3qZNkcdN:b-7zi8ni) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L177)):*
> ```XML
>         .with_state(ProviderState::Pending)
>         .build();
>     expected_l1_provider.assert_eq(&l1_provider);
> ```

WDYT?



_Suggestion:_
```XML
    let mut txs = vec![l1_handler(1), l1_handler(2), l1_handler(3)];
    // Setup.
    let mut l1_provider = L1ProviderContentBuilder::new()
        .with_txs(txs.clone())
        .with_height(BlockNumber(10))
        .with_state(ProviderState::Propose)
        .build_into_l1_provider();

    // Commit block during proposal.
    for (i, state) in [ProviderState::Propose, ProviderState::Pending, ProviderState::Validate]
        .into_iter()
        .enumerate()
    {
        l1_provider.state = state;
        let height = BlockNumber(10 + i as u64);
        l1_provider.commit_block(&[txs.pop().unwrap().tx_hash], height).unwrap();
        let expected_l1_provider = L1ProviderContentBuilder::new()
            .with_txs(txs.clone())
            .with_height(height)
            .with_state(ProviderState::Pending)
            .build();
        expected_l1_provider.assert_eq(&l1_provider);
    }
```
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 186 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxgC2KBvhGRQsidXvc:-OGxgC2KBvhGRQsidXvd:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L186)):*
> ```XML
>         backlog: [10 => [2], 11 => [4]],
>         catch_up: 10
>     ));
> ```





_Suggestion:_
```XML
    let initial_bootstrap_state =
        ProviderState::Bootstrap(make_bootstrapper(10, &[(10, &[2]), (11, &[4])]));
```
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 195 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxrRw-6hziLTsZ9zAG:-OGxrRw-6hziLTsZ9zAH:b-plnhr5) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L195)):*
> ```XML
>     // Test.
>     // Commit height too low to affect backlog.
>     l1_provider.commit_block(&[tx_hash!(1)], BlockNumber(8)).unwrap();
> ```

I can live with this macro but there are 2 options I'd like you to consider:
1. `TransactionHash(StarkHash::from(x))`
2. add a local `fn tx_hash(usize) -> TransactionHash` which calls (1) if you really want the short line in the test



_Code quote:_
```XML
tx_hash!
```
___
*[`crates/starknet_l1_provider/src/l1_provider.rs` line 66 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGeAsHeAJbT0BDaLHTJ:-OGeAsHeAJbT0BDaLHTK:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/f89f4904a86979d51f7d73840d882a9faac8bc1c/crates/starknet_l1_provider/src/l1_provider.rs#L66)):*
> ```XML
>             ProviderState::Validate => Ok(self.tx_manager.validate_tx(tx_hash)),
>             ProviderState::Propose => Err(L1ProviderError::ValidateTransactionConsensusBug),
>             ProviderState::Pending | ProviderState::Bootstrap { .. } => {
> ```





_Suggestion:_
```XML
            ProviderState::Pending | ProviderState::Bootstrap(_) => {
```
___
*[`crates/starknet_l1_provider_types/src/errors.rs` line 15 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxsko2D9G5gzGox6Ki:-OGxsko2D9G5gzGox6Kj:bvjgs7o) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider_types/src/errors.rs#L15)):*
> ```XML
>         "`get_txs` called when provider is not in proposer state, likely due to a crash; restart \
>          block proposal"
>     )]
> ```

"likely due to a crash; restart \
         block proposal"
" 

move this to a comment I think. Samve for below



_Suggestion:_
```XML
    #[error(
        "`get_txs` called when provider is not in proposer state"     )]
```
___
*[`crates/starknet_l1_provider_types/src/errors.rs` line 26 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxsz9T4QKf0JBrlJa8:-OGxsz9T4QKf0JBrlJa9:b3kre) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider_types/src/errors.rs#L26)):*
> ```XML
>     #[error("Cannot transition from {from} to {to}")]
>     UnexpectedProviderStateTransition { from: String, to: String },
>     #[error("`validate` called while in `Propose`")]
> ```

?



_Suggestion:_
```XML
    #[error("`validate` called while in `Propose` state")]
```
___
*[`crates/starknet_l1_provider/src/test_utils.rs` line 119 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxsJu7585UYwHYjlUp:-OGxsJu7585UYwHYjlUq:bg2qtej) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/test_utils.rs#L119)):*
> ```XML
>         }
>     }
> }
> ```

WDYT instead of implementing `PartialEq` and then calling the assert inline?



_Code quote:_
```XML
impl TransactionManagerContent {
    #[track_caller]
    fn assert_eq(&self, tx_manager: &TransactionManager) {
        if let Some(txs) = &self.txs {
            assert_eq!(
                txs,
                &tx_manager.txs.txs.values().map(|tx| tx.transaction.clone()).collect_vec()
            );
        }
    }
}
```
___
*[`crates/starknet_l1_provider/src/l1_provider.rs` line 97 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxS9E28K5QsK5802V9:-OGxS9E3-wJmhC-BAmcn:bakfjn1) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider.rs#L97)):*
> ```XML
>     fn bootstrap(&mut self, committed_txs: &[TransactionHash], height: BlockNumber) {
>         match height.cmp(&self.current_height) {
>             Less => unreachable!("We should never get `commit_block`s for old heights."),
> ```

Can't this be reached by the client sending an old commit? If so I think returning an error would express the invariant best.

If you really want to abort I recommend panic. I prefer using `unreachable` for something where the invariants within the module itself make the branch unreachable, not some expectation about user behavior.



_Code quote:_
```XML
            Less => unreachable!("We should never get `commit_block`s for old heights."),
```
___
*[`crates/starknet_l1_provider/src/l1_provider.rs` line 113 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxUs9d86VfZCtRcn40:-OGxUs9d86VfZCtRcn41:b-ortghc) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider.rs#L113)):*
> ```XML
>             .state
>             .get_bootstrapper()
>             .expect("This method should only be called when bootstrapping.");
> ```

Move to the top of the fn since we call this in all (reachable) branches?



_Code quote:_
```XML
        let bootstrapper = self
            .state
            .get_bootstrapper()
            .expect("This method should only be called when bootstrapping.");
```
___
*[`crates/starknet_l1_provider/src/l1_provider.rs` line 125 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxa4ac1TNB14s-w9Ww:-OGxa4ac1TNB14s-w9Wx:b-66fljn) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider.rs#L125)):*
> ```XML
>                             .all(|height| height[1].height == height[0].height.unchecked_next()),
>                 "Backlog {backlog:?} must have sequential heights starting sequentially after \
>                  catch_up_height: {}",
> ```

Backlog is `[[TX]]`. I think it'd probably be better just to print the heights. Regardless, I think list should be at the end of the error message, otherwise it will be hard to read.



_Code quote:_
```XML
                "Backlog {backlog:?} must have sequential heights starting sequentially after \
                 catch_up_height: {}",
```

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3329)* status: all files reviewed, 12 unresolved discussions (waiting on @matan-starkware)

___
*[`crates/starknet_l1_provider/src/l1_provider.rs` line 97 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxS9E28K5QsK5802V9:-OH1VLDtAunVCK74F5ec:btyxbqv) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider.rs#L97)):*
>Can't this be reached by the client sending an old commit

Nope, the sync task is initialized from the provider's current height, and the provider can only advance through commit_blocks from it. If we get an old height during bootstrapping then it's a sync task bug.

Changed to panic, but it can probably also be a tracing::error!("...") and skip, because eventually it'll get to the correct height, but this can mask bugs like trying to sync up from genesis by mistake (if it isn't by mistake then cur_height would be 0).
___
*[`crates/starknet_l1_provider/src/l1_provider.rs` line 113 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxUs9d86VfZCtRcn40:-OH1W0tOD13ONKwyTLDX:b-b5njna) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider.rs#L113)):*

Move to the top of the fn since we call this in all (reachable) branches?
</blockquote></details>

Ya this bugged me too, but i don't think that's possible due to the borrow-checker: `get_bootstrapper` mut borrows `self`, but also `apply_commit_block`, so the latter isn't allowed because the `bootstrapper` one is still live.
___
*[`crates/starknet_l1_provider/src/l1_provider.rs` line 125 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxa4ac1TNB14s-w9Ww:-OH1XDH-DIeQFXUo6QYA:b-luq4na) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider.rs#L125)):*

Backlog is `[[TX]]`. I think it'd probably be better just to print the heights. Regardless, I think list should be at the end of the error message, otherwise it will be hard to read.
</blockquote></details>

Oh good point!
Done
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 27 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxeScaABgW6RzzJ9PZ:-OH1Y_3B0sFTGPTmxoMi:b-ckg1kc) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L27)):*

There were a couple imports but I was able to replace all calls to the macro with this
</blockquote></details>

Yr right.
We aren't gonna need any more args other than tx_hash here anytime soon, so the macros aren't necessary and the kwargs isn't dispelling any kind of ambiguity (these were the use-cases I had in mind when I wrote it).
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 143 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxjpl-BcumDgUUqu0k:-OH1cGPJ4X2SyIw_V8H5:b5toshj) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L143)):*

Didn't the previous test do this? Looks like we're committing across many different states.
</blockquote></details>

Prev test was only about testing an empty block, state was arbitrary.

This one focused on the states, rather than the content which was arbitrary.
I split the latter test into smaller tests, one for each state, to make the distinction clearer.
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 177 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxrYZv48zI3qZNkcdM:-OH1kQD7DadrV-MWY7Op:bjb716z) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L177)):*

WDYT?
</blockquote></details>

This can work, do you prefer this over splitting into three unit tests?
The reason I split it was because the multi-state test was a bit forced, there wasn't any reason why this had to be a single test, it just makes the test logic unnecessarily complex.
I think if we want to test state transition we should do it properly with start-block calls, but that already sounds like a flow test.
WDYT?
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 195 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxrRw-6hziLTsZ9zAG:-OH1fIE2DRI3FY1l-kHo:bwvmfh7) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L195)):*

I can live with this macro but there are 2 options I'd like you to consider:
1. `TransactionHash(StarkHash::from(x))`
2. add a local `fn tx_hash(usize) -> TransactionHash` which calls (1) if you really want the short line in the test
</blockquote></details>

Added a function.

The reason why this is a macro is probably to handle different input types succinctly.
This function is limited to `usize`, which is fine ATM because I'm using literals, but won't be useful for other types like u8 without explicit casts.
This will probably be good enough in this test though, but i think we should be wary of duplicating SNAPI test utils too much, since that might imply some hidden meaning as to why they can't be used here.
___
*[`crates/starknet_l1_provider/src/test_utils.rs` line 119 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxsJu7585UYwHYjlUp:-OH1e9Iz-PCjZeefL_kq:b-n6e8hw) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/test_utils.rs#L119)):*

WDYT instead of implementing `PartialEq` and then calling the assert inline?
</blockquote></details>

Not sure i follow :thinking:  did you mean the PartialEq on `TransactionManagerContent`? It was just added for QOL of the test-utils users (I used it for debugging for example), but right now it isn't used directly in the code.
___
*[`crates/starknet_l1_provider_types/src/errors.rs` line 15 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxsko2D9G5gzGox6Ki:-OH1dJzZBntLs1QUrP3F:b-qk23rn) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider_types/src/errors.rs#L15)):*

"likely due to a crash; restart \
         block proposal"
" 

move this to a comment I think. Samve for below
</blockquote></details>

Done, though i think having it in the error message will make flow-test failures (or production ones :scream: ) easier to debug.
In particular, the user might not necessarily go to the error enum definition when they get this error, and might miss out on the debugging tip.
___
*[`crates/starknet_l1_provider/src/bootstrapper.rs` line 8 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxZozn5Y_cWPWhYY_8:-OH1UhD75tpbo6qhBpKr:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/bootstrapper.rs#L8)):*
> ```XML
>     pub catch_up_height: BlockNumber,
>     pub commit_block_backlog: Vec<CommitBlockBacklog>,
> }
> ```

Done.
___
*[`crates/starknet_l1_provider/src/bootstrapper.rs` line 11 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxZj2v9zE7knpsNbZq:-OH1Ujj-5T7N1_UrYwe8:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/bootstrapper.rs#L11)):*
> ```XML
> 
> impl Bootstrapper {
>     pub fn is_caught_up(&self, current_provider_height: BlockNumber) -> bool {
> ```

Done.
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 41 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxgHgUEKlaSV5yLT3E:-OH1ZqPg3yBepqeopkBR:bc2jszm) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L41)):*
> ```XML
>         }
>     }};
> }
> ```

This one might make the callsites less clear, in other words, readers (myself included) will most likely have to go and read this function in order to understand what the args mean:
```rust
/// callsite before
    let initial_bootstrap_state = ProviderState::Bootstrap(bootstrapper!(
        backlog: [10 => [2], 11 => [4]],
        catch_up: 10
    ));

/// callsite after
    let initial_bootstrap_state = ProviderState::Bootstrap(make_bootstrapper(
        10, 
        &[(10, [2]), (11, [4])],
    )); 
```

In other words, the macro here was used in order to improve the readability of the unit test, which I think should be as simple to understand as possible.

If I run into this function in the future, I won't trust my memory (if I remember at all) what those numbers mean, which means that this adds overhead to all readers of every bootstrapping test.

I wouldn't do this kind of thing in production, but for tests I think it's a very good usecase for a macro: they spare the user from having to read test-utils they probably don't care about when they have a failing test.
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 186 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxgC2KBvhGRQsidXvc:-OH1jFkz5VUrrTVdpwEU:b-5jmwah) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L186)):*
> ```XML
>         backlog: [10 => [2], 11 => [4]],
>         catch_up: 10
>     ));
> ```

Answered in make_bootstrapper thread

matan-starkware (approved):
> Reviewed 3 of 8 files at r9, 4 of 5 files at r10, 3 of 3 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3329)* status: all files reviewed, 2 unresolved discussions (waiting on @giladchase)

___
*[`crates/starknet_l1_provider/src/l1_provider.rs` line 97 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxS9E28K5QsK5802V9:-OH1nkzvDVz71x_SPIhz:b393nxh) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider.rs#L97)):*

>Can't this be reached by the client sending an old commit

Nope, the sync task is initialized from the provider's current height, and the provider can only advance through commit_blocks from it. If we get an old height during bootstrapping then it's a sync task bug.

Changed to panic, but it can probably also be a tracing::error!("...") and skip, because eventually it'll get to the correct height, but this can mask bugs like trying to sync up from genesis by mistake (if it isn't by mistake then cur_height would be 0).
</blockquote></details>

"it's a sync task bug" = "client sending an old commit" 
- I agree this is a bug, but it's a bug by the user no the L1 Provider. To me this feels like returning an error, just like if the batcher calls `validate` during the `Propose` state.
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 177 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxrYZv48zI3qZNkcdM:-OH1oo_07HpdgMs5YAkQ:b-7zvker) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L177)):*

This can work, do you prefer this over splitting into three unit tests?
The reason I split it was because the multi-state test was a bit forced, there wasn't any reason why this had to be a single test, it just makes the test logic unnecessarily complex.
I think if we want to test state transition we should do it properly with start-block calls, but that already sounds like a flow test.
WDYT?
</blockquote></details>

Your new way is better
___
*[`crates/starknet_l1_provider/src/l1_provider_tests.rs` line 195 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxrRw-6hziLTsZ9zAG:-OH1p6zi0jYj8ThAIrND:b-qnhz23) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider_tests.rs#L195)):*

Added a function.

The reason why this is a macro is probably to handle different input types succinctly.
This function is limited to `usize`, which is fine ATM because I'm using literals, but won't be useful for other types like u8 without explicit casts.
This will probably be good enough in this test though, but i think we should be wary of duplicating SNAPI test utils too much, since that might imply some hidden meaning as to why they can't be used here.
</blockquote></details>

Like I said, I can live with this one.
___
*[`crates/starknet_l1_provider/src/test_utils.rs` line 119 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxsJu7585UYwHYjlUp:-OH2Tcav39D13wkxy8T_:b-dzdr35) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/test_utils.rs#L119)):*

Not sure i follow :thinking:  did you mean the PartialEq on `TransactionManagerContent`? It was just added for QOL of the test-utils users (I used it for debugging for example), but right now it isn't used directly in the code.
</blockquote></details>

No I meant `PartialEq<TransactionManager> for TransactionManagerContent`, and this fn is the impl (returning a bool)

Then in tests you can write `assert_eq(manager, manager_content)`. instead of `manager_content.assert_eq(manager)`

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3329)* status: 0 of 13 files reviewed, 1 unresolved discussion (waiting on @matan-starkware)

___
*[`crates/starknet_l1_provider/src/l1_provider.rs` line 97 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxS9E28K5QsK5802V9:-OH2thYL2dE1_eI6GO8x:bs7wumb) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/l1_provider.rs#L97)):*

"it's a sync task bug" = "client sending an old commit" 
- I agree this is a bug, but it's a bug by the user no the L1 Provider. To me this feels like returning an error, just like if the batcher calls `validate` during the `Propose` state.
</blockquote></details>

Done.
___
*[`crates/starknet_l1_provider/src/test_utils.rs` line 119 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/3329#-OGxsJu7585UYwHYjlUp:-OH2vSC03wr-WGBCWEz9:b4e7w2i) ([raw file](https://github.com/starkware-libs/sequencer/blob/68820e1d9bf19197074d40da42e073cbbb752692/crates/starknet_l1_provider/src/test_utils.rs#L119)):*

No I meant `PartialEq<TransactionManager> for TransactionManagerContent`, and this fn is the impl (returning a bool)

Then in tests you can write `assert_eq(manager, manager_content)`. instead of `manager_content.assert_eq(manager)`
</blockquote></details>

IMO I tend to avoid creating operators between different types; they tend to be suprising. 

For example, here, the comparison ignores the `committed` field since it is currently just an implementation detail (it doesn't really affect the API, just internal consistency). 

The only reason it's there is for when one wants to initialize a real `TransactionManager` in a test, at a specific state.

(Now that I think about it, maybe the current method name `assert_eq` is surprising for the same reason :thinking: )

matan-starkware (approved):
> Reviewed 2 of 13 files at r12, 6 of 7 files at r13, 4 of 7 files at r14, 3 of 3 files at r15, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/3329)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

PR #3486, opened 2025-01-19
chore(ci): cache anvil installation

adi-yakovian-starkware:


adi-yakovian-starkware:
Not the right anvil package...

PR #4259, opened 2025-02-19
chore(starknet_l1_gas_price): add gas price configurations to node config

guy-starkware:

PR #4202, opened 2025-02-17
feat(starknet_l1_gas_price): serialize the config for gas price scraper and provider

guy-starkware:
Allow the gas price provider and scraper's configs to be serialized.

PR #4139, opened 2025-02-13
feat(papyrus_base_layer): automock the base layer

guy-starkware:

PR #4124, opened 2025-02-13
chore(starknet_l1_gas_price): add new crate starknet_l1_gas_price_types

guy-starkware:

PR #4265, opened 2025-02-19
chore(starknet_l1_gas_price_types): add provider trait to types

guy-starkware:
Moves the L1GasPriceProviderClient trait to the new "types" crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -11302,7 +11302,6 @@ dependencies = [
 name = "starknet_l1_gas_price"
 version = "0.0.0"
 dependencies = [
- "mockall",
  "papyrus_base_layer",
  "papyrus_config",
  "serde",
@@ -11319,6 +11318,12 @@ dependencies = [
 name = "starknet_l1_gas_price_types"
 version = "0.0.0"
 dependencies = [
+ "async-trait",
+ "mockall",
+ "papyrus_base_layer",
+ "serde",
+ "starknet_api",
+ "starknet_sequencer_infra",
  "thiserror 1.0.69",
 ]
 
--- a/crates/starknet_l1_gas_price/Cargo.toml
+++ b/crates/starknet_l1_gas_price/Cargo.toml
@@ -6,7 +6,6 @@ repository.workspace = true
 license.workspace = true
 
 [dependencies]
-mockall = { workspace = true, optional = true }
 papyrus_base_layer.workspace = true
 papyrus_config.workspace = true
 serde.workspace = true
@@ -19,11 +18,8 @@ tracing.workspace = true
 validator.workspace = true
 
 [dev-dependencies]
-mockall.workspace = true
 papyrus_base_layer = { workspace = true, features = ["testing"] }
+starknet_l1_gas_price_types = { workspace = true, features = ["testing"] }
 
 [lints]
 workspace = true
-
-[features]
-testing = ["mockall"]
--- a/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs
+++ b/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs
@@ -1,32 +1,15 @@
 use std::collections::VecDeque;
 
-#[cfg(any(feature = "testing", test))]
-use mockall::automock;
-use papyrus_base_layer::PriceSample;
+use papyrus_base_layer::{L1BlockNumber, PriceSample};
 use serde::{Deserialize, Serialize};
-use starknet_api::block::{BlockNumber, BlockTimestamp};
+use starknet_api::block::BlockTimestamp;
 use starknet_l1_gas_price_types::errors::L1GasPriceProviderError;
+use starknet_l1_gas_price_types::{L1GasPriceProviderResult, PriceInfo};
 use validator::Validate;
 
 #[cfg(test)]
 #[path = "l1_gas_price_provider_test.rs"]
 pub mod l1_gas_price_provider_test;
-
-// TODO(guyn, Gilad): consider moving this to starknet_l1_provider_types/lib.rs?
-// This is an interface that allows sharing the provider with the scraper across threads.
-#[cfg_attr(any(feature = "testing", test), automock)]
-pub trait L1GasPriceProviderClient: Send + Sync {
-    fn add_price_info(
-        &self,
-        height: BlockNumber,
-        sample: PriceSample,
-    ) -> Result<(), L1GasPriceProviderError>;
-
-    fn get_price_info(
-        &self,
-        timestamp: BlockTimestamp,
-    ) -> Result<(u128, u128), L1GasPriceProviderError>;
-}
 
 #[derive(Clone, Debug, Serialize, Deserialize, Validate, PartialEq)]
 pub struct L1GasPriceProviderConfig {
@@ -48,73 +31,80 @@ impl Default for L1GasPriceProviderConfig {
 }
 
 #[derive(Clone, Debug)]
+pub struct GasPriceData {
+    pub height: L1BlockNumber,
+    pub sample: PriceSample,
+}
+
+#[derive(Clone, Debug)]
 pub struct L1GasPriceProvider {
     config: L1GasPriceProviderConfig,
-    data: VecDeque<(BlockNumber, PriceSample)>,
+    price_samples_by_block: VecDeque<GasPriceData>,
 }
 
 impl L1GasPriceProvider {
     pub fn new(config: L1GasPriceProviderConfig) -> Self {
-        Self { config, data: VecDeque::new() }
+        let storage_limit = config.storage_limit;
+        Self { config, price_samples_by_block: VecDeque::with_capacity(storage_limit) }
     }
 
     pub fn add_price_info(
         &mut self,
-        height: BlockNumber,
+        height: L1BlockNumber,
         sample: PriceSample,
-    ) -> Result<(), L1GasPriceProviderError> {
-        let last_plus_one = self.data.back().map(|(h, _)| h.0 + 1).unwrap_or(0);
-        if height.0 != last_plus_one {
-            return Err(L1GasPriceProviderError::InvalidHeight(format!(
-                "Block height is not consecutive: expected {}, got {}",
-                last_plus_one, height.0
-            )));
+    ) -> L1GasPriceProviderResult<()> {
+        let last_plus_one =
+            self.price_samples_by_block.back().map(|data| data.height + 1).unwrap_or(0);
+        if height != last_plus_one {
+            return Err(L1GasPriceProviderError::UnexpectedHeightError {
+                expected: last_plus_one,
+                found: height,
+            });
         }
-        self.data.push_back((height, sample));
-        if self.data.len() > self.config.storage_limit {
-            self.data.pop_front();
+        self.price_samples_by_block.push_back(GasPriceData { height, sample });
+        if self.price_samples_by_block.len() > self.config.storage_limit {
+            self.price_samples_by_block.pop_front();
         }
         Ok(())
     }
 
-    pub fn get_price_info(
-        &self,
-        timestamp: BlockTimestamp,
-    ) -> Result<(u128, u128), L1GasPriceProviderError> {
-        let mut gas_price = 0;
-        let mut data_gas_price = 0;
-
+    pub fn get_price_info(&self, timestamp: BlockTimestamp) -> L1GasPriceProviderResult<PriceInfo> {
         // This index is for the last block in the mean (inclusive).
-        let index_last_timestamp_rev = self.data.iter().rev().position(|(_, sample)| {
-            sample.timestamp <= timestamp.0 - self.config.lag_margin_seconds
-        });
+        let index_last_timestamp_rev =
+            self.price_samples_by_block.iter().rev().position(|data| {
+                data.sample.timestamp <= timestamp.0 - self.config.lag_margin_seconds
+            });
 
         // Could not find a block with the requested timestamp and lag.
         let Some(last_index_rev) = index_last_timestamp_rev else {
-            return Err(L1GasPriceProviderError::MissingData(format!(
-                "No block price data from time {} - {}s",
-                timestamp.0, self.config.lag_margin_seconds
-            )));
+            return Err(L1GasPriceProviderError::MissingDataError {
+                timestamp: timestamp.0,
+                lag: self.config.lag_margin_seconds,
+            });
         };
         // We need to convert the index to the forward direction.
-        let last_index = self.data.len() - last_index_rev;
+        let last_index = self.price_samples_by_block.len() - last_index_rev;
 
         let num_blocks = usize::try_from(self.config.number_of_blocks_for_mean)
             .expect("number_of_blocks_for_mean is too large to fit into a usize");
         if last_index < num_blocks {
-            return Err(L1GasPriceProviderError::MissingData(format!(
-                "Insufficient block price history: expected at least {}, found only {}",
-                num_blocks, last_index
-            )));
+            return Err(L1GasPriceProviderError::InsufficientHistoryError {
+                expected: num_blocks,
+                found: last_index,
+            });
         }
         // Go over all elements between last_index-num_blocks to last_index (non-inclusive).
-        for (_height, sample) in self.data.iter().skip(last_index - num_blocks).take(num_blocks) {
-            gas_price += sample.base_fee_per_gas;
-            data_gas_price += sample.blob_fee;
-        }
-        Ok((
-            gas_price / u128::from(self.config.number_of_blocks_for_mean),
-            data_gas_price / u128::from(self.config.number_of_blocks_for_mean),
-        ))
+        let (gas_price, data_gas_price) =
+            self.price_samples_by_block.iter().skip(last_index - num_blocks).take(num_blocks).fold(
+                (0, 0),
+                |(sum_base, sum_blob), data| {
+                    (sum_base + data.sample.base_fee_per_gas, sum_blob + data.sample.blob_fee)
+                },
+            );
+
+        Ok(PriceInfo {
+            base_fee_per_gas: gas_price / u128::from(self.config.number_of_blocks_for_mean),
+            blob_fee: data_gas_price / u128::from(self.config.number_of_blocks_for_mean),
+        })
     }
 }
--- a/crates/starknet_l1_gas_price/src/l1_gas_price_provider_test.rs
+++ b/crates/starknet_l1_gas_price/src/l1_gas_price_provider_test.rs
@@ -1,5 +1,6 @@
 use papyrus_base_layer::PriceSample;
-use starknet_api::block::{BlockNumber, BlockTimestamp};
+use starknet_api::block::BlockTimestamp;
+use starknet_l1_gas_price_types::PriceInfo;
 
 use crate::l1_gas_price_provider::{
     L1GasPriceProvider,
@@ -23,7 +24,7 @@ fn make_provider() -> (L1GasPriceProvider, Vec<PriceSample>) {
         let time = (i * 2).try_into().unwrap();
         let sample = PriceSample { timestamp: time, base_fee_per_gas: price, blob_fee: price + 1 };
         samples.push(sample.clone());
-        provider.add_price_info(BlockNumber(block_num), sample).unwrap();
+        provider.add_price_info(block_num, sample).unwrap();
     }
     (provider, samples)
 }
@@ -38,7 +39,7 @@ fn gas_price_provider_mean_prices() {
 
     // This calculation will grab config.number_of_blocks_for_mean samples from the middle of the
     // range.
-    let (gas_price, data_gas_price) =
+    let PriceInfo { base_fee_per_gas: gas_price, blob_fee: data_gas_price } =
         provider.get_price_info(BlockTimestamp(final_timestamp + lag)).unwrap();
 
     // The gas prices should go from block 1 to 3.
@@ -58,14 +59,14 @@ fn gas_price_provider_adding_samples() {
     // Timestamp for sample[3] is used to define the interval of samples 1 to 3.
     let final_timestamp = samples[3].timestamp;
 
-    let (gas_price, data_gas_price) =
+    let PriceInfo { base_fee_per_gas: gas_price, blob_fee: data_gas_price } =
         provider.get_price_info(BlockTimestamp(final_timestamp + lag)).unwrap();
 
     // Add a block to the provider.
     let sample = PriceSample { timestamp: 10, base_fee_per_gas: 10, blob_fee: 11 };
-    provider.add_price_info(BlockNumber(5), sample).unwrap();
+    provider.add_price_info(5, sample).unwrap();
 
-    let (gas_price_new, data_gas_price_new) =
+    let PriceInfo { base_fee_per_gas: gas_price_new, blob_fee: data_gas_price_new } =
         provider.get_price_info(BlockTimestamp(final_timestamp + lag)).unwrap();
 
     // This should not change the results if we ask for the same timestamp.
@@ -74,11 +75,11 @@ fn gas_price_provider_adding_samples() {
 
     // Add another block to the provider.
     let sample = PriceSample { timestamp: 12, base_fee_per_gas: 12, blob_fee: 13 };
-    provider.add_price_info(BlockNumber(6), sample).unwrap();
+    provider.add_price_info(6, sample).unwrap();
 
     // Should fail because the memory of the provider is full, and we added another block.
     let ret = provider.get_price_info(BlockTimestamp(final_timestamp + lag));
-    matches!(ret, Result::Err(L1GasPriceProviderError::MissingData(_)));
+    matches!(ret, Result::Err(L1GasPriceProviderError::MissingDataError { .. }));
 }
 #[test]
 fn gas_price_provider_timestamp_changes_mean() {
@@ -87,11 +88,11 @@ fn gas_price_provider_timestamp_changes_mean() {
     // Timestamp for sample[3] is used to define the interval of samples 1 to 3.
     let final_timestamp = samples[3].timestamp;
 
-    let (gas_price, data_gas_price) =
+    let PriceInfo { base_fee_per_gas: gas_price, blob_fee: data_gas_price } =
         provider.get_price_info(BlockTimestamp(final_timestamp + lag)).unwrap();
 
     // If we take a higher timestamp the gas prices should change.
-    let (gas_price_new, data_gas_price_new) =
+    let PriceInfo { base_fee_per_gas: gas_price_new, blob_fee: data_gas_price_new } =
         provider.get_price_info(BlockTimestamp(final_timestamp + lag * 2)).unwrap();
     assert_ne!(gas_price_new, gas_price);
     assert_ne!(data_gas_price_new, data_gas_price);
--- a/crates/starknet_l1_gas_price/src/l1_gas_price_scraper.rs
+++ b/crates/starknet_l1_gas_price/src/l1_gas_price_scraper.rs
@@ -5,15 +5,13 @@ use papyrus_base_layer::{BaseLayerContract, L1BlockNumber};
 use papyrus_config::converters::deserialize_float_seconds_to_duration;
 use papyrus_config::validators::validate_ascii;
 use serde::{Deserialize, Serialize};
-use starknet_api::block::BlockNumber;
 use starknet_api::core::ChainId;
-use starknet_l1_gas_price_types::errors::L1GasPriceProviderError;
+use starknet_l1_gas_price_types::errors::L1GasPriceClientError;
+use starknet_l1_gas_price_types::L1GasPriceProviderClient;
 use starknet_sequencer_infra::component_client::ClientError;
 use thiserror::Error;
 use tracing::error;
 use validator::Validate;
-
-use crate::l1_gas_price_provider::L1GasPriceProviderClient;
 
 #[cfg(test)]
 #[path = "l1_gas_price_scraper_test.rs"]
@@ -27,7 +25,7 @@ pub enum L1GasPriceScraperError<T: BaseLayerContract + Send + Sync> {
     #[error("Base layer error: {0}")]
     BaseLayerError(T::Error),
     #[error("Could not update gas price provider: {0}")]
-    GasPriceProviderError(L1GasPriceProviderError),
+    GasPriceClientError(L1GasPriceClientError),
     // Leaky abstraction, these errors should not propagate here.
     #[error(transparent)]
     NetworkError(ClientError),
@@ -94,24 +92,19 @@ impl<B: BaseLayerContract + Send + Sync> L1GasPriceScraper<B> {
         &mut self,
         mut block_num: L1BlockNumber,
     ) -> L1GasPriceScraperResult<L1BlockNumber, B> {
-        loop {
-            let price_sample = self
-                .base_layer
-                .get_price_sample(block_num)
+        while let Some(sample) = self
+            .base_layer
+            .get_price_sample(block_num)
+            .await
+            .map_err(L1GasPriceScraperError::BaseLayerError)?
+        {
+            self.l1_gas_price_provider
+                .add_price_info(block_num, sample)
                 .await
-                .map_err(L1GasPriceScraperError::BaseLayerError)?;
-            match price_sample {
-                None => break,
-                Some(sample) => {
-                    self.l1_gas_price_provider
-                        .add_price_info(BlockNumber(block_num), sample)
-                        .map_err(L1GasPriceScraperError::GasPriceProviderError)?;
+                .map_err(L1GasPriceScraperError::GasPriceClientError)?;
 
-                    block_num += 1;
-                }
-            }
+            block_num += 1;
         }
-
         Ok(block_num)
     }
 }
--- a/crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs
+++ b/crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs
@@ -1,15 +1,14 @@
 use std::sync::Arc;
 
 use papyrus_base_layer::{MockBaseLayerContract, PriceSample};
+use starknet_l1_gas_price_types::MockL1GasPriceProviderClient;
 
-use crate::l1_gas_price_provider::MockL1GasPriceProviderClient;
 use crate::l1_gas_price_scraper::{L1GasPriceScraper, L1GasPriceScraperConfig};
 
 const BLOCK_TIME: u64 = 2;
 const GAS_PRICE: u128 = 42;
 const DATA_PRICE: u128 = 137;
 
-#[allow(clippy::as_conversions)]
 fn setup_scraper(
     end_block: u64,
     expected_number_of_blocks: usize,
@@ -21,8 +20,8 @@ fn setup_scraper(
         } else {
             Ok(Some(PriceSample {
                 timestamp: block_number * BLOCK_TIME,
-                base_fee_per_gas: block_number as u128 * GAS_PRICE,
-                blob_fee: block_number as u128 * DATA_PRICE,
+                base_fee_per_gas: u128::from(block_number) * GAS_PRICE,
+                blob_fee: u128::from(block_number) * DATA_PRICE,
             }))
         }
     });
@@ -30,10 +29,10 @@ fn setup_scraper(
     let mut mock_provider = MockL1GasPriceProviderClient::new();
     mock_provider
         .expect_add_price_info()
-        .withf(|block_number, price_sample| {
-            price_sample.timestamp == block_number.0 * BLOCK_TIME
-                && price_sample.base_fee_per_gas == block_number.0 as u128 * GAS_PRICE
-                && price_sample.blob_fee == block_number.0 as u128 * DATA_PRICE
+        .withf(|&block_number, price_sample| {
+            price_sample.timestamp == block_number * BLOCK_TIME
+                && price_sample.base_fee_per_gas == u128::from(block_number) * GAS_PRICE
+                && price_sample.blob_fee == u128::from(block_number) * DATA_PRICE
         })
         .times(expected_number_of_blocks)
         .returning(|_, _| Ok(()));
@@ -54,7 +53,6 @@ async fn run_l1_gas_price_scraper_single_block() {
     scraper.update_prices(START_BLOCK).await.unwrap();
 }
 
-#[allow(clippy::as_conversions)]
 #[tokio::test]
 async fn run_l1_gas_price_scraper_two_blocks() {
     const START_BLOCK: u64 = 2;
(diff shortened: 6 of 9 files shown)


guy-starkware:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4265?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4081** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4081?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4259** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4259?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4202** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4202?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4266** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4266?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4265** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4265?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4265?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#4124** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4124?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4000** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4000?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4139** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4139?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

asmaastarkware (approved):
why did u make add/get price async?

> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-:-OJYEumU8bs50FfWM0Ki:b-muyued)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @guy-starkware)

guy-starkware (commented):
Ok so now I understand what you are asking :) the old code had the L1GasPriceProviderClient as the definition for L1GasPriceProvider, but this is not what we need. 

The client trait is used by the infrastructure to pass messages to the actual component (which is the provider struct). The implentation of the client trait requires we use async functions, as these are really just channels we "send" into. 

The provider struct has its own impl where the add/get price functions are not async. This is what ultimately gets called inside the server by the infrastructure. 

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-:-OJYLyjY7YqsCt7C2j8p:bgco9lj)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

giladchase (changes requested):
> Reviewed 5 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4265)* status: 6 of 8 files reviewed, 6 unresolved discussions (waiting on @asmaastarkware)

___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_scraper.rs` line 104 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5Myiz45mJKikswOra:-OK5Myj-C9YaZhjVZMJM:bwd898z) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_scraper.rs#L104)):*
> ```XML
>             match price_sample {
>                 None => break,
>                 Some(sample) => {
> ```

Lookup `while let`



_Code quote:_
```XML
        loop {
            let price_sample = self
                .base_layer
                .get_price_sample(block_num)
                .await
                .map_err(L1GasPriceScraperError::BaseLayerError)?;
            match price_sample {
                None => break,
                Some(sample) => {
```
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_scraper.rs` line 106 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5OJcrD0jtzfcav2T-:-OK5OJcrD0jtzfcav2T0:b-e8c1br) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_scraper.rs#L106)):*
> ```XML
>                 Some(sample) => {
>                     self.l1_gas_price_provider
>                         .add_price_info(BlockNumber(block_num), sample)
> ```

If the height argument in add_price_info is an _L1_ height, then it should retain its `L1BlockNumber` height. `BlockNumber` is for Starknet block numbers only.



_Code quote:_
```XML
                        .add_price_info(BlockNumber(block_num), sample)
```
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 30 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5E6fn1d-FrZ6LbRk8:-OK5E6fn1d-FrZ6LbRk9:b428w0f) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L30)):*
> ```XML
>             storage_limit: usize::try_from(10 * MEAN_NUMBER_OF_BLOCKS).unwrap(),
>         }
>     }
> ```

Default is typically expected to be trivial values, so i'd expect these to be all zeros. If you want common initial values I suggest adding them to the `new` constructor.



_Code quote:_
```XML
    fn default() -> Self {
        const MEAN_NUMBER_OF_BLOCKS: u64 = 300;
        Self {
            number_of_blocks_for_mean: MEAN_NUMBER_OF_BLOCKS,
            lag_margin_seconds: 60,
            storage_limit: usize::try_from(10 * MEAN_NUMBER_OF_BLOCKS).unwrap(),
        }
    }
```
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 36 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5EhEqAym0OA1IDB1K:-OK5EhEqAym0OA1IDB1L:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L36)):*
> ```XML
> pub struct L1GasPriceProvider {
>     config: L1GasPriceProviderConfig,
>     data: VecDeque<(BlockNumber, PriceSample)>,
> ```





_Suggestion:_
```XML
    price_samples_by_block: VecDeque<BlockPriceSample>,
```
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 41 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5KJhp2YqXyDpslJmm:-OK5KJhp2YqXyDpslJmn:bq31jk9) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L41)):*
> ```XML
> impl L1GasPriceProvider {
>     pub fn new(config: L1GasPriceProviderConfig) -> Self {
>         Self { config, data: VecDeque::new() }
> ```

Saves re-allocations, we're gonna always reach this capacity so might as well alloc it once.



_Suggestion:_
```XML
        Self { config, data: VecDeque::with_capacity(config.storage_limit) }
```
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 49 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5FyLYCzSNfUGkGuUK:-OK5FyLYCzSNfUGkGuUL:b-s3d4ul) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L49)):*
> ```XML
>         sample: PriceSample,
>     ) -> L1GasPriceProviderResult<()> {
>         let last_plus_one = self.data.back().map(|(h, _)| h.0 + 1).unwrap_or(0);
> ```

When operating on types always look up existing APIs, and consider what order of operations makes most tvedesense when using combinators. 

Also in the original code it was unwrap_or(0) (now it becomes 1 after the unchecked_next), was that correct? 



_Suggestion:_
```XML
        let last_plus_one = self.data.back().unwrap_or_default().unchecked_next();
```
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 59 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5KwzjEhkVHiPQMvPS:-OK5KwzjEhkVHiPQMvPT:bjt54l9) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L59)):*
> ```XML
>         if self.data.len() > self.config.storage_limit {
>             self.data.pop_front();
>         }
> ```

This is too hidden.
Consider wrapping this `VecDeque` with a very simple new DS called `RingBuffer`, that supports a single `push` operation that handles this circular push logic (I'm assuming we're only pushing_back)
The VecDeque type is communicating "unbounded ring-buffer",  and readers will not know that it has a hidden size limit without catching these 4 lines



_Code quote:_
```XML
        self.data.push_back((height, sample));
        if self.data.len() > self.config.storage_limit {
            self.data.pop_front();
        }
```
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 66 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5D8q_CJ4EQEoeUKuG:-OK5D8q_CJ4EQEoeUKuH:brc11bm) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L66)):*
> ```XML
>         &self,
>         timestamp: BlockTimestamp,
>     ) -> L1GasPriceProviderResult<(u128, u128)> {
> ```

Sig should be descriptive, reader shoudl not have to read the function or other functions in order to understand what these numbers are.

If there are other functions that use these two numbers they should also use this abstraction.



_Suggestion:_
```XML
    ) -> L1GasPriceProviderResult<PriceInfo> {
```
___
*[`crates/starknet_l1_gas_price_types/src/errors.rs` line 8 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5AeKqCKlY_ctMN1mg:-OK5AeKqCKlY_ctMN1mh:bk02s9j) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price_types/src/errors.rs#L8)):*
> ```XML
> pub enum L1GasPriceProviderError {
>     #[error("Failed to add price info: {0}")]
>     InvalidHeight(String),
> ```

This is only used once, and the string used is more descriptive than the one used here, make the error string `Block height is not consecutive: expected {expected}, got {height}",` and rename the variant into `UnexpectedHeight{expected: u64, height: height}`. Look over the other errors in the same way, the error should be descriptive, doesn't have to be generic, you can always add more variants to the enum if you have different errors related to height.

Also by convention all error variants should end with Error.



_Code quote:_
```XML
    #[error("Failed to add price info: {0}")]
    InvalidHeight(String),
```

guy-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4265)* status: 1 of 9 files reviewed, 7 unresolved discussions (waiting on @asmaastarkware and @giladchase)

___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 30 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5E6fn1d-FrZ6LbRk8:-OK5XtFs8IVmAvNFHFHq:b-oct6m) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L30)):*

Default is typically expected to be trivial values, so i'd expect these to be all zeros. If you want common initial values I suggest adding them to the `new` constructor.
</blockquote></details>

I'm not sure that's true for config structs. 

Assuming we don't make these "required" parameters, this is probably the place where I will define their default values. 
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 49 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5FyLYCzSNfUGkGuUK:-OK5Z-Ec89hjqGuEUKap:bkarkbk) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L49)):*

When operating on types always look up existing APIs, and consider what order of operations makes most tvedesense when using combinators. 

Also in the original code it was unwrap_or(0) (now it becomes 1 after the unchecked_next), was that correct?
</blockquote></details>

This would probably work if it were returning `Option(BlockNumber)` but I'm returning an option on a tuple. 

Also, I really want it to return 0 in case on None, because that sets me up to correctly expect height zero in case the VecDeque is empty. 

(there's a bug in here that if it is indeed empty then the only acceptable height is zero, but I fix that in the next PR). 
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 59 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5KwzjEhkVHiPQMvPS:-OK5_GSI29vdFFNbSUFu:buhecw1) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L59)):*

This is too hidden.
Consider wrapping this `VecDeque` with a very simple new DS called `RingBuffer`, that supports a single `push` operation that handles this circular push logic (I'm assuming we're only pushing_back)
The VecDeque type is communicating "unbounded ring-buffer",  and readers will not know that it has a hidden size limit without catching these 4 lines
</blockquote></details>

I like this idea! 

In fact I love it so much I'm doing it in a separate PR. 
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 66 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5D8q_CJ4EQEoeUKuG:-OK5b6xl3MMFe_MbBPuf:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L66)):*

Sig should be descriptive, reader shoudl not have to read the function or other functions in order to understand what these numbers are.

If there are other functions that use these two numbers they should also use this abstraction.
</blockquote></details>

Done.
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_scraper.rs` line 104 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5Myiz45mJKikswOra:-OK5bbdU6inwxXvbqo_5:b-wiyacz) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_scraper.rs#L104)):*

Lookup `while let`
</blockquote></details>

Very cool! 
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_scraper.rs` line 106 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5OJcrD0jtzfcav2T-:-OK5dunnD7ExQQTAU5m4:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_scraper.rs#L106)):*

If the height argument in add_price_info is an _L1_ height, then it should retain its `L1BlockNumber` height. `BlockNumber` is for Starknet block numbers only.
</blockquote></details>

Done.
___
*[`crates/starknet_l1_gas_price_types/src/errors.rs` line 8 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5AeKqCKlY_ctMN1mg:-OK5g2t6EkENKg9W3AfN:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price_types/src/errors.rs#L8)):*

This is only used once, and the string used is more descriptive than the one used here, make the error string `Block height is not consecutive: expected {expected}, got {height}",` and rename the variant into `UnexpectedHeight{expected: u64, height: height}`. Look over the other errors in the same way, the error should be descriptive, doesn't have to be generic, you can always add more variants to the enum if you have different errors related to height.

Also by convention all error variants should end with Error.
</blockquote></details>

Done.
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 36 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5EhEqAym0OA1IDB1K:-OK5YDxs5AZMVHfIGDPo:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L36)):*
> ```XML
> pub struct L1GasPriceProvider {
>     config: L1GasPriceProviderConfig,
>     data: VecDeque<(BlockNumber, PriceSample)>,
> ```

Done.

giladchase (changes requested):
> Reviewed 3 of 5 files at r3, 1 of 3 files at r4, 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4265)* status: 7 of 9 files reviewed, 4 unresolved discussions (waiting on @asmaastarkware and @guy-starkware)

___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 30 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5E6fn1d-FrZ6LbRk8:-OKX5YoC5Tg1NKG2JCyW:bw6v0lk) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L30)):*
<details><summary><i>Previously, guy-starkware wrote…</i></summary><blockquote>

I'm not sure that's true for config structs. 

Assuming we don't make these "required" parameters, this is probably the place where I will define their default values.
</blockquote></details>

You mean its a requirement from infra? they are forcing us to make default use nontrivial values?
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 36 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5EhEqAym0OA1IDB1K:-OKX6EcR7F-h5M57uEqd:b-5vh72z) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L36)):*
<details><summary><i>Previously, guy-starkware wrote…</i></summary><blockquote>

Done.
</blockquote></details>

also add a struct instead of the tuple plz
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 49 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5FyLYCzSNfUGkGuUK:-OKX6WTB8NRHxjAkOkTI:byzg4pj) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L49)):*
<details><summary><i>Previously, guy-starkware wrote…</i></summary><blockquote>

This would probably work if it were returning `Option(BlockNumber)` but I'm returning an option on a tuple. 

Also, I really want it to return 0 in case on None, because that sets me up to correctly expect height zero in case the VecDeque is empty. 

(there's a bug in here that if it is indeed empty then the only acceptable height is zero, but I fix that in the next PR).
</blockquote></details>

I think i'm missing something, where is the option here?
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 66 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5D8q_CJ4EQEoeUKuG:-OKX5q8Q6hCLsROMT8-z:b4cmiu1) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L66)):*
<details><summary><i>Previously, guy-starkware wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Struct plz, if they can be named they should in most cases
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs` line 101 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OKX8Hot34WSJZ1cM_CD:-OKX8Hot34WSJZ1cM_CE:bf6n2nd) ([raw file](https://github.com/starkware-libs/sequencer/blob/244640e3a2ae9beaa551e3a143fffe0a1818da50/crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs#L101)):*
> ```XML
>         .withf(|block_number, price_sample| {
>             price_sample.timestamp == *block_number * BLOCK_TIME
>                 && price_sample.base_fee_per_gas == *block_number as u128 * GAS_PRICE
> ```

Not sure how that `as` passed clippy :scream: 
We have a ban on it on Cargo.toml, can you replace it with `from` or `into`?
The reason it was banned was because its a bit evil, it truncates and saturates surprisingly, we caught it hiding bugs in tests more than once in the past.

I'll check async why clippy wasn't enforced on this test



_Code quote:_
```XML
                && price_sample.base_fee_per_gas == *block_number as u128 * GAS_PRICE
```
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs` line 102 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OKX7h7LD4wcups5TOsg:-OKX7h7LD4wcups5TOsh:b45h) ([raw file](https://github.com/starkware-libs/sequencer/blob/244640e3a2ae9beaa551e3a143fffe0a1818da50/crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs#L102)):*
> ```XML
>             price_sample.timestamp == *block_number * BLOCK_TIME
>                 && price_sample.base_fee_per_gas == *block_number as u128 * GAS_PRICE
>                 && price_sample.blob_fee == *block_number as u128 * DATA_PRICE
> ```





_Suggestion:_
```XML
        .withf(|&block_number, price_sample| {
            price_sample.timestamp == block_number * BLOCK_TIME
                && price_sample.base_fee_per_gas == block_number as u128 * GAS_PRICE
                && price_sample.blob_fee == block_number as u128 * DATA_PRICE
```
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider_test.rs` line 81 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OKX6pgcEbzXOhzyj0X_:-OKX6pgcEbzXOhzyj0Xa:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/244640e3a2ae9beaa551e3a143fffe0a1818da50/crates/starknet_l1_gas_price/src/l1_gas_price_provider_test.rs#L81)):*
> ```XML
>     // Should fail because the memory of the provider is full, and we added another block.
>     let ret = provider.get_price_info(BlockTimestamp(final_timestamp + lag));
>     matches!(ret, Result::Err(L1GasPriceProviderError::MissingDataError { timestamp: _, lag: _ }));
> ```





_Suggestion:_
```XML
    matches!(ret, Result::Err(L1GasPriceProviderError::MissingDataError { .. }));
```

guy-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4265)* status: 7 of 9 files reviewed, 4 unresolved discussions (waiting on @asmaastarkware and @giladchase)

___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 30 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5E6fn1d-FrZ6LbRk8:-OK_W4LYCHD_WRpMgqyI:b-qhxfx6) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L30)):*

You mean its a requirement from infra? they are forcing us to make default use nontrivial values?
</blockquote></details>

I don't know if that is a requirement but I think that is the convention. 
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 36 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5EhEqAym0OA1IDB1K:-OK_aM0l6nbhUcpU-qas:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L36)):*

also add a struct instead of the tuple plz
</blockquote></details>

Done.
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 49 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5FyLYCzSNfUGkGuUK:-OK_aUjH5Psb1w_9QbqA:b3tr59q) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L49)):*

I think i'm missing something, where is the option here?
</blockquote></details>

When I do `.back()` I can either get the last thing on the vector or None (if it is empty). The `map` allows me to do calculations on the Some and pass the None (which I replace with zero using `unwrap_or`). 
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 66 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5D8q_CJ4EQEoeUKuG:-OK_bwBJ0_6kWvUNogMd:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L66)):*

Struct plz, if they can be named they should in most cases
</blockquote></details>

Done.
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs` line 101 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OKX8Hot34WSJZ1cM_CD:-OK_cGPO3YGkEeo8DY70:b-h08qcs) ([raw file](https://github.com/starkware-libs/sequencer/blob/244640e3a2ae9beaa551e3a143fffe0a1818da50/crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs#L101)):*

Not sure how that `as` passed clippy :scream: 
We have a ban on it on Cargo.toml, can you replace it with `from` or `into`?
The reason it was banned was because its a bit evil, it truncates and saturates surprisingly, we caught it hiding bugs in tests more than once in the past.

I'll check async why clippy wasn't enforced on this test
</blockquote></details>

I think it was allowed because I added `#\[allow(clippy::as\_conversions)\]`
I got the impression we are ok with this conversion inside tests. 

In this specific case I was using small numbers (which, in the context of `u128 as u64` is a fairly broad range on numbers). 

I removed the conversions as well as the "allow" decorator. 
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs` line 102 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OKX7h7LD4wcups5TOsg:-OK_dqFZ8KMdNnxE5T8F:b9vf7pw) ([raw file](https://github.com/starkware-libs/sequencer/blob/244640e3a2ae9beaa551e3a143fffe0a1818da50/crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs#L102)):*
> ```XML
>             price_sample.timestamp == *block_number * BLOCK_TIME
>                 && price_sample.base_fee_per_gas == *block_number as u128 * GAS_PRICE
>                 && price_sample.blob_fee == *block_number as u128 * DATA_PRICE
> ```

Good idea!

giladchase (changes requested):
> Reviewed 5 of 5 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4265)* status: all files reviewed, 3 unresolved discussions (waiting on @guy-starkware)

___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 30 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5E6fn1d-FrZ6LbRk8:-OKuEODM4iY9yAA1Y4mJ:b-6qhkts) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L30)):*
<details><summary><i>Previously, guy-starkware wrote…</i></summary><blockquote>

I don't know if that is a requirement but I think that is the convention.
</blockquote></details>

Can you verify? cause this is breaking rust convention in which defaults should be trivial.
When ppl see `..Default::default()` they read it as `..dont_care()`, and if there's meaning there it might be surprising?
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 49 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5FyLYCzSNfUGkGuUK:-OKuEtcyAkpniV3R6eJ5:b-tgph3p) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L49)):*
<details><summary><i>Previously, guy-starkware wrote…</i></summary><blockquote>

When I do `.back()` I can either get the last thing on the vector or None (if it is empty). The `map` allows me to do calculations on the Some and pass the None (which I replace with zero using `unwrap_or`).
</blockquote></details>

I think there's a disconnect, `Option<u64>.unwrap_or_default()` is the same as `Option<u64>.unwrap_or(0)`.

Also i'm not sure i understand why it's `unwrap_or(0)` and not `unwrap_or(1)`, if this value is `last_plus_one`, then the none-flow is not plus-oned (because it's returning 0). Is that correct? Can you tell me what the flow is here? why are we plus-1ing and in what flow can the data be None?

RE the map, plz avoid `.0` if you can help it. In this instance unchecked_next does the same thing and is part of the  object's API.
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs` line 101 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OKX8Hot34WSJZ1cM_CD:-OKuGJ-N-VzFnKBV_7re:b-wp7f7f) ([raw file](https://github.com/starkware-libs/sequencer/blob/244640e3a2ae9beaa551e3a143fffe0a1818da50/crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs#L101)):*
<details><summary><i>Previously, guy-starkware wrote…</i></summary><blockquote>

I think it was allowed because I added `#\[allow(clippy::as\_conversions)\]`
I got the impression we are ok with this conversion inside tests. 

In this specific case I was using small numbers (which, in the context of `u128 as u64` is a fairly broad range on numbers). 

I removed the conversions as well as the "allow" decorator.
</blockquote></details>

Tnx.
I know it sounds yekke but we got burned too many times, and `into` isn't that bad (except for the size<-> u64 try_into's which _are_ very much yekke :unamused: )
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs` line 67 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OKuLiD7BKdV4tNUgwqr:-OKuLiD7BKdV4tNUgwqs:b-c5k85s) ([raw file](https://github.com/starkware-libs/sequencer/blob/67f8ad70c4e50597a86736f9fc79f2b9fa5b922a/crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs#L67)):*
> ```XML
>     mock_contract
>         .expect_get_price_sample()
>         .times(usize::try_from(END_BLOCK1).unwrap() - usize::try_from(START_BLOCK).unwrap() + 1)
> ```

also other places



_Suggestion:_
```XML
   .times(usize::try_from(END_BLOCK1 - START_BLOCK + 1).unwrap())
```
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 105 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OKuJRMABgvSqRSq-LJg:-OKuJRMABgvSqRSq-LJh:be3aym0) ([raw file](https://github.com/starkware-libs/sequencer/blob/67f8ad70c4e50597a86736f9fc79f2b9fa5b922a/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L105)):*
> ```XML
>             gas_price += data.sample.base_fee_per_gas;
>             data_gas_price += data.sample.blob_fee;
>         }
> ```

Technically we can also fold into a `PriceInfo`, but it feels more complex maybe given that we divide later.



_Suggestion:_
```XML
        let (gas_price, data_gas_price) =
            self.price_samples_by_block.iter().skip(last_index - num_blocks).take(num_blocks).fold(
                (0, 0),
                |(sum_base, sum_blob), data| {
                    (sum_base + data.sample.base_fee_per_gas, sum_blob + data.sample.blob_fee)
                },
            );
```
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 108 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OKuK1lh8gaNtBem13-o:-OKuK1lh8gaNtBem13-p:b-1t06g0) ([raw file](https://github.com/starkware-libs/sequencer/blob/67f8ad70c4e50597a86736f9fc79f2b9fa5b922a/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L108)):*
> ```XML
>         Ok(PriceInfo {
>             base_fee_per_gas: gas_price / u128::from(self.config.number_of_blocks_for_mean),
>             blob_fee: data_gas_price / u128::from(self.config.number_of_blocks_for_mean),
> ```

Where do we protect against numerical errors? Like 0 div, or the division result comes out 0 (which might cause issues down the line to users of this API).

Were all of these edge cases discussed previously? (OK if they weren't, but if they weren't I suggest moving back to WIP stage and going over all arithmetic ops) 



_Code quote:_
```XML
            base_fee_per_gas: gas_price / u128::from(self.config.number_of_blocks_for_mean),
            blob_fee: data_gas_price / u128::from(self.config.number_of_blocks_for_mean),
```

guy-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4265)* status: all files reviewed, 3 unresolved discussions (waiting on @giladchase)

___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 30 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5E6fn1d-FrZ6LbRk8:-OKuZk6o7G_RM1dXJg83:b48ndcq) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L30)):*

Can you verify? cause this is breaking rust convention in which defaults should be trivial.
When ppl see `..Default::default()` they read it as `..dont_care()`, and if there's meaning there it might be surprising?
</blockquote></details>

I can give you an example: in the L1 message handler there are configs (for both scraper and provider) that are not setting all values to zero when you call default for them... 
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 49 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5FyLYCzSNfUGkGuUK:-OKu_0DREOBZSSRCv6Ja:b-3d0rx9) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L49)):*

I think there's a disconnect, `Option<u64>.unwrap_or_default()` is the same as `Option<u64>.unwrap_or(0)`.

Also i'm not sure i understand why it's `unwrap_or(0)` and not `unwrap_or(1)`, if this value is `last_plus_one`, then the none-flow is not plus-oned (because it's returning 0). Is that correct? Can you tell me what the flow is here? why are we plus-1ing and in what flow can the data be None?

RE the map, plz avoid `.0` if you can help it. In this instance unchecked_next does the same thing and is part of the  object's API.
</blockquote></details>

The flow is this: 
- If there is anything in the vecdeq, give me the next number after it. 
- If the vecdeq is empty, give me zero. 
We want it this way because "last_plus_one" represents the number we expect to receive and put into the vecdeq in consecutive order after what is there (or not). 

If I could, I would have simplified it to "give me the last number or if none, give me -1" but I cannot put -1 into this u64. 

I am also simplifying this code in a later PR, so let's not litigate this too much (although I am happy to explain it thoroughly). 
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 105 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OKuJRMABgvSqRSq-LJg:-OKuaBky5hxY0TQhv931:bxv0ymg) ([raw file](https://github.com/starkware-libs/sequencer/blob/67f8ad70c4e50597a86736f9fc79f2b9fa5b922a/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L105)):*

Technically we can also fold into a `PriceInfo`, but it feels more complex maybe given that we divide later.
</blockquote></details>

I accept your change although I am not sure it is better. It is not less code, and I don't think it is more readable than a loop (where the operations are completely explicite). 

Maybe it is more "rustic" but I don't feel that strongly about that, either way. 
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 108 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OKuK1lh8gaNtBem13-o:-OKuab1u3-YFlTBvkE7Q:b-xs6nc0) ([raw file](https://github.com/starkware-libs/sequencer/blob/67f8ad70c4e50597a86736f9fc79f2b9fa5b922a/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L108)):*

Where do we protect against numerical errors? Like 0 div, or the division result comes out 0 (which might cause issues down the line to users of this API).

Were all of these edge cases discussed previously? (OK if they weren't, but if they weren't I suggest moving back to WIP stage and going over all arithmetic ops)
</blockquote></details>

Divide by zero can only happen if the config field `number_of_blocks_for_mean` is zero. I don't think that can happen (it will probably fail elsewhere first). 
If the sum comes out to zero (or less than 1, which is rounded to zero) then it is a surprising result but I don't think that's very likely to occur, given that gas prices are generally much larger than 1. 

These edge cases were not discussed, AFAIK, but I'm not sure about your idea to move it back to WIP? It seems like we should discuss and then open another PR to fix/protect against any edge cases we are worried about. As it stands now, this code is correct in the "happy flow" which AFAIK is the thing we do first, and then address the "unhappy flows". Or maybe I'm missing some other consideration in this case? 
___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs` line 67 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OKuLiD7BKdV4tNUgwqr:-OKuc-3m4yFws5kXK7Re:b-7xt2v3) ([raw file](https://github.com/starkware-libs/sequencer/blob/67f8ad70c4e50597a86736f9fc79f2b9fa5b922a/crates/starknet_l1_gas_price/src/l1_gas_price_scraper_test.rs#L67)):*

also other places
</blockquote></details>

Yes!

giladchase (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-:-OL-LDIk6UaEKXSYFzXr:bnfp4nl)* status: 7 of 9 files reviewed, all discussions resolved (waiting on @guy-starkware)

___
*[`crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs` line 49 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4265#-OK5FyLYCzSNfUGkGuUK:-OL-KewJ0gt-v6A6cvMs:b9yiujl) ([raw file](https://github.com/starkware-libs/sequencer/blob/30b3bf6968d7728c8305332431743f48d30ee289/crates/starknet_l1_gas_price/src/l1_gas_price_provider.rs#L49)):*
<details><summary><i>Previously, guy-starkware wrote…</i></summary><blockquote>

The flow is this: 
- If there is anything in the vecdeq, give me the next number after it. 
- If the vecdeq is empty, give me zero. 
We want it this way because "last_plus_one" represents the number we expect to receive and put into the vecdeq in consecutive order after what is there (or not). 

If I could, I would have simplified it to "give me the last number or if none, give me -1" but I cannot put -1 into this u64. 

I am also simplifying this code in a later PR, so let's not litigate this too much (although I am happy to explain it thoroughly).
</blockquote></details>

Unchecked_next + variable name change, nonblocking

PR #4322, opened 2025-02-20
refactor(consensus): begin rewriting tests for the stream handler

matan-starkware:
Goals:
1. Utilize new genericiy to just use simple channels instead of using mock_register_broadcast_topic
2. Tests should try to use only the public API (ie messages in and out)
    - This may be broken later when testing how malicious cases are defended

### DIFF
--- a/crates/starknet_consensus/src/stream_handler_test.rs
+++ b/crates/starknet_consensus/src/stream_handler_test.rs
@@ -547,3 +547,196 @@ mod tests {
         );
     }
 }
+
+mod tests_v2 {
+    use std::collections::BTreeSet;
+
+    use futures::channel::mpsc::{self, Receiver, SendError, Sender};
+    use futures::{FutureExt, SinkExt, StreamExt};
+    use papyrus_network::network_manager::{BroadcastTopicClientTrait, ReceivedBroadcastedMessage};
+    use papyrus_network_types::network_types::BroadcastedMessageMetadata;
+    use papyrus_protobuf::consensus::{ProposalInit, ProposalPart, StreamMessageBody};
+
+    use super::{TestStreamId, CHANNEL_SIZE};
+    use crate::stream_handler::StreamHandler;
+
+    type StreamMessage = papyrus_protobuf::consensus::StreamMessage<ProposalPart, TestStreamId>;
+
+    struct FakeBroadcastClient {
+        sender: Sender<StreamMessage>,
+    }
+
+    #[async_trait::async_trait]
+    impl BroadcastTopicClientTrait<StreamMessage> for FakeBroadcastClient {
+        async fn broadcast_message(&mut self, message: StreamMessage) -> Result<(), SendError> {
+            self.sender.send(message).await
+        }
+
+        async fn report_peer(&mut self, _: BroadcastedMessageMetadata) -> Result<(), SendError> {
+            todo!()
+        }
+
+        async fn continue_propagation(
+            &mut self,
+            _: &BroadcastedMessageMetadata,
+        ) -> Result<(), SendError> {
+            todo!()
+        }
+    }
+
+    #[allow(clippy::type_complexity)]
+    fn setup() -> (
+        StreamHandler<
+            ProposalPart,
+            TestStreamId,
+            Receiver<ReceivedBroadcastedMessage<StreamMessage>>,
+            FakeBroadcastClient,
+        >,
+        Sender<ReceivedBroadcastedMessage<StreamMessage>>,
+        Receiver<Receiver<ProposalPart>>,
+        Sender<(TestStreamId, Receiver<ProposalPart>)>,
+        Receiver<StreamMessage>,
+    ) {
+        let (inbound_internal_sender, inbound_internal_receiver) = mpsc::channel(CHANNEL_SIZE);
+        let (inbound_network_sender, inbound_network_receiver) = mpsc::channel(CHANNEL_SIZE);
+        let (outbound_internal_sender, outbound_internal_receiver) = mpsc::channel(CHANNEL_SIZE);
+        let (outbound_network_sender, outbound_network_receiver) = mpsc::channel(CHANNEL_SIZE);
+        let outbound_network_sender = FakeBroadcastClient { sender: outbound_network_sender };
+        let stream_handler = StreamHandler::new(
+            inbound_internal_sender,
+            inbound_network_receiver,
+            outbound_internal_receiver,
+            outbound_network_sender,
+        );
+
+        (
+            stream_handler,
+            inbound_network_sender,
+            inbound_internal_receiver,
+            outbound_internal_sender,
+            outbound_network_receiver,
+        )
+    }
+
+    fn build_init_message(round: u32, stream_id: u64, message_id: u32) -> StreamMessage {
+        StreamMessage {
+            message: StreamMessageBody::Content(ProposalPart::Init(ProposalInit {
+                round,
+                ..Default::default()
+            })),
+            stream_id: TestStreamId(stream_id),
+            message_id: message_id.into(),
+        }
+    }
+
+    fn build_fin_message(stream_id: u64, message_id: u32) -> StreamMessage {
+        StreamMessage {
+            message: StreamMessageBody::Fin,
+            stream_id: TestStreamId(stream_id),
+            message_id: message_id.into(),
+        }
+    }
+
+    #[tokio::test]
+    async fn outbound_single() {
+        let num_messages = 5;
+        let stream_id = 1;
+        let (
+            mut stream_handler,
+            _,
+            _,
+            mut client_to_streamhandler_sender,
+            mut streamhandler_to_network_receiver,
+        ) = setup();
+
+        // Create a new stream to send.
+        let (mut sender, stream_receiver) = mpsc::channel(CHANNEL_SIZE);
+        client_to_streamhandler_sender
+            .send((TestStreamId(stream_id), stream_receiver))
+            .await
+            .unwrap();
+        stream_handler.handle_next_msg().await;
+
+        // Send the content of the stream.
+        for i in 0..num_messages {
+            let init = ProposalPart::Init(ProposalInit { round: i, ..Default::default() });
+            sender.send(init).await.unwrap();
+        }
+
+        // Check the content is sent to the network in order.
+        for i in 0..num_messages {
+            stream_handler.handle_next_msg().await;
+            let actual = streamhandler_to_network_receiver.next().now_or_never().unwrap().unwrap();
+            assert_eq!(actual, build_init_message(i, stream_id, i));
+        }
+
+        // Close the stream and check that a Fin is sent to the network.
+        sender.close_channel();
+        stream_handler.handle_next_msg().await;
+        assert_eq!(
+            streamhandler_to_network_receiver.next().now_or_never().unwrap().unwrap(),
+            build_fin_message(stream_id, num_messages)
+        );
+    }
+
+    #[tokio::test]
+    async fn outbound_multiple() {
+        let num_messages = 5;
+        let num_streams = 3;
+        let (
+            mut stream_handler,
+            _,
+            _,
+            mut client_to_streamhandler_sender,
+            mut streamhandler_to_network_receiver,
+        ) = setup();
+
+        // client opens up multiple outbound streams.
+        let mut stream_senders = Vec::new();
+        for stream_id in 0..num_streams {
+            let (sender, stream_receiver) = mpsc::channel(CHANNEL_SIZE);
+            stream_senders.push(sender);
+            client_to_streamhandler_sender
+                .send((TestStreamId(stream_id), stream_receiver))
+                .await
+                .unwrap();
+            stream_handler.handle_next_msg().await;
+        }
+
+        // Send messages on all of the streams.
+        for stream_id in 0..num_streams {
+            let sender =
+                stream_senders.get_mut(TryInto::<usize>::try_into(stream_id).unwrap()).unwrap();
+            for i in 0..num_messages {
+                let init = ProposalPart::Init(ProposalInit { round: i, ..Default::default() });
+                sender.send(init).await.unwrap();
+            }
+        }
+
+        // {StreamId : [Msgs]} - asserts order received matches expected order per stream.
+        let mut expected_msgs = (0..num_streams).map(|_| Vec::new()).collect::<Vec<_>>();
+        let mut actual_msgs = expected_msgs.clone();
+        for stream_id in 0..num_streams {
+            for i in 0..num_messages {
+                // The order the stream handler selects from among multiple streams is undefined.
+                stream_handler.handle_next_msg().await;
+                let msg = streamhandler_to_network_receiver.next().now_or_never().unwrap().unwrap();
+                actual_msgs[TryInto::<usize>::try_into(msg.stream_id.0).unwrap()].push(msg);
+                expected_msgs[TryInto::<usize>::try_into(stream_id).unwrap()]
+                    .push(build_init_message(i, stream_id, i));
+            }
+        }
+        assert_eq!(actual_msgs, expected_msgs);
+
+        // Drop all the senders and check Fins are sent.
+        stream_senders.clear();
+        let mut stream_ids = (0..num_streams).collect::<BTreeSet<_>>();
+        for _ in 0..num_streams {
+            stream_handler.handle_next_msg().await;
+            let fin = streamhandler_to_network_receiver.next().now_or_never().unwrap().unwrap();
+            assert_eq!(fin.message, StreamMessageBody::Fin);
+            assert_eq!(fin.message_id, u64::from(num_messages));
+            assert!(stream_ids.remove(&fin.stream_id.0));
+        }
+    }
+}


guy-starkware (commented):
This is much clearer! 

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-:-OJYlwU7ERj2YdTEBBbj:bj2fajj)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @giladchase)

guy-starkware (approved):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4322)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4322)* status: all files reviewed, 1 unresolved discussion (waiting on @matan-starkware)

___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 552 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKWY_xD39VyscGu81eN:-OKWY_xE5YPV5Y_KTl2f:b-8u87yt) ([raw file](https://github.com/starkware-libs/sequencer/blob/9c2266e8afc862db6a85a45b2402dc41869b97c8/crates/starknet_consensus/src/stream_handler_test.rs#L552)):*
> ```XML
> 
> #[cfg(test)]
> mod tests_v2 {
> ```

I'm not sure i understand why this is necessary, also the other mod tests one.
Asking cause the whole file is a test file and already compiled with cfg(test)



_Code quote:_
```XML
#[cfg(test)]
mod tests_v2 {
```
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 634 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbO6ab4GqRBAP0Knjr:-OKbO6ab4GqRBAP0Knjs:byy4ql1) ([raw file](https://github.com/starkware-libs/sequencer/blob/9c2266e8afc862db6a85a45b2402dc41869b97c8/crates/starknet_consensus/src/stream_handler_test.rs#L634)):*
> ```XML
> 
>     #[tokio::test]
>     async fn outbound_in_order_single() {
> ```

There is quite a bit of boilerplate so I'm not sure which API is tested.
Also what is in order? 



_Code quote:_
```XML
async fn outbound_in_order_single() 
```
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 642 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbMskF8DWWpTS-M_-Q:-OKbMskF8DWWpTS-M_-R:b-fur1hx) ([raw file](https://github.com/starkware-libs/sequencer/blob/9c2266e8afc862db6a85a45b2402dc41869b97c8/crates/starknet_consensus/src/stream_handler_test.rs#L642)):*
> ```XML
>         let (mut sender, stream_receiver) = mpsc::channel(CHANNEL_SIZE);
>         outbound_internal_sender.send((stream_id, stream_receiver)).await.unwrap();
>         stream_handler.handle_next_msg().await;
> ```

what message is sent here? 
More specifically, what is outbound_internal_sender?



_Code quote:_
```XML
        outbound_internal_sender.send((stream_id, stream_receiver)).await.unwrap();
        stream_handler.handle_next_msg().await;
```
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 656 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbNtgcC6P5rLti-Gcr:-OKbNtgdBN-f86cfE-0J:bhljz38) ([raw file](https://github.com/starkware-libs/sequencer/blob/9c2266e8afc862db6a85a45b2402dc41869b97c8/crates/starknet_consensus/src/stream_handler_test.rs#L656)):*
> ```XML
> 
>         sender.close_channel();
>         stream_handler.handle_next_msg().await;
> ```

what message is this? a close message that is sent in the previous line?



_Code quote:_
```XML
        stream_handler.handle_next_msg().await;
```
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 683 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbPyAl3akk56p1RDu_:-OKbPyAl3akk56p1RDua:b-j96b3l) ([raw file](https://github.com/starkware-libs/sequencer/blob/9c2266e8afc862db6a85a45b2402dc41869b97c8/crates/starknet_consensus/src/stream_handler_test.rs#L683)):*
> ```XML
>             let sender = stream_senders.get_mut(stream_id as usize).unwrap();
>             for _ in 0..num_messages {
>                 sender.send(ProposalPart::Init(ProposalInit::default())).await.unwrap();
> ```

Should this use make_test?



_Code quote:_
```XML
                sender.send(ProposalPart::Init(ProposalInit::default())).await.unwrap();
```
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 695 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbPA7O4ScpR4oa4lFi:-OKbPA7O4ScpR4oa4lFj:bljdj9o) ([raw file](https://github.com/starkware-libs/sequencer/blob/9c2266e8afc862db6a85a45b2402dc41869b97c8/crates/starknet_consensus/src/stream_handler_test.rs#L695)):*
> ```XML
>                 stream_handler.handle_next_msg().await;
>                 let msg = outbound_network_receiver.next().now_or_never().unwrap().unwrap();
>                 actual_msgs[msg.stream_id.0 as usize].push(msg);
> ```

Not sure how this got past clippy, Lior banned `as` due to surprise truncations and saturations.



_Code quote:_
```XML
as usize
```
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 703 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbQ3cm2NDh8duA9wLK:-OKbQ3cm2NDh8duA9wLL:b-3izsse) ([raw file](https://github.com/starkware-libs/sequencer/blob/9c2266e8afc862db6a85a45b2402dc41869b97c8/crates/starknet_consensus/src/stream_handler_test.rs#L703)):*
> ```XML
>             }
>         }
>         assert_eq!(actual_msgs, expected_msgs);
> ```

What can cause this assert to fail?  Does this check that the stream doesn't shuffle the messages?



_Code quote:_
```XML
        assert_eq!(actual_msgs, expected_msgs);
```

matan-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4322)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @giladchase and @guy-starkware)

___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbQZQl0wZVwfZyTF_V:-OKbWfRwBpHy6kk8iStP:bk9qtph):*

Perhaps it should be a cargo integration test to enforce this? (as in a test in `tests/`). Not sure, these tests tend to cause problems.
Alternatively they can also be added as unit tests in starknet-integration-tests, then they will also have to use public api.
</blockquote></details>

If we do this here then we need to do this everywhere. My point is that the existing set of tests does look at private fields and I am making an entirely different set of tests. This is not fundamentally different than what we try to do for all unit tests though.
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 552 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKWY_xD39VyscGu81eN:-OKbX-EWDZTDBTkFRdnr:b-7qwp84) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L552)):*

I'm not sure i understand why this is necessary, also the other mod tests one.
Asking cause the whole file is a test file and already compiled with cfg(test)
</blockquote></details>

I felt that it would be easier to just recreate a new set of tests based on the changes I have made than to change the old ones. Therefore I want to contain these in a mod and then just delete the other mod once I am ready. When I do that I will remove this "mod" keyword.

Removed the cfg
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 634 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbO6ab4GqRBAP0Knjr:-OKbXTxGDHgc3Z2XMAiV:b6g1yi6) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L634)):*

There is quite a bit of boilerplate so I'm not sure which API is tested.
Also what is in order?
</blockquote></details>

Interesting, I feel like there is very little boilerplate here. We have maybe 3 lines, num_messages, setup, and stream_id. After that we are on to business:
1. Create a new stream to send
2. Send messages on the stream
3. Check the messages that are sent to the network
4. Close the channel and check that Fin is sent to the network.
I added comments in the test.

in order - we're testing that a single outbound stream is send to the network in order. admittedly maybe in order is an obvious property. Renamed to `outbound_single`
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 642 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbMskF8DWWpTS-M_-Q:-OKbYIUY0--wXcou5IRb:b-w2662m) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L642)):*

what message is sent here? 
More specifically, what is outbound_internal_sender?
</blockquote></details>

what message is sent here?
- This is consensus sending a new proposal stream to the stream handler. There is no external message, rather it is the receiver that will receive all of the ProposalParts for this specific proposal.

outbound_internal_sender
- This is a sender connecting the client (consensus) to the stream handler. It is used to initiate a new outbound stream. (internal because it is client to stream handler, not stream handler to network). Would you prefer `client_to_streamhandler_sender`?
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 656 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbNtgcC6P5rLti-Gcr:-OKbZ3pZ5xHS-DoUg1H1:bensh23) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L656)):*

what message is this? a close message that is sent in the previous line?
</blockquote></details>

When the client closes the sender it tells the streamhandler that this stream has completed, and so the streamhandler can send a Fin message to the network. I tried removing `make_test_message` to make this more obvious.
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 683 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbPyAl3akk56p1RDu_:-OKb_ja36UP9y6w9W7fC:b-ful71s) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L683)):*

Should this use make_test?
</blockquote></details>

No. `make_test_message` makes a StreamMessage. The client sends only the content, they don't need to consider the StreamMessage wrapper.
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 695 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbPA7O4ScpR4oa4lFi:-OKbaGn33GByeezDDiYp:b3gyn4t) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L695)):*

Not sure how this got past clippy, Lior banned `as` due to surprise truncations and saturations.
</blockquote></details>

I thought it was allowed in tests. Do you think this needs to be changed?
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 703 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbQ3cm2NDh8duA9wLK:-OKbaPLA5m4Auw3yBNjN:b-hv84t0) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L703)):*

What can cause this assert to fail?  Does this check that the stream doesn't shuffle the messages?
</blockquote></details>

Yeah it checks that all streams were sent out to the network in order - we assert the order of each stream but we do not care in what order messages between streams were sent.

giladchase (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-:-OKeYXt72RyNdTkf0GcH:bnfp4nl)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @guy-starkware)

___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbQZQl0wZVwfZyTF_V:-OKeR1Dp7OdW-ywJFgxU:b-4xgvcg):*

If we do this here then we need to do this everywhere. My point is that the existing set of tests does look at private fields and I am making an entirely different set of tests. This is not fundamentally different than what we try to do for all unit tests though.
</blockquote></details>

Yr call, though i dont think we need to do it everywhere necessarily.
However I do agree that we for black-box testing we should only exercise the API without private access. For some unit test its okay though, for example if you want to test a specific API in isolation, then private access is sometimes necessary, or when snapshot testing
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 634 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbO6ab4GqRBAP0Knjr:-OKeWoqJ3yuyGKXnKHlh:b-onh50q) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L634)):*

Interesting, I feel like there is very little boilerplate here. We have maybe 3 lines, num_messages, setup, and stream_id. After that we are on to business:
1. Create a new stream to send
2. Send messages on the stream
3. Check the messages that are sent to the network
4. Close the channel and check that Fin is sent to the network.
I added comments in the test.

in order - we're testing that a single outbound stream is send to the network in order. admittedly maybe in order is an obvious property. Renamed to `outbound_single`
</blockquote></details>

OK, though these are a lot of things to test in a single unit test, perhaps it can be split into smaller unit tests. Or otherwise if this is a "test happy flow", maybe it should be made into a full-blown flow-test.

For example, if (3) is retested in every test, that is motivation for testing it only once (in as many flavors as necessary) and subsequently hiding it in further tests through test utils or even mocks.

Also, `outbound_multiple_streams` seems quite complex (for a unit test at least). If it fails due to a bug/refactor, will there be other, smaller tests that indicate the issue? If my only indication would be, for example, an assert failure 10 asserts in, this will be time consuming to debug.

Regardless, I'm not blocking.
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 642 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbMskF8DWWpTS-M_-Q:-OKeSM2r1hoaFPgDaaMV:bc4k820) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L642)):*

what message is sent here?
- This is consensus sending a new proposal stream to the stream handler. There is no external message, rather it is the receiver that will receive all of the ProposalParts for this specific proposal.

outbound_internal_sender
- This is a sender connecting the client (consensus) to the stream handler. It is used to initiate a new outbound stream. (internal because it is client to stream handler, not stream handler to network). Would you prefer `client_to_streamhandler_sender`?
</blockquote></details>

So its a dedicated stream for intiating new streams whose message type is always `(stream_id,stream_receiver)`? 

`client_to_streamhandler_sender` is good,ya. Also if stream_id and receiver are passed together, should they have a message type like "proposalStart"?
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 683 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbPyAl3akk56p1RDu_:-OKeWDB_BdI8RUu_4CU4:b7x9wls) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L683)):*

No. `make_test_message` makes a StreamMessage. The client sends only the content, they don't need to consider the StreamMessage wrapper.
</blockquote></details>

Note that it isn't immediately obvious why this is asserted to equal make_test, which by itself isn't a descriptive test-util.
I had to doubleback a few times in order to understand this test, and read through the test utils in detail (and ideally I don't want people to have to read test utils, they should be "believable")   
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 695 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbPA7O4ScpR4oa4lFi:-OKeTPPs5ZizPlwaDQNU:bc901a4) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L695)):*

I thought it was allowed in tests. Do you think this needs to be changed?
</blockquote></details>

Nope its due to [this ignore](https://github.com/starkware-libs/sequencer/blob/f73bbb68c2d654f88e9cb35943f26f279354e3c9/crates/starknet_consensus/src/stream_handler.rs#L21).

In general I agree with Lior's call, we caught bugs when we removed it and it is indeed quite surprising in its behavior.
And I think except for when converting `float`s to ints, all trivial conversions can be equally trivial with `into`, and nontrivial ones (like u128 as u64) are better represented with `try_into`s.

matan-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4322)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @giladchase and @guy-starkware)

___
*[`-- commits` line 6 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbQZQl0wZVwfZyTF_V:-OKeat_07EYuONtyNx1Z:b2ih2z9):*

Yr call, though i dont think we need to do it everywhere necessarily.
However I do agree that we for black-box testing we should only exercise the API without private access. For some unit test its okay though, for example if you want to test a specific API in isolation, then private access is sometimes necessary, or when snapshot testing
</blockquote></details>

I agree with everything you are saying, I just had this comment since the old set of tests didn't follow this guideline. As you say this is a guideline not a hard rule.
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 634 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbO6ab4GqRBAP0Knjr:-OKebDm2DTbFF-hQ45-t:bjc8nac) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L634)):*

OK, though these are a lot of things to test in a single unit test, perhaps it can be split into smaller unit tests. Or otherwise if this is a "test happy flow", maybe it should be made into a full-blown flow-test.

For example, if (3) is retested in every test, that is motivation for testing it only once (in as many flavors as necessary) and subsequently hiding it in further tests through test utils or even mocks.

Also, `outbound_multiple_streams` seems quite complex (for a unit test at least). If it fails due to a bug/refactor, will there be other, smaller tests that indicate the issue? If my only indication would be, for example, an assert failure 10 asserts in, this will be time consuming to debug.

Regardless, I'm not blocking.
</blockquote></details>

Perhaps the multiple streams is better categorized as a property test than as a unit test? I think though that this is a core property of the stream handler, that it can handle multiple streams and each one will be sent in order. So if a refactor breaks what we check I would definitely want to know.
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 683 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbPyAl3akk56p1RDu_:-OKeinZU2-nEycTEPgNJ:bpz10x7) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L683)):*

Note that it isn't immediately obvious why this is asserted to equal make_test, which by itself isn't a descriptive test-util.
I had to doubleback a few times in order to understand this test, and read through the test utils in detail (and ideally I don't want people to have to read test utils, they should be "believable")
</blockquote></details>

I'm not sure what you mean by "believable". I did rename the helper function so it's clear that it's just building a message.
___
*[`crates/starknet_consensus/src/stream_handler_test.rs` line 695 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4322#-OKbPA7O4ScpR4oa4lFi:-OKejD7ZCWFXJ7HSQ3gh:bvjgjuw) ([raw file](https://github.com/starkware-libs/sequencer/blob/7fca518bb1e645d79e9cf48fcf68a1316f7d82ba/crates/starknet_consensus/src/stream_handler_test.rs#L695)):*

Nope its due to [this ignore](https://github.com/starkware-libs/sequencer/blob/f73bbb68c2d654f88e9cb35943f26f279354e3c9/crates/starknet_consensus/src/stream_handler.rs#L21).

In general I agree with Lior's call, we caught bugs when we removed it and it is indeed quite surprising in its behavior.
And I think except for when converting `float`s to ints, all trivial conversions can be equally trivial with `into`, and nontrivial ones (like u128 as u64) are better represented with `try_into`s.
</blockquote></details>

alright I removed all the as conversions

matan-starkware (commented):
> Reviewed 1 of 1 files at r5, 1 of 1 files at r6.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4322)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @matan-starkware)

matan-starkware (commented):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4322)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @matan-starkware)

PR #4418, opened 2025-02-24
refactor(starknet_sequencer_node): remove tokio select from server runner

AlonLStarkWare:


AlonLStarkWare:
* **#4418** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4418?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4418?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Itay-Tsabary-Starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4418)* status: 0 of 1 files reviewed, 4 unresolved discussions

___
*[`crates/starknet_sequencer_node/src/servers.rs` line 438 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-OJrOVXqDO4ow7YuHiMp:-OJrOVXqDO4ow7YuHiMq:b-6itm0x) ([raw file](https://github.com/starkware-libs/sequencer/blob/ebfcc76a2de63b43057bb954debd0dcaa43034d5/crates/starknet_sequencer_node/src/servers.rs#L438)):*
> ```XML
> async fn create_servers<ReturnType: Send + 'static>(
>     labeled_futures: Vec<(impl Future<Output = ReturnType> + Send + 'static, String)>,
> ) -> FuturesUnordered<Pin<Box<dyn Future<Output = (ReturnType, String)> + Send>>> {
> ```

IIUC this no longer has to be a generic function, as `ReturnType` is always `()`. Is that correct?



_Code quote:_
```XML
ReturnType
```
___
*[`crates/starknet_sequencer_node/src/servers.rs` line 441 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-OJrOCof-hqTigO-Ow60:-OJrOCof-hqTigO-Ow61:b8abyy2) ([raw file](https://github.com/starkware-libs/sequencer/blob/ebfcc76a2de63b43057bb954debd0dcaa43034d5/crates/starknet_sequencer_node/src/servers.rs#L441)):*
> ```XML
>     let tasks = FuturesUnordered::new();
>     for (future, label) in labeled_futures {
>         tasks.push(future.map(move |res| (res, label.clone())).boxed());
> ```

Is this necessary? Have you tried consuming the vector using `into_iter()`?



_Code quote:_
```XML
.clone()
```
___
*[`crates/starknet_sequencer_node/src/servers.rs` line 663 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-OJrOlaqD3ufYqmXX34c:-OJrOlaqD3ufYqmXX34d:b-ot4yck) ([raw file](https://github.com/starkware-libs/sequencer/blob/ebfcc76a2de63b43057bb954debd0dcaa43034d5/crates/starknet_sequencer_node/src/servers.rs#L663)):*
> ```XML
>     all_servers.extend(servers.local_servers.run().await);
>     all_servers.extend(servers.remote_servers.run().await);
>     all_servers.extend(servers.wrapper_servers.run().await);
> ```

Can this code use the `create_servers` fn as well?



_Code quote:_
```XML
    let mut all_servers = FuturesUnordered::new();
    all_servers.extend(servers.local_servers.run().await);
    all_servers.extend(servers.remote_servers.run().await);
    all_servers.extend(servers.wrapper_servers.run().await);
```
___
*[`crates/starknet_sequencer_node/src/servers.rs` line 669 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-OJrOvGxEYl8_-byMBzk:-OJrOvGxEYl8_-byMBzl:b-q35g11) ([raw file](https://github.com/starkware-libs/sequencer/blob/ebfcc76a2de63b43057bb954debd0dcaa43034d5/crates/starknet_sequencer_node/src/servers.rs#L669)):*
> ```XML
>     } else {
>         Err(anyhow::anyhow!("Failed to run component servers."))
>     }
> ```

Could you please explain this condition? It seems both paths result in a similar error.



_Code quote:_
```XML
        Err(anyhow::anyhow!("{} Servers ended unexpectedly.", servers_type))
    } else {
        Err(anyhow::anyhow!("Failed to run component servers."))
    }
```

Itay-Tsabary-Starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4418)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @AlonLStarkWare)

___
*[`crates/starknet_sequencer_node/src/servers.rs` line 663 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-OJrOlaqD3ufYqmXX34c:-OJs7GLz9UI6EiQmskby:b88qkhg) ([raw file](https://github.com/starkware-libs/sequencer/blob/ebfcc76a2de63b43057bb954debd0dcaa43034d5/crates/starknet_sequencer_node/src/servers.rs#L663)):*
<details><summary><i>Previously, AlonLStarkWare (Alon-Lukatch-Starkware) wrote…</i></summary><blockquote>

Unfortunately I couldn't find a way to create a recursive future like we discussed so I can't use create_servers here. This is the best way I could find for merging the futures.
</blockquote></details>

`servers.local_servers.run()` is a future, you can pair it with a label, e.g., "local servers"
ditto for other servers

and then plug them into `create_servers`.

Does this work? If not, then why?
___
*[`crates/starknet_sequencer_node/src/servers.rs` line 669 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-OJrOvGxEYl8_-byMBzk:-OJs4XCUEuJn4L74Vzlz:bw9up50) ([raw file](https://github.com/starkware-libs/sequencer/blob/ebfcc76a2de63b43057bb954debd0dcaa43034d5/crates/starknet_sequencer_node/src/servers.rs#L669)):*
<details><summary><i>Previously, AlonLStarkWare (Alon-Lukatch-Starkware) wrote…</i></summary><blockquote>

The first one will happen in case one of the server futures groups has returned (e.g. local servers returned).
The second is theoretically unreachable - calling next().await on a FuturesUnordered can yield None only if the struct is empty (either no futures were inserted or they are all resolved). I didn't give enough thought to this error message tbh.
I didn't want to panic here as preparation to your [graceful shutdown](https://github.com/starkware-libs/sequencer/blob/02-24-refactor_starknet_sequencer_node_remove_tokio_select_from_server_runner/crates/starknet_sequencer_node/src/main.rs#L24)
</blockquote></details>

Thanks for the clarification.
A few thoughts:
1. If a branch condition is unreachable then it should use `unreachable!`.
2. This fn doesn't have to return `anyhow::Result<()> `, it can simply have an empty return type. This is preferred imo: if a fn doesn't return a success variant then it probably shouldn't have a `result` type to begin with.
3. Graceful shutdown will require a proper design and substantial work, let's not burden ourselves with it at the moment. And, it is not mine :-)
4. What about explicit cancenlation of other running tasks? The base version had that (to some extend, with the call to `abort`).  Can we have that as well?

AlonLStarkWare (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4418)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @Itay-Tsabary-Starkware)

___
*[`crates/starknet_sequencer_node/src/servers.rs` line 663 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-OJrOlaqD3ufYqmXX34c:-OJs9lm6Adjhxh2SsRdv:blozqqk) ([raw file](https://github.com/starkware-libs/sequencer/blob/ebfcc76a2de63b43057bb954debd0dcaa43034d5/crates/starknet_sequencer_node/src/servers.rs#L663)):*

`servers.local_servers.run()` is a future, you can pair it with a label, e.g., "local servers"
ditto for other servers

and then plug them into `create_servers`.

Does this work? If not, then why?
</blockquote></details>

`FuturesUnordered` isn't a future, and even if it were, because of the generic signature of the `servers.run()` methods their opaque types could differ so the compiler does not allow creating a vector from them.
___
*[`crates/starknet_sequencer_node/src/servers.rs` line 669 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-OJrOvGxEYl8_-byMBzk:-OJsAxw30s9apRj3NYLi:b4u7he) ([raw file](https://github.com/starkware-libs/sequencer/blob/ebfcc76a2de63b43057bb954debd0dcaa43034d5/crates/starknet_sequencer_node/src/servers.rs#L669)):*

Thanks for the clarification.
A few thoughts:
1. If a branch condition is unreachable then it should use `unreachable!`.
2. This fn doesn't have to return `anyhow::Result<()> `, it can simply have an empty return type. This is preferred imo: if a fn doesn't return a success variant then it probably shouldn't have a `result` type to begin with.
3. Graceful shutdown will require a proper design and substantial work, let's not burden ourselves with it at the moment. And, it is not mine :-)
4. What about explicit cancenlation of other running tasks? The base version had that (to some extend, with the call to `abort`).  Can we have that as well?
</blockquote></details>

1. Done
2. Ok, changed it to panic in the first case and unreachable in the second
3. Understandable
4. not required with FuturesUnordered, it is taken care of during `drop`. Here is ChatGPTs response when asked about that:

Itay-Tsabary-Starkware (changes requested):
> Reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4418)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @AlonLStarkWare)

___
*[`crates/starknet_sequencer_node/src/servers.rs` line 663 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-OJrOlaqD3ufYqmXX34c:-OJtYqv07omuWS1LL3vh:b4q897v) ([raw file](https://github.com/starkware-libs/sequencer/blob/ebfcc76a2de63b43057bb954debd0dcaa43034d5/crates/starknet_sequencer_node/src/servers.rs#L663)):*
<details><summary><i>Previously, AlonLStarkWare (Alon-Lukatch-Starkware) wrote…</i></summary><blockquote>

`FuturesUnordered` isn't a future, and even if it were, because of the generic signature of the `servers.run()` methods their opaque types could differ so the compiler does not allow creating a vector from them.
</blockquote></details>

This probably requires a conversation, so in the mean time just add a todo to circle back to this.
___
*[`crates/starknet_sequencer_node/src/servers.rs` line 669 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-OJrOvGxEYl8_-byMBzk:-OJtZ6ig5E_bglS8UKoD:b-22q1vg) ([raw file](https://github.com/starkware-libs/sequencer/blob/ebfcc76a2de63b43057bb954debd0dcaa43034d5/crates/starknet_sequencer_node/src/servers.rs#L669)):*
<details><summary><i>Previously, AlonLStarkWare (Alon-Lukatch-Starkware) wrote…</i></summary><blockquote>

1. Done
2. Ok, changed it to panic in the first case and unreachable in the second
3. Understandable
4. not required with FuturesUnordered, it is taken care of during `drop`. Here is ChatGPTs response when asked about that:

</blockquote></details>

1-3: :thumbsup: 
4: Chatgpt is a liar. Tokio tasks are not cancelled on drop. The whole runtime will be shut down, which will cancel these tasks as well, but that's not the cleanest of exits. Anyhow, just add a todo for the mean time to address that.

AlonLStarkWare (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4418)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @Itay-Tsabary-Starkware)

___
*[`crates/starknet_sequencer_node/src/servers.rs` line 663 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-OJrOlaqD3ufYqmXX34c:-OJwq56O7xvzhZ_WjDMX:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/ebfcc76a2de63b43057bb954debd0dcaa43034d5/crates/starknet_sequencer_node/src/servers.rs#L663)):*

This probably requires a conversation, so in the mean time just add a todo to circle back to this.
</blockquote></details>

Done.
___
*[`crates/starknet_sequencer_node/src/servers.rs` line 669 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-OJrOvGxEYl8_-byMBzk:-OJwQCv_4SRC0rpZgJSB:bw8vana) ([raw file](https://github.com/starkware-libs/sequencer/blob/ebfcc76a2de63b43057bb954debd0dcaa43034d5/crates/starknet_sequencer_node/src/servers.rs#L669)):*

1-3: :thumbsup: 
4: Chatgpt is a liar. Tokio tasks are not cancelled on drop. The whole runtime will be shut down, which will cancel these tasks as well, but that's not the cleanest of exits. Anyhow, just add a todo for the mean time to address that.
</blockquote></details>

I tried looking into the comments on [drop implementation for FuturesUnordered](https://docs.rs/futures-util/0.3.31/src/futures_util/stream/futures_unordered/mod.rs.html#575), and from what I understand dropping the FuturesUnordered is the best way to cancel the inner tasks.
Either way I added a TODO.

Itay-Tsabary-Starkware (approved):
:lgtm:

> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4418#-:-OJya1DSE6XWzMMr5v43:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @AlonLStarkWare)

AlonLStarkWare (commented):
> Reviewed 2 of 2 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4418)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @AlonLStarkWare)

PR #4434, opened 2025-02-24
feat(starknet_mempool): remove tentative mempool state

dafnamatsry:


dafnamatsry:
* **#4434** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4434?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/4434?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4434)* status: 0 of 2 files reviewed, 2 unresolved discussions

___
*[`crates/starknet_mempool/src/mempool.rs` line 91 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2aGLQECl4yiaCpkjt:-OK2aGLQECl4yiaCpkju:b-85azwo) ([raw file](https://github.com/starkware-libs/sequencer/blob/582ae000383ca09724faa9c22da04128f4156d2b/crates/starknet_mempool/src/mempool.rs#L91)):*
> ```XML
>             .or_insert(nonce);
>         *tentative_nonce
>     }
> ```

If we're not saving the account nonces for gw txs from accounts we don't have recent info from the Batcher, this means that we are going to store txs in the mempool with no information at all about their account nonces. This will make debugging the mempool harder i think.
Specifically, if we are debugging a tx that was incorrectly queued (or less critically not queued), we don't have any readily available information for understanding why the mempool made this decision other than going over logs.



_Code quote:_
```XML
    fn get_or_insert(&mut self, address: ContractAddress, nonce: Nonce) -> Nonce {
        if let Some(staged_or_committed_nonce) =
            self.staged.get(&address).or_else(|| self.committed.get(&address)).copied()
        {
            return staged_or_committed_nonce;
        }

        let tentative_nonce = self
            .tentative
            .entry(address)
            .and_modify(|tentative_nonce| {
                if nonce > *tentative_nonce {
                    *tentative_nonce = nonce;
                }
            })
            .or_insert(nonce);
        *tentative_nonce
    }
```
___
*[`crates/starknet_mempool/src/mempool_test.rs` line 884 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2cJ-zAddp3JEmkh11:-OK2cJ-zAddp3JEmkh12:b-ldyqc0) ([raw file](https://github.com/starkware-libs/sequencer/blob/e68917762dd725ba64198ebebe3462e6cdacb834/crates/starknet_mempool/src/mempool_test.rs#L884)):*
> ```XML
> 
>     // The account has no txs in the pool, but is known through a committed block.
>     commit_block(&mut mempool, [(ACCOUNT_ADDRESS, 1)], []);
> ```

nitpicking: don't know if this convention was kept, but back in the day we tried to only test a single API per unit-test, and put all multi-api tests in `flow_test.rs`.



_Code quote:_
```XML
    add_tx(
        &mut mempool,
        &add_tx_input!(tx_hash: 1, address: ACCOUNT_ADDRESS, tx_nonce: 0, account_nonce: 0),
    );
    assert_eq!(mempool.contains_tx_from(account_address), true);

    // The account has a staged tx in the mempool.
    let get_tx_response = mempool.get_txs(1).unwrap();
    assert_eq!(get_tx_response.first().unwrap().contract_address(), account_address);
    assert_eq!(mempool.contains_tx_from(account_address), true);

    // The account has no txs in the pool, but is known through a committed block.
    commit_block(&mut mempool, [(ACCOUNT_ADDRESS, 1)], []);
```
___
*[`crates/starknet_mempool/src/mempool.rs` line 215 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2WCVREBtyvhTz5f-u:-OK2WCVREBtyvhTz5f-v:b45h) ([raw file](https://github.com/starkware-libs/sequencer/blob/e68917762dd725ba64198ebebe3462e6cdacb834/crates/starknet_mempool/src/mempool.rs#L215)):*
> ```XML
>         self.tx_pool.insert(tx)?;
> 
>         // Align to account nonce, only if it is at least the one stored.
> ```


___
*[`crates/starknet_mempool/src/mempool.rs` line 217 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2UahJDnMvN3q2VCdU:-OK2UahJDnMvN3q2VCdV:b-psti6o) ([raw file](https://github.com/starkware-libs/sequencer/blob/e68917762dd725ba64198ebebe3462e6cdacb834/crates/starknet_mempool/src/mempool.rs#L217)):*
> ```XML
>         // Align to account nonce, only if it is at least the one stored.
>         let AccountState { address, nonce: incoming_account_nonce } = account_state;
>         let account_nonce = self.state.get(address).unwrap_or(incoming_account_nonce);
> ```

I saw this also in validate_incoming_tx, does this mean that all nonce accesses in any add_tx flow has to do this as well? I'm not sure this is clear to readers why we're doing this, and if someone adds something to the flow they might not spot the fact that they must do this check.



_Code quote:_
```XML
        let account_nonce = self.state.get(address).unwrap_or(incoming_account_nonce);
```
___
*[`crates/starknet_mempool/src/mempool.rs` line 218 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2TGoy0W6OZCOxDrbI:-OK2TGoy0W6OZCOxDrbJ:bdb3ucq) ([raw file](https://github.com/starkware-libs/sequencer/blob/e68917762dd725ba64198ebebe3462e6cdacb834/crates/starknet_mempool/src/mempool.rs#L218)):*
> ```XML
>         let AccountState { address, nonce: incoming_account_nonce } = account_state;
>         let account_nonce = self.state.get(address).unwrap_or(incoming_account_nonce);
>         if tx_reference.nonce == account_nonce {
> ```

I think this misses a few flows, i'm kind bummed that our tests didn't catch this (other than the removed nonce reduction test which only catches the first one :smiling_face_with_tear:):

1.  `add_tx(tx_nonce: 3, account_nonce: 3)` then `add_tx(tx_nonce: 2, account_nonce: 2)` then `get_txs(1)` returns the tx_nonce 2 (should be 3)
2.  `add_tx(tx_nonce: 3, account_nonce: 2)` then `add_tx(tx_nonce: 4, account_nonce: 3)` then get_txs(1) return [], should be 3. 

Regardless of what happens here i'll add the missing test for 2 and a better test for 1 tomorrow.

Both of the above assume that the GW was updated and fired off the request faster than the batcher, which AFAIK is a possibility (for example due to network latency or the batcher service being stressed). That is, when we added the tentative thing we assumed that GW txs can appear out of order (due to parallel GW instances) and will always have a race with the batcher over the commit_block thing. 



_Code quote:_
```XML
        let account_nonce = self.state.get(address).unwrap_or(incoming_account_nonce);
        if tx_reference.nonce == account_nonce {
```
___
*[`crates/starknet_mempool/src/mempool.rs` line 292 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2d9he7KppYOCM-MoA:-OK2d9he7KppYOCM-MoB:b-68v70) ([raw file](https://github.com/starkware-libs/sequencer/blob/e68917762dd725ba64198ebebe3462e6cdacb834/crates/starknet_mempool/src/mempool.rs#L292)):*
> ```XML
> 
>     pub fn contains_tx_from(&self, account_address: ContractAddress) -> bool {
>         self.state.get(account_address).is_some() || self.tx_pool._contains_account(account_address)
> ```

I think the second cond is strictly more general than the first now, and they both have the same complexity.
The leading `_` can be removed if you're using it, it was just to clear the linter.



_Suggestion:_
```XML
        self.tx_pool.contains_account(account_address)
```

dafnamatsry (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4434)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @giladchase)

___
*[`crates/starknet_mempool/src/mempool.rs` line 218 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2TGoy0W6OZCOxDrbI:-OK6RdrYEuaaKO8MLj4E:b-2elltb) ([raw file](https://github.com/starkware-libs/sequencer/blob/e68917762dd725ba64198ebebe3462e6cdacb834/crates/starknet_mempool/src/mempool.rs#L218)):*

I think this misses a few flows, i'm kind bummed that our tests didn't catch this (other than the removed nonce reduction test which only catches the first one :smiling_face_with_tear:):

1.  `add_tx(tx_nonce: 3, account_nonce: 3)` then `add_tx(tx_nonce: 2, account_nonce: 2)` then `get_txs(1)` returns the tx_nonce 2 (should be 3)
2.  `add_tx(tx_nonce: 3, account_nonce: 2)` then `add_tx(tx_nonce: 4, account_nonce: 3)` then get_txs(1) return [], should be 3. 

Regardless of what happens here i'll add the missing test for 2 and a better test for 1 tomorrow.

Both of the above assume that the GW was updated and fired off the request faster than the batcher, which AFAIK is a possibility (for example due to network latency or the batcher service being stressed). That is, when we added the tentative thing we assumed that GW txs can appear out of order (due to parallel GW instances) and will always have a race with the batcher over the commit_block thing.
</blockquote></details>

1. This case means the account nonce got smaller. This can't happen. If for some reason the mempool didn't get updated on the new nonce, worst case it will get tx_nonce 2 and reject it.
2. This case will be same also before this PR. In this case we should have got a commit_block with the new nonce, which would have insert the first tx to the queue. If like you say we didn't get the commit_block, the mempool should eventually get it and catch up.

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4434)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @dafnamatsry)

___
*[`crates/starknet_mempool/src/mempool.rs` line 218 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2TGoy0W6OZCOxDrbI:-OK6WVq_D978C5YjgrRc:baqdprf) ([raw file](https://github.com/starkware-libs/sequencer/blob/e68917762dd725ba64198ebebe3462e6cdacb834/crates/starknet_mempool/src/mempool.rs#L218)):*
1. AFAIK the GWs are parallelized, can't it happen that two requests arrive in different instances so that txs arrive in opposite order? How about through p2p, how are we assured that txs will always arrive in order?

2.
>This case will be same also before this PR

Ohh nice this is actually a bug in closing nonce holes in add_tx (before this PR)! Before this PR, add_tx should fix the queue based on the retrieved account_nonce, it has all the information to do so! We only handled closing nonce holes in the case that the tx closed the hole, but not the case where the `account_nonce` closed a hole. 

>mempool should eventually get it and catch up.

Ya but we have received all the information we need in order to not backlog the transaction, why is waiting better than not waiting?  The only trade is cost, and _cost_ for not waiting is basically saving a number per account of a tx that hasn't been recently committed, and as a result we get:
a) faster delivery to get_txs (don't have to wait until commit_block if there are races)
b) a black box (MempoolState) that encapsulates all account nonce logic, without any implicit logic in the the main API. For example, if one encounters `get(address).unwrap_or(incoming_account_nonce)`, why not update the state with this account nonce? They have to know in advance that this nonce is not reliable, whereas the mempool state nonce is reliable. 

BTW, if the mempool don't save the GW account nonces anywhere and doesn't use them to close holes, then effectively this arg becomes a bit that the GW sets as `add_to_queue=true` iff the tx nonce equals the account nonce.

dafnamatsry (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4434)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @giladchase)

___
*[`crates/starknet_mempool/src/mempool.rs` line 218 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2TGoy0W6OZCOxDrbI:-OKPaNhW6Utx8CtFs-CA:b-3ouyh1) ([raw file](https://github.com/starkware-libs/sequencer/blob/e68917762dd725ba64198ebebe3462e6cdacb834/crates/starknet_mempool/src/mempool.rs#L218)):*
1. They can arrive out of order, but in your example the account nonce has changed. In a normal scenario, the mempool should already be aware of the new state. If for some reason it didn't, and the GW is ahead of the mempool, the worst case would be either:
- Mempool will send an old tx to the batcher, which will reject it.
- Eligible txs will wait in the mempool, until it will get updated.

2. I don't see how keeping the tentative nonce would help. You would still have all the information you need in `add_tx` to close the holes.

> a black box (MempoolState) that encapsulates all account nonce logic, without any implicit logic in the the main API. For example, if one encounters `get(address).unwrap_or(incoming_account_nonce)`, why not update the state with this account nonce?


The `incoming_account_state` can be an input to the MempoolState, so this logic will be encapsulated there. 

> BTW, if the mempool don't save the GW account nonces anywhere and doesn't use them to close holes, then effectively this arg becomes a bit that the GW sets as `add_to_queue=true` iff the tx nonce equals the account nonce.

Well, this is effectively the situation now.

dafnamatsry (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4434)* status: 0 of 3 files reviewed, 1 unresolved discussion (waiting on @alonh5, @ayeletstarkware, and @giladchase)

___
*[`crates/starknet_mempool/src/mempool.rs` line 217 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2UahJDnMvN3q2VCdU:-OKeIWae9adfOGF3LvIw:b-kw1q63) ([raw file](https://github.com/starkware-libs/sequencer/blob/e68917762dd725ba64198ebebe3462e6cdacb834/crates/starknet_mempool/src/mempool.rs#L217)):*

I saw this also in validate_incoming_tx, does this mean that all nonce accesses in any add_tx flow has to do this as well? I'm not sure this is clear to readers why we're doing this, and if someone adds something to the flow they might not spot the fact that they must do this check.
</blockquote></details>

Moved this logic to be inside the MempoolState.
___
*[`crates/starknet_mempool/src/mempool.rs` line 292 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2d9he7KppYOCM-MoA:-OKeJ4vI23-hXD3_v5zb:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/e68917762dd725ba64198ebebe3462e6cdacb834/crates/starknet_mempool/src/mempool.rs#L292)):*

I think the second cond is strictly more general than the first now, and they both have the same complexity.
The leading `_` can be removed if you're using it, it was just to clear the linter.
</blockquote></details>

Done.
___
*[`crates/starknet_mempool/src/mempool_test.rs` line 884 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2cJ-zAddp3JEmkh11:-OKeJ6CJAGL1_gRnJ72g:bbn8vtn) ([raw file](https://github.com/starkware-libs/sequencer/blob/e68917762dd725ba64198ebebe3462e6cdacb834/crates/starknet_mempool/src/mempool_test.rs#L884)):*

nitpicking: don't know if this convention was kept, but back in the day we tried to only test a single API per unit-test, and put all multi-api tests in `flow_test.rs`.
</blockquote></details>

I don't think we kept this convention... 
___
*[`crates/starknet_mempool/src/mempool.rs` line 215 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OK2WCVREBtyvhTz5f-u:-OKeITv52pMH0Y9ySjW8:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/e68917762dd725ba64198ebebe3462e6cdacb834/crates/starknet_mempool/src/mempool.rs#L215)):*
> ```XML
>         self.tx_pool.insert(tx)?;
> 
>         // Align to account nonce, only if it is at least the one stored.
> ```

Done.

dafnamatsry (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4434)* status: all files reviewed, 1 unresolved discussion (waiting on @ayeletstarkware)

___
*[`crates/starknet_mempool/src/mempool.rs` line 316 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/4434#-OKzGu3HByQnVjkK8YGr:-OKzP2Dt4JGaWbdKoj4A:b-jibehm) ([raw file](https://github.com/starkware-libs/sequencer/blob/0a18c2a429210267785edab219f792bbd732fcda/crates/starknet_mempool/src/mempool.rs#L316)):*

Rename in separate PR?
</blockquote></details>

Will do in a separate PR (should be changed also in the communication layer).

PR #4627, opened 2025-03-03
fix(starknet_l1_provider): fix initialize high finality flow

giladchase:
This fixes a scenario that will only happne in tests in practice, in which the given finality is higher than number of blocks on L1.

We return an error since this is assumed to be a configuration error by the user; test code will have to align with this behavior.

### DIFF
--- a/crates/starknet_l1_provider/src/l1_scraper.rs
+++ b/crates/starknet_l1_provider/src/l1_scraper.rs
@@ -74,9 +74,7 @@ impl<B: BaseLayerContract + Send + Sync> L1Scraper<B> {
 
     #[instrument(skip(self), err)]
     pub async fn initialize(&mut self) -> L1ScraperResult<(), B> {
-        let Some((latest_l1_block, events)) = self.fetch_events().await? else {
-            return Ok(());
-        };
+        let (latest_l1_block, events) = self.fetch_events().await?;
 
         // If this gets too high, send in batches.
         let initialize_result = self.l1_provider_client.initialize(events).await;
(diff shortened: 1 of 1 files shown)


alonh5 (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4627)* status: 0 of 1 files reviewed, 1 unresolved discussion

___
*[`crates/starknet_l1_provider/src/l1_scraper.rs` line 77 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4627#-OKVSJ-Y679F6zOP6tAB:-OKVSJ-Y679F6zOP6tAC:br3an5b) ([raw file](https://github.com/starkware-libs/sequencer/blob/7d132f469f9fcdc153a5332e22ad6eeabd3d59a2/crates/starknet_l1_provider/src/l1_scraper.rs#L77)):*
> ```XML
>     #[instrument(skip(self), err)]
>     pub async fn initialize(&mut self) -> L1ScraperResult<(), B> {
>         let Some((latest_l1_block, events)) = self.fetch_events().await? else {
> ```

Wouldn't it be better if fetch_events returned this error? returning none in this case is wrong because it's an error. And fetch_events is indeed called elsewhere and we can miss that too.

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4627)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @alonh5)

___
*[`crates/starknet_l1_provider/src/l1_scraper.rs` line 77 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/4627#-OKVSJ-Y679F6zOP6tAB:-OKVzqsE2458i0qd5NsM:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/7d132f469f9fcdc153a5332e22ad6eeabd3d59a2/crates/starknet_l1_provider/src/l1_scraper.rs#L77)):*

Wouldn't it be better if fetch_events returned this error? returning none in this case is wrong because it's an error. And fetch_events is indeed called elsewhere and we can miss that too.
</blockquote></details>

Done.

PR #4639, opened 2025-03-04
fix(starknet_class_manager): do not `rename` class files

elintul:


giladchase (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/4639#-:-OKVXlPE4BsFodPkHalm:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @yair-starkware)

___
*[`-- commits` line 11 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/4639#-OKVXMZ3Af8tJf6m2Wqo:-OKVXMZ3Af8tJf6m2Wqp:bi9estx):*
:pray: 



_Code quote:_
```
  We have an atomic marker (in a separate `mdbx` table), representing class existence.
  If a partial write occures, class will be considered nonexistent, which
  is okay.
  This is a quick fix for the following linking error we saw inside the
  pod: `error=Invalid cross-device link (os error 18)`.
  To prevent partial writes, which might make the DB unclean, I'll restore
  the temporary-write-and-rename but on the same mount as the persistent
  directory, and will check the linking error doesn't happen.
```

PR #4682, opened 2025-03-05
Make all `ArchivalDataResources` fields public

cptartur:
Made all `ArchivalDataResources` fields public. We need to create this struct directly in Starknet Foundry.


cptartur:
@dorimedini-starkware Closing this PR, turns out these changes are already on 0.13.5 branch

PR #5109, opened 2025-03-20
feat(starknet_consensus_orchestrator): Make the proposal building abort the work if the proposal is interrupted.

guyf-starkware:
Also:

- Added unit test
- Removed outdated to do
- Add "." where missing


guyf-starkware:
Test

PR #5221, opened 2025-03-24
docs(starknet_mempool): add and fix some comments

guyf-starkware:


dafnamatsry (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/5221)* status: all files reviewed, 2 unresolved discussions (waiting on @guyf-starkware)

___
*[`crates/starknet_mempool/src/mempool.rs` line 111 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/5221#-OM7J7388FvX_cJgp0Sd:-OM7J7388FvX_cJgp0Se:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/68d1225ebd38159bebd920ded07414d1f1369bb5/crates/starknet_mempool/src/mempool.rs#L111)):*
> ```XML
> 
>     /// Retuns the addresses which need to be rewinded (i.e. addressed which were staged but did not
>     /// make to the commit).
> ```





_Suggestion:_
```XML
    /// Updates the committed nonces, and returns the addresses which need to be rewinded (i.e. addressed which were staged but did not
    /// make to the commit).
```
___
*[`crates/starknet_mempool/src/mempool.rs` line 319 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/5221#-OMC7KDiEb6i0GOPIUJ5:-OMC7KDiEb6i0GOPIUJ6:b-7vtld2) ([raw file](https://github.com/starkware-libs/sequencer/blob/68d1225ebd38159bebd920ded07414d1f1369bb5/crates/starknet_mempool/src/mempool.rs#L319)):*
> ```XML
>             // 2. A newer nonce: Not possible since account_nonce is >= the account tx in the queue.
>             // 3. A nonce equal to the tx nonce, in which case the tx would have already been
>             //    removed in handle_fee_escalation.
> ```

In most cases the queue won't hold any transaction from this account.

Maybe comment something like:
```
// Remove queued transactions the account might have. This includes old nonce transactions that have 
// become obsolete; those with an equal nonce should already have been removed in 
// `handle_fee_escalation`.
```



_Code quote:_
```XML
            // Three options, the queue holds a tx with:
            // 1. An older nonce: It's no longer relevant since a newer tx has since been written to
            //    a block so we want to remove it.
            // 2. A newer nonce: Not possible since account_nonce is >= the account tx in the queue.
            // 3. A nonce equal to the tx nonce, in which case the tx would have already been
            //    removed in handle_fee_escalation.
```

guyf-starkware:
Made a mess of git. Recreated using graphite. Closing this pr

PR #5260, opened 2025-03-25
chore(blockifier): simplify v1-bound Cairo 1 check

Yoni-Starkware:


liorgold2 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/5260)* status: all files reviewed, 2 unresolved discussions

___
*[`crates/blockifier/src/execution/syscalls/hint_processor.rs` line 640 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/5260#-OMIf8Gv8_nNb8LjpFTG:-OMIf8Gv8_nNb8LjpFTH:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/7c78f864806b062177f6f7fdbcb8707ed3cb6b8c/crates/blockifier/src/execution/syscalls/hint_processor.rs#L640)):*
> ```XML
>         // Note: the returned version might not be equal to the actual transaction version.
>         // Also, the returned version is a property of the current VM execution (of some contract
>         // call) so it is okay to allocate it once whithout re-checking the version in every
> ```





_Suggestion:_
```XML
        // Also, the returned version is a property of the current entry-point execution,
        // so it is okay to allocate it once without re-checking the version in every
```
___
*[`crates/blockifier/src/execution/syscalls/hint_processor.rs` line 640 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/5260#-OMIfHFL9KDgDR99xpsB:-OMIfHFL9KDgDR99xpsC:blfgl3o) ([raw file](https://github.com/starkware-libs/sequencer/blob/7c78f864806b062177f6f7fdbcb8707ed3cb6b8c/crates/blockifier/src/execution/syscalls/hint_processor.rs#L640)):*
> ```XML
>         // Note: the returned version might not be equal to the actual transaction version.
>         // Also, the returned version is a property of the current VM execution (of some contract
>         // call) so it is okay to allocate it once whithout re-checking the version in every
> ```

I wonder if this comment should be here or next to the cache (`self.execution_info_ptr`) implementation.
In the code in this function you don't allocate it once.



_Code quote:_
```XML
        // call) so it is okay to allocate it once whithout re-checking the version in every
```

PR #5456, opened 2025-03-31
chore(apollo_integration_tests): increase mmap file params for tests

matanl-starkware:


matanl-starkware:
Decided to drop this PR, since the default values were good enough for tests that are not so heavy in performance.

PR #5909, opened 2025-04-16
feat(starknet_patricia): impl subtraction for SubTreeHeight

rotem-starkware:


dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/5909)* status: all files reviewed, 1 unresolved discussion (waiting on @nimrod-starkware)

___
*[`crates/starknet_patricia/src/patricia_merkle_tree/types.rs` line 47 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/5909#-ONyn1baBEpej7jjr7X2:-ONyn1baBEpej7jjr7X3:bc4gp23) ([raw file](https://github.com/starkware-libs/sequencer/blob/898d9891acb34720f1d0ec3aa8429452b1d5ef16/crates/starknet_patricia/src/patricia_merkle_tree/types.rs#L47)):*
> ```XML
>         Self::new(self.0 - rhs)
>     }
> }
> ```

this can underflow, right?
1. if this is a test util, then ok, but add a `#[cfg]` for it
2. if not, I would implement `checked_sub`, not `sub`, and force the caller to handle potential underflow



_Code quote:_
```XML
impl Sub<u8> for SubTreeHeight {
    type Output = Self;

    fn sub(self, rhs: u8) -> Self::Output {
        Self::new(self.0 - rhs)
    }
}
```

PR #6122, opened 2025-04-24
fix(starknet_api): allow serialization of chain id with leading zeros

nimrod-starkware:


nimrod-starkware:
* **#6122** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/6122?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/6122?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6122)* status: all files reviewed, 1 unresolved discussion (waiting on @nimrod-starkware)

___
*[`crates/starknet_api/src/core.rs` line 65 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6122#-OObTUuTD0OiNRbeVCn6:-OObTUuTD0OiNRbeVCn7:b-4qmngw) ([raw file](https://github.com/starkware-libs/sequencer/blob/bc7137f564d068770910348ba5ddeb8a4159d463/crates/starknet_api/src/core.rs#L65)):*
> ```XML
>     fn from(s: String) -> Self {
>         let s_ref: &str = s.as_ref();
>         // Allow leading null bytes.
> ```

why was this needed? where did we get null bytes and why?



_Code quote:_
```XML
// Allow leading null bytes.
```

nimrod-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6122)* status: all files reviewed, 1 unresolved discussion (waiting on @dorimedini-starkware)

___
*[`crates/starknet_api/src/core.rs` line 65 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6122#-OObTUuTD0OiNRbeVCn6:-OObUP155chV1Aa8_grl:bnh6tw5) ([raw file](https://github.com/starkware-libs/sequencer/blob/bc7137f564d068770910348ba5ddeb8a4159d463/crates/starknet_api/src/core.rs#L65)):*

why was this needed? where did we get null bytes and why?
</blockquote></details>

PTAL [here](https://github.com/starkware-industries/starkware/pull/37780):
The value of the python enum uses `to_bytes` method which pads with zeros.
Then when we deserialize we get something like in the test.
Now when I think of it maybe we should remove the zeros in python and do nothing here

dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6122)* status: all files reviewed, 1 unresolved discussion (waiting on @nimrod-starkware)

___
*[`crates/starknet_api/src/core.rs` line 65 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6122#-OObTUuTD0OiNRbeVCn6:-OObVK_U6_siQh1X8FfL:b2rh7os) ([raw file](https://github.com/starkware-libs/sequencer/blob/bc7137f564d068770910348ba5ddeb8a4159d463/crates/starknet_api/src/core.rs#L65)):*

PTAL [here](https://github.com/starkware-industries/starkware/pull/37780):
The value of the python enum uses `to_bytes` method which pads with zeros.
Then when we deserialize we get something like in the test.
Now when I think of it maybe we should remove the zeros in python and do nothing here
</blockquote></details>

how does the blockifier pass the chain IDs to the native blockifier?
why isn't that an issue there?

in any case, since the current method can (in theory) cause collisions, i suggest the python side serializes the chain ID to an integer, and we deserialize the integer to the chain ID on the rust side. WDYT?

PR #6194, opened 2025-04-27
test(apollo_network): changed tests to not rely on one bootstrap node

sirandreww-starkware:

PR #6136, opened 2025-04-24
refactor(apollo_network): moved bootstrap behaviour to prepare for multiple peers

sirandreww-starkware:

PR #6196, opened 2025-04-27
chore(apollo_network): added mechanism to wake up bootstrap and kad requesting behaviours


sirandreww-starkware:


sirandreww-starkware:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/6196?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#6197** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/6197?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6196** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/6196?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/6196?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#6195** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/6195?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6194** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/6194?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#6136** <a href="https://app.graphite.dev/github/pr/starkware-libs/sequencer/6136?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

ShahakShama (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-:-OPy40Ap3IJQi0YnrkcQ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sirandreww-starkware)

ShahakShama:
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 9 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OPy5n_I6nTgOumJodKw:-OPy5n_I6nTgOumJodKx:b8nw2dv) ([raw file](https://github.com/starkware-libs/sequencer/blob/bb475b95da15736abb358f03e8429cfbe6b18a5e/crates/apollo_network/src/discovery/behaviours/mod.rs#L9)):*
> ```XML
> 
> /// Function that sets up the waker of the context to wake up at a specific instant.
> pub fn configure_context_to_wake_at_instant(waker: Waker, instant: Instant) {
> ```

Rename the function accordingly

ShahakShama (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: all files reviewed, 1 unresolved discussion (waiting on @sirandreww-starkware)

___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 9 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OPy5n_I6nTgOumJodKw:-OPy5q-11DsNZC-ly8Go:b-z35nwk) ([raw file](https://github.com/starkware-libs/sequencer/blob/bb475b95da15736abb358f03e8429cfbe6b18a5e/crates/apollo_network/src/discovery/behaviours/mod.rs#L9)):*

Rename the function accordingly
</blockquote></details>

wake_waker_at_instant

sirandreww-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: all files reviewed, 1 unresolved discussion (waiting on @ShahakShama)

___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 9 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OPy5n_I6nTgOumJodKw:-OPyVc6s0OKKNNmCm7B9:b-3x0exf) ([raw file](https://github.com/starkware-libs/sequencer/blob/bb475b95da15736abb358f03e8429cfbe6b18a5e/crates/apollo_network/src/discovery/behaviours/mod.rs#L9)):*

wake_waker_at_instant
</blockquote></details>

This function changes in a future PR

ShahakShama (changes requested):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: all files reviewed, 9 unresolved discussions (waiting on @sirandreww-starkware)

___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 12 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ25V8UEdSaY0Zco3ZD:-OQ25V8UEdSaY0Zco3ZE:b-x5yu01) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L12)):*
> ```XML
> 
> #[derive(Debug, Default)]
> pub struct TimeWakerManager {
> ```

IMO you can remove the Time here, but I'll leave it up to you
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 12 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ25_fpBVbVK5m6rjvm:-OQ25_fpBVbVK5m6rjvn:b-3o3g0a) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L12)):*
> ```XML
> 
> #[derive(Debug, Default)]
> pub struct TimeWakerManager {
> ```

This can be private. AFAIK you can use private structs inside the same module. If I'm wrong it can be pub(self) or pub(crate)
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 22 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ24i_e-OkHsWZHcsvK:-OQ24i_f60AXUcI_4Z3U:b-3cy3ja) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L22)):*
> ```XML
>     /// * Should likely be called at the start of a `poll` function
>     /// * **Aborts previous wake timers**
>     pub fn set_waker(&mut self, waker: Waker) {
> ```

Change this to add_waker and add it to a vector of wakers.
There might be a case where we poll the behaviour from 2 different places and then you'll override the waker from the other place's polling
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 25 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ272LO021uLnDqxBgk:-OQ272LO021uLnDqxBgl:bsptpx2) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L25)):*
> ```XML
>         self.current_waker = Some(waker);
> 
>         for handle in self.join_handles.iter_mut() {
> ```

When you accept my request above you can remove this code
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 38 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ27nju1s385YCYbhJZ:-OQ27nju1s385YCYbhJ_:benm1we) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L38)):*
> ```XML
> 
>     /// Spawns a task that will wake the waker at a specific instant
>     pub fn wake_at(&mut self, instant: Instant) {
> ```

wake_at and wake_now should clear the waker (set it to None or clear the vector if you change it to vector)
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 54 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ25HGe2XcPrgroO6VM:-OQ25HGe2XcPrgroO6VN:b-86n1ut) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L54)):*
> ```XML
> 
>     /// calls wake on the waker
>     pub fn wake_now(&mut self) {
> ```

wake_now -> wake
It's pretty obvious that it will wake now
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 58 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ259344I15afHcOB0s:-OQ259344I15afHcOB0t:bfjdycz) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L58)):*
> ```XML
>             waker.wake_by_ref();
>         } else {
>             // This should never happen
> ```

why? I think it's a valid case
___
*[`crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs` line 34 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ27DMI47GhoDpz6iRF:-OQ27DMI47GhoDpz6iRG:bdenffb) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs#L34)):*
> ```XML
>     bootstrap_dial_retry_strategy: ExponentialBackoff,
>     time_for_next_bootstrap_dial: Instant,
>     waker_manager: TimeWakerManager,
> ```

Like I said in previous PR, Split this to 2 waker managers:
1. waker_manager_dial_attempt
2. waker_manager_connected
And add the waker only if it's in the relevant scenario

sirandreww-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: 0 of 3 files reviewed, 8 unresolved discussions (waiting on @ShahakShama)

___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 12 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ25V8UEdSaY0Zco3ZD:-OQ2Kr3r2MMHOIAnmCDz:b-5qjxer) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L12)):*

IMO you can remove the Time here, but I'll leave it up to you
</blockquote></details>

There are many ways to use wakers I could imagine someone making a waker manager for a specific event. Thus this manager is the time based manager.
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 12 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ25_fpBVbVK5m6rjvm:-OQ2QmkJ0AGuuIHCmNQ5:b9y6c4k) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L12)):*

This can be private. AFAIK you can use private structs inside the same module. If I'm wrong it can be pub(self) or pub(crate)
</blockquote></details>

I thought someone from the might want to use this in another behaviour in the future, should I make it pub(crate)?
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 22 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ24i_e-OkHsWZHcsvK:-OQ2KmGc2RftdycOh-JW:b9dha8y) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L22)):*

Change this to add_waker and add it to a vector of wakers.
There might be a case where we poll the behaviour from 2 different places and then you'll override the waker from the other place's polling
</blockquote></details>

I see, you're right
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 25 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ272LO021uLnDqxBgk:-OQ2REKRDJlzdkpvsbGp:bv3v6g6) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L25)):*

When you accept my request above you can remove this code
</blockquote></details>

I still need to go over the handles and clear what was there, where would be a correct time to do so?
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 38 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ27nju1s385YCYbhJZ:-OQ2KbwK5rzEZVrD7EOm:bjbq4hp) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L38)):*

wake_at and wake_now should clear the waker (set it to None or clear the vector if you change it to vector)
</blockquote></details>

Why? What if an event occurs where we want to wake in 10 seconds then immediately after another even occurs where we want to wake in 5 seconds? The second wake will now never work....
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 54 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ25HGe2XcPrgroO6VM:-OQ2SYMe-ai-PDKRqXKC:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L54)):*

wake_now -> wake
It's pretty obvious that it will wake now
</blockquote></details>

Done.
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 58 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ259344I15afHcOB0s:-OQ2RmT9BUrrKrKV6slR:bo954qi) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L58)):*

why? I think it's a valid case
</blockquote></details>

This function should not be called in a case where the waker does not exist and have the caller not know about it. This might lead to a bug if one is not careful. I'll return a result and have the caller deal with it.
___
*[`crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs` line 34 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ27DMI47GhoDpz6iRF:-OQ2S9JtCQAagQ97YjKL:blbsnn8) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs#L34)):*

Like I said in previous PR, Split this to 2 waker managers:
1. waker_manager_dial_attempt
2. waker_manager_connected
And add the waker only if it's in the relevant scenario
</blockquote></details>

Why? we have an object that manages waking according to time, essentially an alarm clock. Why would we need 2 depending on the state of the stream?

ShahakShama (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: all files reviewed, 7 unresolved discussions (waiting on @sirandreww-starkware)

___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 12 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ25_fpBVbVK5m6rjvm:-OQ7L3wI8acoj1_EQE-C:b1w5kdg) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L12)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

I thought someone from the might want to use this in another behaviour in the future, should I make it pub(crate)?
</blockquote></details>

Yes. When that someone in the future comes then they'll change this struct to pub (and possibly move it to a different place)
If you can make it private that's best
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 25 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ272LO021uLnDqxBgk:-OQ7LNZX2iXFdUvDLrZu:b9dw6vq) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L25)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

I still need to go over the handles and clear what was there, where would be a correct time to do so?
</blockquote></details>

When you wake them
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 38 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ27nju1s385YCYbhJZ:-OQ7LVgO9xr-4E-V5sd7:bwd5v40) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L38)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

Why? What if an event occurs where we want to wake in 10 seconds then immediately after another even occurs where we want to wake in 5 seconds? The second wake will now never work....
</blockquote></details>

When you wake, then the polling happens again and the polling should re-register the waker if needed
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 58 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ259344I15afHcOB0s:-OQ7Lhoa38Ykg0YVEU2C:b-skbfgj) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L58)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

This function should not be called in a case where the waker does not exist and have the caller not know about it. This might lead to a bug if one is not careful. I'll return a result and have the caller deal with it.
</blockquote></details>

Why? If for example I'm awaking every time there's a connection established and it happens to be that no one is waiting for it then I'll receive an error even though it's a flow that will happen when our node runs
___
*[`crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs` line 34 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ27DMI47GhoDpz6iRF:-OQ7LyuX-ThwSyRLGA69:b-snpsdt) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs#L34)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

Why? we have an object that manages waking according to time, essentially an alarm clock. Why would we need 2 depending on the state of the stream?
</blockquote></details>

It's not only based on time, it's also based on events, and I want that when you're registering a waker you're aware to what you're registering it to
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 22 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ7LJeO0-VvcV_Hkbln:-OQ7LJeO0-VvcV_Hkblo:bda8m9c) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L22)):*
> ```XML
>     /// * Should likely be called at the start of a `poll` function
>     /// * **Aborts previous wake timers**
>     pub fn set_waker(&mut self, waker: Waker) {
> ```

Document the return value
___
*[`crates/apollo_network/src/discovery/behaviours/kad_requesting.rs` line 75 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ7MAeHAwFD4PVQLXyG:-OQ7MAeHAwFD4PVQLXyH:b-2s4tjq) ([raw file](https://github.com/starkware-libs/sequencer/blob/e84774513929711aea970c3d88f8d218711615df/crates/apollo_network/src/discovery/behaviours/kad_requesting.rs#L75)):*
> ```XML
>         }
> 
>         self.waker_manager.wake_at(self.time_for_next_kad_query).unwrap();
> ```

We use unwrap only in test code. In source code we use expect

sirandreww-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: all files reviewed, 6 unresolved discussions (waiting on @ShahakShama)

___
*[`crates/apollo_network/src/discovery/behaviours/kad_requesting.rs` line 75 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ7MAeHAwFD4PVQLXyG:-OQ86Cav95fQknjBgPpW:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/11c19c3bef1743106c1183470f11ea598ede153b/crates/apollo_network/src/discovery/behaviours/kad_requesting.rs#L75)):*

We use unwrap only in test code. In source code we use expect
</blockquote></details>

Done.
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 12 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ25_fpBVbVK5m6rjvm:-OQ7ahX60cSLPo3oSdOz:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L12)):*

Yes. When that someone in the future comes then they'll change this struct to pub (and possibly move it to a different place)
If you can make it private that's best
</blockquote></details>

Done.
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 22 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ7LJeO0-VvcV_Hkbln:-OQ7af6VDvDlVe-gszrW:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L22)):*

Document the return value
</blockquote></details>

Done.
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 25 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ272LO021uLnDqxBgk:-OQ86AYBC7Af6TFgOjib:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L25)):*

When you wake them
</blockquote></details>

Done.
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 38 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ27nju1s385YCYbhJZ:-OQ84xgp5otlDJJwb_-p:b-g3odtg) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L38)):*

When you wake, then the polling happens again and the polling should re-register the waker if needed
</blockquote></details>

I ended up going with sharing the wakers vector approach, that way multiple wakers are held until the next time they are woken up then once the call to wake them up is made we remove them and they will be added again in the next poll if needed.
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 58 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ259344I15afHcOB0s:-OQ84nR0EbOrgq0qbuZv:b-kegew8) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/mod.rs#L58)):*

Why? If for example I'm awaking every time there's a connection established and it happens to be that no one is waiting for it then I'll receive an error even though it's a flow that will happen when our node runs
</blockquote></details>

I left it up to the user to decide if this is something they want to check for or not...
___
*[`crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs` line 34 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQ27DMI47GhoDpz6iRF:-OQ85F-w2BJpC95kDyZf:b8rxg32) ([raw file](https://github.com/starkware-libs/sequencer/blob/f6fbae6e0948c9de2373b4f9b45f421d687cb846/crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs#L34)):*

It's not only based on time, it's also based on events, and I want that when you're registering a waker you're aware to what you're registering it to
</blockquote></details>

Why? how would this even be possible? Again, TimeWakerManager is an alarm clock, use it however you like, why do you need 2? When you wake up (get polled), you check your state and then you make a decision in accordance to your state, what would the added clock be used for?

ShahakShama (changes requested):
> Reviewed 1 of 3 files at r5, 6 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: all files reviewed, 4 unresolved discussions (waiting on @guyf-starkware)

___
*[`crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs` line 29 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIyOg04tiGX8IPR717:-OQIyOg04tiGX8IPR718:b-l6qjgy) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs#L29)):*
> ```XML
>         // add the future to the list of timers
>         self.timers.push(future);
>         let _ = self.timers.poll_next_unpin(cx);
> ```

This means you only poll timers when you call wake_at. Maybe have a separate impl Future for this and then you'll poll the time waker manager inside the poll function? You can use the Never type for the return value of the future
___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 1 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIfjjQ3sUa6WgFX6kx:-OQIfjjQ3sUa6WgFX6ky:bjbume3) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L1)):*
> ```XML
> use std::task::Waker;
> ```

We could use this utility in sqmr behaviour. add a TODO to extract to a util file and use it in sqmr too
Can be done in a future PR
___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 22 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIfxcK0ozcORnpba6J:-OQIfxcK0ozcORnpba6K:bhm1g3p) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L22)):*
> ```XML
>     /// calls wake on the wakers and clears the list.
>     /// Returns Ok(()) if there were wakers to wake, Err(()) if there were none.
>     pub fn wake(&mut self) -> Result<(), ()> {
> ```

The return value should be bool
___
*[`crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs` line 35 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIyphM2wdZhjjl2h1B:-OQIyphM2wdZhjjl2h1C:bcsqgww) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs#L35)):*
> ```XML
>     time_for_next_bootstrap_dial: Instant,
>     time_waker: TimeWakerManager,
>     event_waker: EventWakerManager,
> ```

I prefer to split this to upon_connection_established_waker and upon_connection_closed/dropped_waker

guyf-starkware (changes requested):
> Reviewed 5 of 6 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: all files reviewed, 14 unresolved discussions (waiting on @ShahakShama and @sirandreww-starkware)

___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 22 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIfxcK0ozcORnpba6J:-OQIzKUxET6solJ19TZN:b-qylgbi) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L22)):*

The return value should be bool
</blockquote></details>

Why? what would be the meaning of a return value of false?
___
*[`crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs` line 8 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ-xC3BO5IMKGfhGV0:-OQJ-xC410fdJ2W9fBVT:bujy5to) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs#L8)):*
> ```XML
> use tokio::time::Instant;
> 
> /// A manager for scheduling contexts to wake at specific times.
> ```

/// A manager which wakes a waker at a specified time



_Code quote:_
```XML
/// A manager for scheduling contexts to wake at specific times.
```
___
*[`crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs` line 15 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ0adR5GLlPU_2a-uv:-OQJ0adR5GLlPU_2a-uw:b-k88xqa) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs#L15)):*
> ```XML
> 
> impl TimeWakerManager {
>     /// Will wake the context at a specific instant.
> ```

I don't think a comment is needed here since the name gives the same info. If you still prefer to have comment I suggest: "Wakes the context..."
___
*[`crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs` line 22 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ2QaL7DXnH6Ge1hxV:-OQJ2QaL7DXnH6Ge1hxW:b-kqon3e) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs#L22)):*
> ```XML
>             waker.wake();
>         }
>         .boxed();
> ```

what am i missing here, why do we need to store the wakers and poll them, why aren't we spawing a thread for this? why do we need all the code below this line?
___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 1 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIzcoq5-_NXGtbK9WM:-OQIzcoq5-_NXGtbK9WN:bp5zg7i) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L1)):*
> ```XML
> use std::task::Waker;
> ```

Is there a strong reason to have a separate file for each manager? given the small amount of code and how much they have in common maybe they should live in a single file (waker_managers.rs) ?
___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 1 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIzoZZ6V_278eJ0Zk7:-OQIzoZZ6V_278eJ0Zk8:b955io9) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L1)):*
> ```XML
> use std::task::Waker;
> ```

I personally believe we should have a test for everything. This class' test would be very simple but i think it's still worth doing. What do you think?
___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 3 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIyjVZDn1yy_x8l47m:-OQIyjVZDn1yy_x8l47n:b-4xi9p3) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L3)):*
> ```XML
> use std::task::Waker;
> 
> /// A manager for handling wakers and waking them for some events.
> ```

A manager which handles waking multiple wakers.



_Code quote:_
```XML
/// A manager for handling wakers and waking them for some events.
```
___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 20 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIzZsE-CtlrYy-Pi5V:-OQIzZsE-CtlrYy-Pi5W:bbzl646) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L20)):*
> ```XML
>     }
> 
>     /// calls wake on the wakers and clears the list.
> ```

/// Calls wake on the wakers and clears the list.



_Code quote:_
```XML
/// calls wake on the wakers and clears the list.
```
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 17 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQISCkx8bc8vElDIFBo:-OQISCkx8bc8vElDIFBp:b-6i139r) ([raw file](https://github.com/starkware-libs/sequencer/blob/d27816dc9768e53f44a47870d4e6d55dfd29f5b2/crates/apollo_network/src/discovery/behaviours/mod.rs#L17)):*
> ```XML
> }
> 
> impl TimeWakerManager {
> ```

add a comment about what this struct does. All structs should have comments describing what they are meant for.
I think once this is understood it will be easier to understand what the best API should be.
___
*[`crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs` line 76 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ2uq95xPNuC_xLkWM:-OQJ2uq95xPNuC_xLkWN:bej1hqz) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs#L76)):*
> ```XML
>                     .expect("Dial sleep strategy ended even though it's an infinite iterator.");
>                 self.time_for_next_bootstrap_dial = now + delta_duration;
>                 self.event_waker.wake().unwrap();
> ```

unwrap -> expect ? (same below)
___
*[`crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs` line 120 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ3mPN9T17I0fQP3hh:-OQJ3mPN9T17I0fQP3hi:brmt52a) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs#L120)):*
> ```XML
>     {
>         let now = tokio::time::Instant::now();
>         self.event_waker.add_waker(cx.waker());
> ```

You're adding a waker at the start which you only need in case you don't reach the poll pending at the end
This means you might have wakers which you'll try to wake even though they god a Ready response.

Should this class be directly before the return Poll::Pending ?

ShahakShama:
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 22 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIfxcK0ozcORnpba6J:-OQJALEQ1WgCcj124NIi:b-26pfaa) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L22)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

Why? what would be the meaning of a return value of false?
</blockquote></details>

That no wakers were waked. My main thing is that it's usually not an error and if the user wants it can make it an error
We can also get rid of the return value entirely

guyf-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: all files reviewed, 15 unresolved discussions (waiting on @ShahakShama and @sirandreww-starkware)

___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 22 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIfxcK0ozcORnpba6J:-OQJAhF3BcmnSd5vsQlZ:b4c9lyr) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L22)):*

That no wakers were waked. My main thing is that it's usually not an error and if the user wants it can make it an error
We can also get rid of the return value entirely
</blockquote></details>

:thumbsup:

sirandreww-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: all files reviewed, 12 unresolved discussions (waiting on @guyf-starkware and @ShahakShama)

___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 1 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIzcoq5-_NXGtbK9WM:-OQJ6Obz5jJwM5egPfdU:b-aeu50q) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L1)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

Is there a strong reason to have a separate file for each manager? given the small amount of code and how much they have in common maybe they should live in a single file (waker_managers.rs) ?
</blockquote></details>

I prefer small files to big ones, this stems from my conviction that merging files is easier than splitting them. So whenever I can I try to separate different structs to different files.
___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 1 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIzoZZ6V_278eJ0Zk7:-OQJ6gE5-hLkhGqXFda9:bygj99s) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L1)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

I personally believe we should have a test for everything. This class' test would be very simple but i think it's still worth doing. What do you think?
</blockquote></details>

I'll add a TODO to add this in the future
___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 1 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIfjjQ3sUa6WgFX6kx:-OQJ6vzD57N46KwT-uaT:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L1)):*

We could use this utility in sqmr behaviour. add a TODO to extract to a util file and use it in sqmr too
Can be done in a future PR
</blockquote></details>

Done
___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 3 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIyjVZDn1yy_x8l47m:-OQJ6uf51x1YeOToWdiE:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L3)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

A manager which handles waking multiple wakers.
</blockquote></details>

Done
___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 20 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIzZsE-CtlrYy-Pi5V:-OQJ71HtC_nZuT_9NSoR:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L20)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

/// Calls wake on the wakers and clears the list.
</blockquote></details>

Done
___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 22 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIfxcK0ozcORnpba6J:-OQJ75ZG9s8-ZUB5nIyz:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L22)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

Why? what would be the meaning of a return value of false?
</blockquote></details>

Done
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 17 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQISCkx8bc8vElDIFBo:-OQJ7Z7012s8jWasggh5:bsvtmvq) ([raw file](https://github.com/starkware-libs/sequencer/blob/d27816dc9768e53f44a47870d4e6d55dfd29f5b2/crates/apollo_network/src/discovery/behaviours/mod.rs#L17)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

add a comment about what this struct does. All structs should have comments describing what they are meant for.
I think once this is understood it will be easier to understand what the best API should be.
</blockquote></details>

These comments are placed immediately above the struct
___
*[`crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs` line 8 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ-xC3BO5IMKGfhGV0:-OQJ7oqn62yiNkxXXLvP:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs#L8)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

/// A manager which wakes a waker at a specified time
</blockquote></details>

Done
___
*[`crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs` line 15 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ0adR5GLlPU_2a-uv:-OQJ89Iu3oOq4BTsnpew:b-wfgw76) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs#L15)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

I don't think a comment is needed here since the name gives the same info. If you still prefer to have comment I suggest: "Wakes the context..."
</blockquote></details>

Now a new note is added that makes this important
___
*[`crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs` line 22 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ2QaL7DXnH6Ge1hxV:-OQJ8HEjEcczIjKurS_w:b-vpyggb) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs#L22)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

what am i missing here, why do we need to store the wakers and poll them, why aren't we spawing a thread for this? why do we need all the code below this line?
</blockquote></details>

Spawning a task for this future is error prone, it creates weird situations where calling this function appears to "not register". 
This is because the tokio runtime does not give you guarantees about  spawned tasks. If you create a future and poll it, that future is guaranteed to save a waker and wake when it can progress, essentially assuring that cases where tokio did not spawn the task in time do not occur.
___
*[`crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs` line 29 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIyOg04tiGX8IPR717:-OQJ9fuIBURpptwjPIIa:b-223i0g) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs#L29)):*

This means you only poll timers when you call wake_at. Maybe have a separate impl Future for this and then you'll poll the time waker manager inside the poll function? You can use the Never type for the return value of the future
</blockquote></details>

Good catch!!! Done.
___
*[`crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs` line 35 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIyphM2wdZhjjl2h1B:-OQJAVeUCJcnEPSGCxUU:b1zfk5v) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs#L35)):*

I prefer to split this to upon_connection_established_waker and upon_connection_closed/dropped_waker
</blockquote></details>

Done, but I don't see the value in this. I'm probably missing something
___
*[`crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs` line 76 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ2uq95xPNuC_xLkWM:-OQJAbbSCOOWZUdBqpOW:b-rh3x28) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs#L76)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

unwrap -> expect ? (same below)
</blockquote></details>

no longer relevant, thank you though
___
*[`crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs` line 120 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ3mPN9T17I0fQP3hh:-OQJENk2CRF5s7hGCdWM:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs#L120)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

You're adding a waker at the start which you only need in case you don't reach the poll pending at the end
This means you might have wakers which you'll try to wake even though they god a Ready response.

Should this class be directly before the return Poll::Pending ?
</blockquote></details>

Done
___
*[`crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs` line 150 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ4RmuEv2WW767S_wS:-OQJEXzE4d8WTs5HzGHV:bmcr02e) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs#L150)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

since if i understand correctly we had a bug here before this change, shouldn't we add a test for capturing that bug?
</blockquote></details>

This is added in future PRs

ShahakShama (approved):
> Reviewed 4 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: all files reviewed, 9 unresolved discussions (waiting on @guyf-starkware)

ShahakShama (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-:-OQJFRm33cnY1RuhQFrG:bnfp4nl)* status: all files reviewed, 9 unresolved discussions (waiting on @guyf-starkware)

guyf-starkware (dismissed):
> Reviewed 1 of 4 files at r8.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: all files reviewed, 5 unresolved discussions (waiting on @sirandreww-starkware)

___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 1 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQIzcoq5-_NXGtbK9WM:-OQJF3c99qE2i2YcnrHL:bbnecsx) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L1)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

I prefer small files to big ones, this stems from my conviction that merging files is easier than splitting them. So whenever I can I try to separate different structs to different files.
</blockquote></details>

There is a cost to a lot of small files. It does create "pollution" but this is a judgement call so not blocking on this, your choice.
___
*[`crates/apollo_network/src/discovery/behaviours/mod.rs` line 17 at r6](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQISCkx8bc8vElDIFBo:-OQJH9om9G0sQNy_XIlt:b-1tp06h) ([raw file](https://github.com/starkware-libs/sequencer/blob/d27816dc9768e53f44a47870d4e6d55dfd29f5b2/crates/apollo_network/src/discovery/behaviours/mod.rs#L17)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

These comments are placed immediately above the struct
</blockquote></details>

Sorry this was a comment on an old snapshot. Resolving
___
*[`crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs` line 22 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ2QaL7DXnH6Ge1hxV:-OQJIE8e7Dm-C3Qa5bnB:bjlczio) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs#L22)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

Spawning a task for this future is error prone, it creates weird situations where calling this function appears to "not register". 
This is because the tokio runtime does not give you guarantees about  spawned tasks. If you create a future and poll it, that future is guaranteed to save a waker and wake when it can progress, essentially assuring that cases where tokio did not spawn the task in time do not occur.
</blockquote></details>

Are you sure about this? If spawning tasks in tokio does not guarantee that the spawned code it called i know of many places where we would have big issues.
The reason I keep asking here is that if we can do it using tokio spawn it will make this code and the API so much simpler and not require us to have an explicit poll requirement
___
*[`crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs` line 150 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJ4RmuEv2WW767S_wS:-OQJGIZQApqU_8IP8Eji:bnvblu6) ([raw file](https://github.com/starkware-libs/sequencer/blob/49f0bd1ae7158e1bd78c7941a1aa84c0242af09d/crates/apollo_network/src/discovery/behaviours/bootstrapping/bootstrap_peer.rs#L150)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

This is added in future PRs
</blockquote></details>

I'm assuming the PR already exists if so great, if not consider a TODO
___
*[`crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs` line 18 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJFuR7BrJIF3i3i1mD:-OQJFuR7BrJIF3i3i1mE:b1j01wc) ([raw file](https://github.com/starkware-libs/sequencer/blob/31670560bd1a8cf71b46ae624a54b6597593ec63/crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs#L18)):*
> ```XML
> impl TimeWakerManager {
>     /// Will wake the context at a specific instant.
>     /// This function is must be followed by polling this manager in every call to poll.
> ```

1. is must -> must
2. Not sure that this comment is saying. I think what you're saying is "You must call poll follwing one or more calls to this function" ?

I'm still trying to figure out how
___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 23 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJFVQuBAQbDQFJQPJr:-OQJFVQv5R19VJNn1Tgf:b4qpel3) ([raw file](https://github.com/starkware-libs/sequencer/blob/31670560bd1a8cf71b46ae624a54b6597593ec63/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L23)):*
> ```XML
> 
>     /// Calls wake on the wakers and clears the list.
>     /// Returns true if there were wakers to wake, false if there were none.
> ```

you forgot to return any value from the function... right now it doesn't return anything

sirandreww-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: all files reviewed, 5 unresolved discussions (waiting on @guyf-starkware)

___
*[`crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs` line 23 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJFVQuBAQbDQFJQPJr:-OQJNVbq-lscmdHkQTuu:b-2wm5s2) ([raw file](https://github.com/starkware-libs/sequencer/blob/31670560bd1a8cf71b46ae624a54b6597593ec63/crates/apollo_network/src/discovery/behaviours/event_waker_manager.rs#L23)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

you forgot to return any value from the function... right now it doesn't return anything
</blockquote></details>

This was shahak's suggestion
___
*[`crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs` line 18 at r8](https://reviewable.io/reviews/starkware-libs/sequencer/6196#-OQJFuR7BrJIF3i3i1mD:-OQJPKGPAhYtBXKDaz6s:bz64y4o) ([raw file](https://github.com/starkware-libs/sequencer/blob/31670560bd1a8cf71b46ae624a54b6597593ec63/crates/apollo_network/src/discovery/behaviours/time_waker_manager.rs#L18)):*
<details><summary><i>Previously, guyf-starkware wrote…</i></summary><blockquote>

1. is must -> must
2. Not sure that this comment is saying. I think what you're saying is "You must call poll follwing one or more calls to this function" ?

I'm still trying to figure out how
</blockquote></details>

Changed the comment to the struct

sirandreww-starkware (commented):
> Dismissed @guyf-starkware from 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: 4 of 6 files reviewed, all discussions resolved (waiting on @ShahakShama)

sirandreww-starkware (commented):
> Reviewed 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6196)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @sirandreww-starkware)

PR #6489, opened 2025-05-13
feat(blockifier): use dashmap in scheduler.rs

liorgold2:
**Stack**:
- #6519
- #6490
- #6489 ⬅

⚠️ *Part of a stack created by [spr](https://github.com/ejoffe/spr). Do not merge manually using the UI - doing so may have unexpected results.*

PR #6490, opened 2025-05-13
feat(blockifier): make Scheduler mostly independent of chunk_size

liorgold2:
Remove most occurences of chunk_size from Scheduler, as preparation to support tx stream instead of constant sized chunk.

---

**Stack**:
- #6519
- #6490 ⬅
- #6489

### DIFF
--- a/crates/blockifier/src/concurrency/flow_test.rs
+++ b/crates/blockifier/src/concurrency/flow_test.rs
@@ -76,7 +76,6 @@ fn scheduler_flow_test(
                             Task::AskForTask
                         }
                     }
-                    Task::NoTaskAvailable => Task::AskForTask,
                     Task::AskForTask => scheduler.next_task(),
                     Task::Done => break,
                 }
--- a/crates/blockifier/src/concurrency/scheduler.rs
+++ b/crates/blockifier/src/concurrency/scheduler.rs
@@ -1,4 +1,3 @@
-use std::cmp::min;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Mutex, MutexGuard, TryLockError};
 
@@ -31,10 +30,6 @@ impl<'a> TransactionCommitter<'a> {
         if self.scheduler.done() {
             return None;
         };
-        assert!(
-            *self.commit_index_guard < self.scheduler.chunk_size,
-            "The commit index must be less than the chunk size, since the scheduler is not done."
-        );
 
         let mut status = self.scheduler.lock_tx_status(*self.commit_index_guard);
         if *status != TransactionStatus::Executed {
@@ -75,7 +70,7 @@ impl Scheduler {
     pub fn new(chunk_size: usize) -> Scheduler {
         Scheduler {
             execution_index: AtomicUsize::new(0),
-            validation_index: AtomicUsize::new(chunk_size),
+            validation_index: AtomicUsize::new(usize::MAX),
             commit_index: Mutex::new(0),
             chunk_size,
             tx_statuses: DashMap::new(),
@@ -90,10 +85,6 @@ impl Scheduler {
 
         let index_to_validate = self.validation_index.load(Ordering::Acquire);
         let index_to_execute = self.execution_index.load(Ordering::Acquire);
-
-        if min(index_to_validate, index_to_execute) >= self.chunk_size {
-            return Task::NoTaskAvailable;
-        }
 
         if index_to_validate < index_to_execute {
             if let Some(tx_index) = self.next_version_to_validate() {
@@ -114,6 +105,21 @@ impl Scheduler {
     pub fn finish_execution(&self, tx_index: TxIndex) {
         self.set_executed_status(tx_index);
         self.decrease_validation_index(tx_index);
+    }
+
+    pub fn mark_missing_task(&self, tx_index: TxIndex) {
+        let mut status = self.lock_tx_status(tx_index);
+        assert_eq!(
+            *status,
+            TransactionStatus::Executing,
+            "Only executing transactions can be marked as missing. Transaction {tx_index} is not \
+             executing. Transaction status: {:?}.",
+            *status
+        );
+        *status = TransactionStatus::ReadyToExecute;
+
+        // Move the execution index back to the missing transaction.
+        self.decrease_execution_index(tx_index);
     }
 
     pub fn try_validation_abort(&self, tx_index: TxIndex) -> bool {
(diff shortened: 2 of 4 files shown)


Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6490)* status: 0 of 4 files reviewed, 1 unresolved discussion

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 121 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6490#-OQCDIBx5t09IR50-lkP:-OQCDIBx5t09IR50-lkQ:b-tfl5gy) ([raw file](https://github.com/starkware-libs/sequencer/blob/0ec57758cf778b6376786f427729e6dc42cacab7/crates/blockifier/src/concurrency/scheduler.rs#L121)):*
> ```XML
>         );
>         *status = TransactionStatus::ReadyToExecute;
>         // TODO: Should we move the execution index back?
> ```

Yes, otherwise you'll never meet a tx that was added after the end of the initial chunk (and you'll blow the dashmap up)
Need to think how to do that. `-1` might not work



_Code quote:_
```XML
        // TODO: Should we move the execution index back?
```

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 111 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6490#-OQETyef6lj38GnZJQGH:-OQETyef6lj38GnZJQGI:b1miwvk) ([raw file](https://github.com/starkware-libs/sequencer/blob/b6942880e9b21394d9c1882fafbf716230c6a773/crates/blockifier/src/concurrency/scheduler.rs#L111)):*
> ```XML
>     }
> 
>     pub fn mark_missing_task(&self, tx_index: TxIndex) {
> ```

Instead of doing this, why not add a function that updates the chunk_size when new transactions arrive?



_Code quote:_
```XML
    pub fn mark_missing_task(&self, tx_index: TxIndex) {
```

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 111 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6490#-OQETyef6lj38GnZJQGH:-OQHTVHS78ass7AII8RT:bxprh9r) ([raw file](https://github.com/starkware-libs/sequencer/blob/b6942880e9b21394d9c1882fafbf716230c6a773/crates/blockifier/src/concurrency/scheduler.rs#L111)):*
<details><summary><i>Previously, liorgold2 wrote…</i></summary><blockquote>

This is also what @Yoni-Starkware suggested. I prefer that the scheduler is not aware of chunk_size at all.
</blockquote></details>

This change offloads some of the scheduling and task management from the Scheduler to the WorkerExecutor. The original design is: 
- Scheduler manages the tx statuses and decides what tasks to assign.
- Workers do the tasks and report back the results.

With this change, the scheduler isn't actually aware of the statuses of all transactions, and the workers need to constantly correct the scheduler when it assigns tasks that don't align with actual transaction statuses.

For example, think of a block with 1 transaction that takes an entire second to execute. One worker will execute the transaction while the rest of the workers will constantly query the scheduler, adding dashmap entries and moving around the execution index and validation index. 28 workers can introduce ~3600 new dashmap entries in the span of 1ms (I did a simple test - I can send you the code), so the dashmap can also explode very fast (I know the current logic makes it stop at 28 because of waiting - just a sanity check).

Alternatively, updating the chunk size at the scheduler will result in exactly the same logic that your change does, while keeping the management of transaction statuses inside the Scheduler module. In my opinion, this makes the behavior of the scheduler more predictable and makes it easier to handle new edge cases or bugs we might run into.

liorgold2 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6490)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @avi-starkware, @noaov1, and @Yoni-Starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 111 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6490#-OQETyef6lj38GnZJQGH:-OQHary5AZiGN69mrHQy:bnns42h) ([raw file](https://github.com/starkware-libs/sequencer/blob/b6942880e9b21394d9c1882fafbf716230c6a773/crates/blockifier/src/concurrency/scheduler.rs#L111)):*
> so the dashmap can also explode very fast

Are you sure? I don't think this can happen, as when each of the 28 workers "finish" a non-existing job, they'll call `self.decrease_execution_index(tx_index);` so they'll move around the next fake 28 tasks, but won't create tasks after that.

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 111 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6490#-OQETyef6lj38GnZJQGH:-OQHhlQU4sONHs_tAMF7:b-lba1g2) ([raw file](https://github.com/starkware-libs/sequencer/blob/b6942880e9b21394d9c1882fafbf716230c6a773/crates/blockifier/src/concurrency/scheduler.rs#L111)):*
> (I know the current logic makes it stop at 28 because of waiting - just a sanity check)
It will stop at 28. I gave the 3600 number to indicate how fast it can explode if there is some edge case we miss (or if such an edge case arises due to future changes).

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 111 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6490#-OQETyef6lj38GnZJQGH:-OQHjMRb5a8EkuSjILcE:b-n0ea5d) ([raw file](https://github.com/starkware-libs/sequencer/blob/b6942880e9b21394d9c1882fafbf716230c6a773/crates/blockifier/src/concurrency/scheduler.rs#L111)):*
<details><summary><i>Previously, avi-starkware (Avi Cohen) wrote…</i></summary><blockquote>

> (I know the current logic makes it stop at 28 because of waiting - just a sanity check)
It will stop at 28. I gave the 3600 number to indicate how fast it can explode if there is some edge case we miss (or if such an edge case arises due to future changes).
</blockquote></details>

(I messed up the last message... Only the first line is the quote.)

Yoni-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6490)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @avi-starkware and @noaov1)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 111 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6490#-OQETyef6lj38GnZJQGH:-OQHjnctD36geczbEWgt:b-x25idh) ([raw file](https://github.com/starkware-libs/sequencer/blob/b6942880e9b21394d9c1882fafbf716230c6a773/crates/blockifier/src/concurrency/scheduler.rs#L111)):*
<details><summary><i>Previously, avi-starkware (Avi Cohen) wrote…</i></summary><blockquote>

(I messed up the last message... Only the first line is the quote.)
</blockquote></details>

For me, it also makes sense to update the scheduler upon a new task instead of correcting it over and over again. But I don't see a problem in the current suggestion, so non-blocking

noaov1 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6490)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @avi-starkware)

___
*[`crates/blockifier/src/concurrency/scheduler.rs` line 42 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/6490#-OQHlynP3C70nOPjXuBy:-OQHlynP3C70nOPjXuBz:brruv5u) ([raw file](https://github.com/starkware-libs/sequencer/blob/09c54c6c3940fed2dd6d19a06fa1780addd91539/crates/blockifier/src/concurrency/scheduler.rs#L42)):*
> ```XML
>         if *self.commit_index_guard == self.scheduler.chunk_size {
>             self.scheduler.done_marker.store(true, Ordering::Release);
>         }
> ```

In the new functionality, when do we want to be "done"?



_Code quote:_
```XML
        if *self.commit_index_guard == self.scheduler.chunk_size {
            self.scheduler.done_marker.store(true, Ordering::Release);
        }
```

avi-starkware:
*[`crates/blockifier/src/concurrency/scheduler.rs` line 111 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6490#-OQETyef6lj38GnZJQGH:-OQI4ODA6d_qqVV8rL2O:b7ngao) ([raw file](https://github.com/starkware-libs/sequencer/blob/b6942880e9b21394d9c1882fafbf716230c6a773/crates/blockifier/src/concurrency/scheduler.rs#L111)):*

For me, it also makes sense to update the scheduler upon a new task instead of correcting it over and over again. But I don't see a problem in the current suggestion, so non-blocking
</blockquote></details>

We disscused the issue offline
It is mostly a matter of style, so also non-blocking
I do still prefer the solution where we update the chunk size though

PR #6840, opened 2025-05-28
apollo_l1_gas_price: only accept gas price info after initializing provider

guy-starkware:


asmaastarkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6840)* status: all files reviewed, 2 unresolved discussions (waiting on @guy-starkware)

___
*[`crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs` line 106 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6840#-OSJzkwA5HyELps1OTig:-OSJzkwB7poaLdLHtwJC:byxd3wp) ([raw file](https://github.com/starkware-libs/sequencer/blob/c8df1fc6e1a089fff9aa46fad4d9875e74413fbb/crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs#L106)):*
> ```XML
>         Self {
>             config,
>             price_samples_by_block: RingBuffer::new(storage_limit),
> ```

we may save it as an optional None, and when initializing we really create it
___
*[`crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs` line 111 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/6840#-OSK-J8l6n9JBbLjniwu:-OSK-J8mE19O0Vji2130:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/c8df1fc6e1a089fff9aa46fad4d9875e74413fbb/crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs#L111)):*
> ```XML
>     }
> 
>     pub fn initialize(&mut self) -> L1GasPriceProviderResult<()> {
> ```





_Suggestion:_
```XML
pub fn initialize(&mut self) {
```

matan-starkware (changes requested):
> Reviewed 5 of 6 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6840)* status: all files reviewed, 3 unresolved discussions (waiting on @asmaastarkware)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/6840#-OSOrNmE1ICZbYnt9jWs:-OSOrNmE1ICZbYnt9jWt:b-v7a937) (no related file):*
So initialize means that an Err will be returned to the batcher if the provider was no initialized yet in response to get_price_info. The batcher already has to handle errors, so no changes on its side.

If both the scraper and provider start up together, the scraper will call initialize when it begins and all good.

If the scraper restarts but the provider does not, it will call initialize and and this will cause the provider to reset all its data. Makes sense.

If the provider restarts, it will then return an error in response to add_price_info, which will cause the scraper to exit. We then assume that the scraper will get restarted by the infra. I'm fine with this, but maybe add a comment if you can think of a good place (idk where is right).

guy-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6840)* status: all files reviewed, 3 unresolved discussions (waiting on @asmaastarkware and @matan-starkware)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/6840#-OSOrNmE1ICZbYnt9jWs:-OSOuwALBhmWqevW1aGk:blhz07n) (no related file):*

So initialize means that an Err will be returned to the batcher if the provider was no initialized yet in response to get_price_info. The batcher already has to handle errors, so no changes on its side.

If both the scraper and provider start up together, the scraper will call initialize when it begins and all good.

If the scraper restarts but the provider does not, it will call initialize and and this will cause the provider to reset all its data. Makes sense.

If the provider restarts, it will then return an error in response to add_price_info, which will cause the scraper to exit. We then assume that the scraper will get restarted by the infra. I'm fine with this, but maybe add a comment if you can think of a good place (idk where is right).
</blockquote></details>

I added a comment in the `add_price_info` right before the not-initialized error. I think that's where people will look if they are thinking about this.

matan-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6840)* status: all files reviewed, 3 unresolved discussions (waiting on @asmaastarkware)

___
*[`crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs` line 114 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/6840#-OSOyv9k4i4glM13CTDx:-OSOyv9k4i4glM13CTDy:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/3d2f52020e390e366ec6917131adc1b8d0bd353b/crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs#L114)):*
> ```XML
>         // In case the provider has been restarted while the scraper is still running,
>         // a NotInitializedError will be returned to scraper. We assume infrastructure will
>         // then cause the scraper to restart, then it will call `initialize`.
> ```





_Suggestion:_
```XML
        // In case the provider has been restarted while the scraper is still running,
        // a NotInitializedError will be returned to the scraper. We expect the scraper to exit with an
        // error, and that infrastructure will restart it, leading to initialization.
```

matan-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6840)* status: all files reviewed, 2 unresolved discussions (waiting on @asmaastarkware)

PR #6871, opened 2025-05-29
refactor(blockifier): add WorkerPoolConfig

liorgold2:
**Stack**:
- #6916
- #6907
- #6906
- #6872
- #6871 ⬅

### DIFF
--- a/crates/apollo_batcher/src/batcher.rs
+++ b/crates/apollo_batcher/src/batcher.rs
@@ -768,10 +768,7 @@ pub fn create_batcher(
         .expect("Failed to open batcher's storage");
 
     let execute_config = &config.block_builder_config.execute_config;
-    let worker_pool = Arc::new(WorkerPool::start(
-        execute_config.stack_size,
-        execute_config.concurrency_config.clone(),
-    ));
+    let worker_pool = Arc::new(WorkerPool::start(&execute_config.get_worker_pool_config()));
     let block_builder_factory = Box::new(BlockBuilderFactory {
         block_builder_config: config.block_builder_config.clone(),
         storage_reader: storage_reader.clone(),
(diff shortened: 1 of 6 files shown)


liorgold2 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6871)* status: all files reviewed, 2 unresolved discussions (waiting on @Itay-Tsabary-Starkware)

___
*[`crates/apollo_batcher/src/batcher.rs` line 771 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6871#-ORQhYr24IJdS61mJkf6:-ORR9nmF0X93egbJ48i4:bxlddp) ([raw file](https://github.com/starkware-libs/sequencer/blob/cfb0d8ba97a98122f42c2f7926bd308b93b5528e/crates/apollo_batcher/src/batcher.rs#L771)):*

Do you need to retain ownership of the config? If not, consider passing it as is, i.e., not as a reference.
</blockquote></details>

I think it's better to get a reference if you don't actually need ownership of the object - which is the case of `WorkerPool::start`.
___
*[`crates/blockifier/src/blockifier/config.rs` line 32 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/6871#-ORQhkRMBVPAbkOcB99K:-ORR9xzV7q17s8ao2D8S:b-grzlmz) ([raw file](https://github.com/starkware-libs/sequencer/blob/cfb0d8ba97a98122f42c2f7926bd308b93b5528e/crates/blockifier/src/blockifier/config.rs#L32)):*

WDYT about implementing `From<...>` instead?
</blockquote></details>

`From`/`Into` should not be used lightly. They are applicable where the translation is really obvious, where one item can really be considered of another type (e.g., `u8` -> `u16`, an error enum to an error enum containing it).
For example, in this case, we ignore the `enable` flag of the original config.

PR #6884, opened 2025-05-29
feat(apollo_mempool): add eviction tracker, tracks txs

ayeletstarkware:


alonh5 (changes requested):
> Reviewed 1 of 6 files at r1, 4 of 4 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/6884)* status: 5 of 6 files reviewed, 2 unresolved discussions

___
*[`crates/apollo_mempool/src/eviction_tracker.rs` line 7 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/6884#-OSK31NSBdsBO9SUHAUW:-OSK31NSBdsBO9SUHAUX:b-ltvqda) ([raw file](https://github.com/starkware-libs/sequencer/blob/2c36976c8d59c6f43ab2680b8389d542adacb479/crates/apollo_mempool/src/eviction_tracker.rs#L7)):*
> ```XML
> /// Suggests accounts to evict transactions from when the Mempool reaches capacity.
> #[derive(Debug, Default, PartialEq, Eq)]
> pub struct EvictionTracker {
> ```

What is the role of this struct? With the current implementation it isn't much different than a HashSet, and just adds indirection.
Do you have additional plans for this struct? If so add the methods here (even if you leave them unimplemented).
If not consider using a HashSet instead 
___
*[`crates/apollo_mempool/src/eviction_tracker.rs` line 10 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/6884#-OSK1X_K7gz64kC3GKrP:-OSK1X_L9XsSj6p99bTN:undefined) ([raw file](https://github.com/starkware-libs/sequencer/blob/2c36976c8d59c6f43ab2680b8389d542adacb479/crates/apollo_mempool/src/eviction_tracker.rs#L10)):*
> ```XML
>     // Tracks accounts with a nonce gap: the Mempool has higher-nonce transactions but lacks one
>     // with the account’s current nonce. NOTE: Delayed declares don’t count until they are
>     // inserted.
> ```





_Suggestion:_
```XML
    // with the account’s current nonce.
    // NOTE: Delayed declares don’t count until they are inserted.
```

PR #7003, opened 2025-06-05
chore(apollo_l1_provider): panic on double commit

giladchase:
Unlikely that the batcher can do anything with "double commit" error
anyway besides panicking as well.

### DIFF
--- a/crates/apollo_l1_provider/src/l1_provider_tests.rs
+++ b/crates/apollo_l1_provider/src/l1_provider_tests.rs
@@ -489,3 +489,14 @@ fn add_new_transaction_not_added_if_rejected() {
     l1_provider.add_events(vec![l1_handler_event(rejected_tx_id)]).unwrap();
     expected_l1_provider.assert_eq(&l1_provider);
 }
+
+#[test]
+#[should_panic(expected = "committed twice")]
+fn commit_block_twice_panics() {
+    // Setup.
+    let mut l1_provider =
+        L1ProviderContentBuilder::new().with_committed([l1_handler(1)]).build_into_l1_provider();
+
+    // Test.
+    l1_provider.commit_block([tx_hash!(1)].into(), [].into(), BlockNumber(0)).unwrap();
+}
--- a/crates/apollo_l1_provider/src/test_utils.rs
+++ b/crates/apollo_l1_provider/src/test_utils.rs
@@ -210,7 +210,9 @@ impl TransactionManagerContentBuilder {
         committed_hashes: impl IntoIterator<Item = TransactionHash>,
     ) -> Self {
         self.committed.get_or_insert_default().extend(
-            committed_hashes.into_iter().map(|tx_hash| (tx_hash, TransactionPayload::HashOnly)),
+            committed_hashes
+                .into_iter()
+                .map(|tx_hash| (tx_hash, TransactionPayload::HashOnly(tx_hash))),
         );
         self
     }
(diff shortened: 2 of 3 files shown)


elintul (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7003)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

___
*[`crates/apollo_l1_provider/src/transaction_manager.rs` line 213 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/7003#-OS5s_1jCqPphNRPi4xF:-OS5s_1jCqPphNRPi4xG:bnnj3sc) ([raw file](https://github.com/starkware-libs/sequencer/blob/99fab3f05bc5eebd47b2fc1d0f55060780ec3041/crates/apollo_l1_provider/src/transaction_manager.rs#L213)):*
> ```XML
>         // Can't return error because committing only part of a block leaves the provider in an
>         // undetermined state.
>         assert!(
> ```

Can you share some wisdom about the situations where using `panic` in prod. code is legit? Unreachables/sanity checks/cases where the caller will 100% need to `panic` as well? :mage_man: 



_Code quote:_
```XML
assert!
```

elintul (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7003#-:-OS5t-7t8GPtJnOdTUXa:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

elintul (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7003)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

elintul (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7003)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7003)* status: all files reviewed, 1 unresolved discussion (waiting on @elintul)

___
*[`crates/apollo_l1_provider/src/transaction_manager.rs` line 213 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/7003#-OS5s_1jCqPphNRPi4xF:-OSDkYxl1RK6TGrT6hOA:b-4uluel) ([raw file](https://github.com/starkware-libs/sequencer/blob/99fab3f05bc5eebd47b2fc1d0f55060780ec3041/crates/apollo_l1_provider/src/transaction_manager.rs#L213)):*

Can you share some wisdom about the situations where using `panic` in prod. code is legit? Unreachables/sanity checks/cases where the caller will 100% need to `panic` as well? :mage_man:
</blockquote></details>

For example, if the batcher has a bug in its internal tx cache, and forgets to remove an l1 handler transaction from the cache after committing it, and then tries to commit it again.
This is a major bug and should crash everything. Returning an error isn't cool due to comment above.

Note: in this specific case we could probably also have just checked the state, but we want to be consistent and in other cases we must check the metadata fields rather than just the state, so to be consist we still check these fields rather than just the state.

elintul (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7003)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

PR #7051, opened 2025-06-08
chore(apollo_infra): add remote client metrics

Itay-Tsabary-Starkware:


Itay-Tsabary-Starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7051)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @matanl-starkware)

___
*[`crates/apollo_infra/src/metrics.rs` line 82 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/7051#-OSEvW7wBL02IJgwipJb:-OSFRey5Al7RfvV-ZN5X:b-g4mvfd) ([raw file](https://github.com/starkware-libs/sequencer/blob/c984b0dcd75e914bb2a196b14f93841619483ee0/crates/apollo_infra/src/metrics.rs#L82)):*

I don't understand why a simple "record" is not enough.
We're talking about very small numbers, no way we "lose" information.
</blockquote></details>

Because `usize` (the inteded recorded value) is not automatically converted to `f64` (the supported recorded value)

PR #7062, opened 2025-06-08
fix(apollo_l1_provider): wrap records with dedicated type

giladchase:
Soon will be moved out of file, thus preventing access of records
not through `Records#with_record()`.

### DIFF
--- a/crates/apollo_l1_provider/src/test_utils.rs
+++ b/crates/apollo_l1_provider/src/test_utils.rs
@@ -195,7 +195,7 @@ impl From<TransactionManagerContent> for TransactionManager {
             .filter_map(|(&tx_hash, record)| record.is_proposable().then_some(tx_hash))
             .collect();
         let current_epoch = StagingEpoch::new();
-        TransactionManager::create_for_testing(records, proposable_index, current_epoch)
+        TransactionManager::create_for_testing(records.into(), proposable_index, current_epoch)
     }
 }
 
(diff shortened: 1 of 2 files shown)


giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7062)* status: all files reviewed, 1 unresolved discussion (waiting on @elintul)

___
*[`crates/apollo_l1_provider/src/transaction_manager.rs` line 431 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/7062#-OSIvmJS9ijkACRDhiVt:-OSJVypV1DNRSEdxm9ZW:bvz9el1) ([raw file](https://github.com/starkware-libs/sequencer/blob/1ac7b415cb4523fcf118eb5129f0d50a346f163b/crates/apollo_l1_provider/src/transaction_manager.rs#L431)):*

Is the convention of `with_X` methods returning the entry?
</blockquote></details>

Good point, they typically return `self`, changed to get_mut_unchecked, where the unchecked is to warn people from using directly not through the tx_manager's with_record.
(BTW I can't completely prevent access to records through `get_mut_unchecked` inside tx_manager, but this goes a long way at helping the next dev that has to mess with this)

elintul (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7062)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

PR #7247, opened 2025-06-12
ci: increase gatekeeper timeout to 1 hour

giladchase:
Not sure why this action is necessary, but its 2100 timeout is strictly smaller than our longest test.


elintul (approved):
:lgtm:

PR #7209, opened 2025-06-11
apollo_batcher: pre-send txs to validators

liorgold2:

PR #7265, opened 2025-06-13
apollo_batcher: support failed txs in validator

liorgold2:
**Stack**:
- #7266
- #7265 ⬅
- #7209

### DIFF
--- a/crates/apollo_batcher/src/block_builder.rs
+++ b/crates/apollo_batcher/src/block_builder.rs
@@ -250,13 +250,13 @@ impl BlockBuilder {
 
             self.handle_executed_txs().await?;
 
-            if lock_executor(&self.executor).is_done() {
+            // Check if the block is full. This is only relevant in propose mode.
+            // In validate mode, this is ignored and we simply wait for the proposer to send the
+            // final number of transactions in the block.
+            // TODO(lior): Check !self.execution_params.is_validator before calling `is_done()`.
+            if lock_executor(&self.executor).is_done() && !self.execution_params.is_validator {
                 info!("Block is full.");
-                if self.execution_params.is_validator {
-                    return Err(BlockBuilderError::FailOnError(FailOnErrorCause::BlockFull));
-                } else {
-                    FULL_BLOCKS.increment(1);
-                }
+                FULL_BLOCKS.increment(1);
                 break;
             }
 
@@ -405,7 +405,6 @@ impl BlockBuilder {
             results,
             &mut self.execution_data,
             &self.output_content_sender,
-            self.execution_params.is_validator,
             &self.pre_confirmed_tx_sender,
         )
         .await
@@ -474,7 +473,6 @@ async fn collect_execution_results_and_stream_txs(
     output_content_sender: &Option<
         tokio::sync::mpsc::UnboundedSender<InternalConsensusTransaction>,
     >,
-    is_validator: bool,
     pre_confirmed_tx_sender: &Option<PreConfirmedTxSender>,
 ) -> BlockBuilderResult<()> {
     assert!(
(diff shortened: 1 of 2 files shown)


liorgold2 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7265)* status: 1 of 5 files reviewed, 1 unresolved discussion (waiting on @alonh5)

___
*[`crates/apollo_batcher/src/block_builder.rs` line 255 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/7265#-OSrq_z53a9HerJ8pt5A:-OSs8EpY6g5Q3DAYLvK-:b-bde4z9) ([raw file](https://github.com/starkware-libs/sequencer/blob/66d0778e80889ec7cbc233a273aa21222f1b3f44/crates/apollo_batcher/src/block_builder.rs#L255)):*

Is it okay if the block is full in validator mode and new txs are still being added to the executor?
Is there a test for that?
</blockquote></details>

It's a valid scenario as for example the proposer may send 100 txs and only then find that the block is full after 50.

In this case, the validator will simply wait until it gets n_txs_in_block. If `n_txs_in_block` is before the tx that resulted in "block is full" (as it should be), then it'll exit immediately. If `n_txs_in_block` is for some reason larger, it'll exit on timeout, failing the proposal.

Refactored this code a bit to avoid the unnecessary lock and the `info` log that would've been printed each iteration again and again.

There's this test: `partial_chunk_execution_validator` which I think tests what you wanted. Let me know if it does.

alonh5 (approved):
:lgtm:

> Reviewed 4 of 4 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7265#-:-OSsT90l0BOZ4DDEUm-k:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @liorgold2)

___
*[`crates/apollo_batcher/src/block_builder.rs` line 255 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/7265#-OSrq_z53a9HerJ8pt5A:-OSsRx5qEr2pUkjkPnZv:b51j6ql) ([raw file](https://github.com/starkware-libs/sequencer/blob/66d0778e80889ec7cbc233a273aa21222f1b3f44/crates/apollo_batcher/src/block_builder.rs#L255)):*
<details><summary><i>Previously, liorgold2 wrote…</i></summary><blockquote>

It's a valid scenario as for example the proposer may send 100 txs and only then find that the block is full after 50.

In this case, the validator will simply wait until it gets n_txs_in_block. If `n_txs_in_block` is before the tx that resulted in "block is full" (as it should be), then it'll exit immediately. If `n_txs_in_block` is for some reason larger, it'll exit on timeout, failing the proposal.

Refactored this code a bit to avoid the unnecessary lock and the `info` log that would've been printed each iteration again and again.

There's this test: `partial_chunk_execution_validator` which I think tests what you wanted. Let me know if it does.
</blockquote></details>

okay sounds good.
Not related to this PR specifically but is there also a test where the validator executes more txs than are eventually included in the block? To test the `execution_data.remove_last_txs`
(non blocking)

liorgold2 (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7265)* status: all files reviewed, 1 unresolved discussion (waiting on @alonh5)

___
*[`crates/apollo_batcher/src/block_builder.rs` line 255 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/7265#-OSrq_z53a9HerJ8pt5A:-OSsVYmr-zVCW6rxnkpc:bova6vg) ([raw file](https://github.com/starkware-libs/sequencer/blob/66d0778e80889ec7cbc233a273aa21222f1b3f44/crates/apollo_batcher/src/block_builder.rs#L255)):*

okay sounds good.
Not related to this PR specifically but is there also a test where the validator executes more txs than are eventually included in the block? To test the `execution_data.remove_last_txs`
(non blocking)
</blockquote></details>

Yes. `test_validate_block_excluded_txs`

alonh5 (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7265)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @liorgold2)

PR #7390, opened 2025-06-17
apollo_consensus: use time crate in manager

giladchase:


matan-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7390)* status: all files reviewed, 1 unresolved discussion (waiting on @giladchase)

___
*[`crates/apollo_consensus/src/manager.rs` line 296 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/7390#-OT6OPEg9aQ294zcNVvN:-OT6VjhG2EOdsYy7amBG:b948f56) ([raw file](https://github.com/starkware-libs/sequencer/blob/5f84a22705776330a3e00a8335ca152bac4e1fbb/crates/apollo_consensus/src/manager.rs#L296)):*

Ya we wanted it to stay just a clock.
RE default impl, we can do that, but it'd make the function not easily testable.
I'll do whichever solution you'd like of course, doesn't matter too much.
</blockquote></details>

This is fine, just wanted to refresh my memory.
Maybe in the future we could do 2 traits, Clock + ClockExt, like Future and FutureExt, but not now.

PR #7445, opened 2025-06-18
l1: impl base layer trait in monitored base layer

giladchase:
Impl all by delegation.

Also added a wrapper around `MutexGuard` to hide the type (it's an
implentation detail).

### DIFF
--- a/crates/papyrus_base_layer/src/lib.rs
+++ b/crates/papyrus_base_layer/src/lib.rs
@@ -37,7 +37,7 @@ pub enum MockError {}
 #[cfg_attr(any(feature = "testing", test), automock(type Error = MockError;))]
 #[async_trait]
 pub trait BaseLayerContract {
-    type Error: Error + PartialEq + Display + Debug;
+    type Error: Error + PartialEq + Display + Debug + Send + Sync;
 
     /// Get the latest Starknet block that is proved on the base layer at a specific L1 block
     /// number. If the number is too low, return an error.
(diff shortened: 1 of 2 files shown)


giladchase (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7445)* status: all files reviewed, 1 unresolved discussion (waiting on @matan-starkware)

___
*[`crates/papyrus_base_layer/src/monitored_base_layer.rs` line 76 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/7445#-OT5srPt5xi5dpefs7sq:-OT5uGU6CcFLn8Li-w4Q:bjlzops) ([raw file](https://github.com/starkware-libs/sequencer/blob/4c01fa7aec63c27c3fdd0b30fbec29130c1b2a1a/crates/papyrus_base_layer/src/monitored_base_layer.rs#L76)):*

Since this keeps repeating do we want to:
```
impl<B> From<B::Error> for MonitoredBaseLayerError<B> {}
```
Then I think we could just return with `?` and that will handle the mapping.
</blockquote></details>

Ya this is what `#[from]` does basically in thiserror, but we can't use it because it's a special case of a generic `From<T>`, which conflicts with the blanket implementation (meaning generic From's are generally impossible due to the conflict, and this is a special case)

giladchase (commented):
> Reviewed 1 of 2 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7445)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @giladchase)

PR #7712, opened 2025-07-01
apollo_gateway: expose l2_gas_price towards reject on insufficient gas price

ArniStarkware:

PR #7751, opened 2025-07-03
apollo_gateway: reject tx with low l2 gas price

ArniStarkware:


ArniStarkware:
Already covered by another PR: #7712

PR #7752, opened 2025-07-03
apollo_network: receiving broadcast messages in network stress test

sirandreww-starkware:


sirandreww-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7752)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @ShahakShama)

___
*[`crates/apollo_network/src/bin/network_stress_test/main.rs` line 124 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/7752#-OUTH9BA1KaQTWC_wlIO:-OUTH9BA1KaQTWC_wlIP:b-j1e67) ([raw file](https://github.com/starkware-libs/sequencer/blob/cefe5132ed5af78ffed006ed871de7d4e2d6d0df/crates/apollo_network/src/bin/network_stress_test/main.rs#L124)):*
> ```XML
>     counter!("message_received_delay_micros_sum").increment(delay);
>     counter!(format!("message_received_delay_micros_sum_from_{}", received_message.id))
>         .increment(delay);
> ```

Not sure about what to log here



_Code quote:_
```XML
    counter!("message_received").increment(1);
    counter!(format!("message_received_from_{}", received_message.id)).increment(1);

    gauge!("message_received_delay_micros").set(delay_float);
    gauge!(format!("message_received_delay_micros_from_{}", received_message.id)).set(delay_float);

    counter!("message_received_delay_micros_sum").increment(delay);
    counter!(format!("message_received_delay_micros_sum_from_{}", received_message.id))
        .increment(delay);
```

ShahakShama (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7752)* status: all files reviewed, 3 unresolved discussions (waiting on @eitanm-starkware and @sirandreww-starkware)

___
*[`crates/apollo_network/src/bin/network_stress_test/main.rs` line 101 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/7752#-OUV-XTkE4gMRJk9IC-T:-OV1Taln1fUTagA8AkQg:b-h66hxk) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3ce143635f400c48fab25a5112f6a00e3148296/crates/apollo_network/src/bin/network_stress_test/main.rs#L101)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

stress_test_message -> stress_test_message_result or message_result
___
*[`crates/apollo_network/src/bin/network_stress_test/main.rs` line 108 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/7752#-OUV-uC32iLcDY3c6SWA:-OV1TSMY3re53gesQfjo:ba08tv2) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3ce143635f400c48fab25a5112f6a00e3148296/crates/apollo_network/src/bin/network_stress_test/main.rs#L108)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

I actually thought that was a case that I want to know about, that's why I removed that part...
This is the original code:
```rust
let end_time = SystemTime::now();
let start_time = received_message.time;
let duration = match end_time.duration_since(start_time) {
  #[allow(clippy::as_conversions)]
  Ok(duration) => duration.as_micros() as i128,
  #[allow(clippy::as_conversions)]
  Err(_) => -(start_time.duration_since(end_time).unwrap().as_micros() as i128),
};
```
It bothers me that duration can be negative, that seems very incorrect, because like, physics :sweat_smile: 
So I thought I would do something about later (like use a more accurate clock library that allows me to compare timestamps across devices accurately)

If I cannot rely on the clock time then our results cannot be trustworthy...
</blockquote></details>

Duration can be negative if the clocks aren't synced. We should sum it as this cancels the clocks being out of sync
___
*[`crates/apollo_network/src/bin/network_stress_test/main.rs` line 115 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/7752#-OUV04BJ82xSmdzkiSVn:-OV1ShnF30RiXBOB4SH4:b-8e2aho) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3ce143635f400c48fab25a5112f6a00e3148296/crates/apollo_network/src/bin/network_stress_test/main.rs#L115)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

There's only `as_secs_f64` in the non-nightly version of rust. I wanted the delay to be in an integer type so I can create a counter metric that accumulates it, that way average calculations will be more accurate since there are no small numeric errors that aggregate across additions
</blockquote></details>

I see no reason to do both. Either work with integers and micros, or with floats and seconds
___
*[`crates/apollo_network/src/bin/network_stress_test/main.rs` line 196 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/7752#-OUV0fb89foJOodNWuux:-OV1T18P5s768qU40Pff:b-cynqkk) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3ce143635f400c48fab25a5112f6a00e3148296/crates/apollo_network/src/bin/network_stress_test/main.rs#L196)):*
<details><summary><i>Previously, sirandreww-starkware (Andrew Luka) wrote…</i></summary><blockquote>

Done. I feel like this is a bit of a nitpick to be honest.
If anything unreachable is more suitable because it indicates that this location in the code should not have been reached.
Why would panic be preferable in your opinion?
</blockquote></details>

IMO unreachable is when the reason this is unreachable is clear in this file and panic is if the reason this should not happen comes from a different file
I'd be happy to hear how you see it :)

sirandreww-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7752)* status: all files reviewed, 3 unresolved discussions (waiting on @eitanm-starkware and @ShahakShama)

___
*[`crates/apollo_network/src/bin/network_stress_test/main.rs` line 101 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/7752#-OUV-XTkE4gMRJk9IC-T:-OV3CwNC7ljmNpRteSzg:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3ce143635f400c48fab25a5112f6a00e3148296/crates/apollo_network/src/bin/network_stress_test/main.rs#L101)):*

stress_test_message -> stress_test_message_result or message_result
</blockquote></details>

Done.
___
*[`crates/apollo_network/src/bin/network_stress_test/main.rs` line 108 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/7752#-OUV-uC32iLcDY3c6SWA:-OV3Cy0kEZxnATwuHfTI:bcotyr3) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3ce143635f400c48fab25a5112f6a00e3148296/crates/apollo_network/src/bin/network_stress_test/main.rs#L108)):*

Duration can be negative if the clocks aren't synced. We should sum it as this cancels the clocks being out of sync
</blockquote></details>

How would summing this cancel out these errors? 
There are very few assumptions you can make about timing between devices, here's an interesting post about this:
<https://infiniteundo.com/post/25326999628/falsehoods-programmers-believe-about-time>
I don't think a negative duration will ever happen, but if it did, I'd like to know so we can have a more in-depth conversation on how to deal with that... Is that something you guys encountered previously?
___
*[`crates/apollo_network/src/bin/network_stress_test/main.rs` line 115 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/7752#-OUV04BJ82xSmdzkiSVn:-OV3DP_H8WGIUHKOkdJb:b-oi706h) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3ce143635f400c48fab25a5112f6a00e3148296/crates/apollo_network/src/bin/network_stress_test/main.rs#L115)):*

I see no reason to do both. Either work with integers and micros, or with floats and seconds
</blockquote></details>

There is though, gauges expect floats but integers will always be more precise when summing is involved.
Since I'd like to have a gauge as a sanity metric that I can always look at to check my queries or assumptions I'd like to have the duration object and use it however is necessary. 
___
*[`crates/apollo_network/src/bin/network_stress_test/main.rs` line 196 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/7752#-OUV0fb89foJOodNWuux:-OV3E2Au85ms2JKhYUvJ:b-wbmrz) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3ce143635f400c48fab25a5112f6a00e3148296/crates/apollo_network/src/bin/network_stress_test/main.rs#L196)):*

IMO unreachable is when the reason this is unreachable is clear in this file and panic is if the reason this should not happen comes from a different file
I'd be happy to hear how you see it :)
</blockquote></details>

Even under this definition this should be unreachable. 
Personally I view unreachable and panic as tools to give more information about the error that occurred. When a programmer thinks some line in the code should not be reached under the assumption he has for his program he would indicate so by using unreachable. This way it would be obvious that these assumptions were violated and this is a bug that was not accounted for. If instead he/she would like to create additional assumptions on inputs he would use a panic.

To clarify:
If I encounter a panic in a library function I called, I check myself and my inputs. If I encounter an unreachable block, I file an issue...

sirandreww-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7752)* status: all files reviewed, 1 unresolved discussion (waiting on @ShahakShama)

___
*[`crates/apollo_network/src/bin/network_stress_test/main.rs` line 108 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/7752#-OUV-uC32iLcDY3c6SWA:-OVCneu_9KAqXuGFArax:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3ce143635f400c48fab25a5112f6a00e3148296/crates/apollo_network/src/bin/network_stress_test/main.rs#L108)):*

Discussed f2f. I'm ok with leaving this panic, just add a proper message to it (consider adding that the clocks are unsynced to the message)
</blockquote></details>

Done
___
*[`crates/apollo_network/src/bin/network_stress_test/main.rs` line 115 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/7752#-OUV04BJ82xSmdzkiSVn:-OVBteRaC6dxW7Z0uz4V:b1n9zut) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3ce143635f400c48fab25a5112f6a00e3148296/crates/apollo_network/src/bin/network_stress_test/main.rs#L115)):*

why integers will be more precise? You're working with micros. The percentage of error of floats is much less than 10^-6
Let's work with float everywhere
</blockquote></details>

You can't work with float everywhere, counters must be integers. Furthermore, the floats will represent seconds because there is no "duration.as_micros_f64" in stable rust, so the 10^-6 can get significant when summing over millions of numbers (20 minutes of actual runtime)
I think it would be best if we extract from duration as needed. What is the issue with calling both "as_secs_f64" and "as_micros"?
___
*[`crates/apollo_network/src/bin/network_stress_test/main.rs` line 196 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/7752#-OUV0fb89foJOodNWuux:-OVCnjWR0_7Bh3_wzBZ0:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3ce143635f400c48fab25a5112f6a00e3148296/crates/apollo_network/src/bin/network_stress_test/main.rs#L196)):*

Feel free to revert back to unreachable if you prefer :)
</blockquote></details>

Done

PR #7767, opened 2025-07-03
starknet_os: test compiled class version

aner-starkware:


dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7767)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @aner-starkware)

___
*[`crates/starknet_os/src/constants_test.rs` line 70 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/7767#-OUFX3Nr8SaMNVUAu7fl:-OUFX3Nr8SaMNVUAu7fm:b9mfjwp) ([raw file](https://github.com/starkware-libs/sequencer/blob/497ee810f390d053ea0a143c99c3fc8d4f79b94c/crates/starknet_os/src/constants_test.rs#L70)):*
> ```XML
> fn test_compiled_class_version() {
>     // The hex string corresponding to b'COMPILED_CLASS_V1' in big-endian.
>     let compiled_class_version = "0x434f4d50494c45445f434c4153535f5631";
> ```

we don't need snapshot testing - we only need consistency testing.
since this const doesn't appear in the sequencer anywhere, it shouldn't be tested here



_Code quote:_
```XML
let compiled_class_version = "0x434f4d50494c45445f434c4153535f5631";
```

PR #7898, opened 2025-07-10
ci: fix merge_branches script to unmerged-deleted mode

matanl-starkware:


dorimedini-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7898)* status: all files reviewed, 1 unresolved discussion (waiting on @matanl-starkware)

___
*[`scripts/merge_branches.py` line 115 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/7898#-OUnycKLAFrQjKtlUiao:-OUnycKLAFrQjKtlUiap:beyrhkq) ([raw file](https://github.com/starkware-libs/sequencer/blob/056c4e769886f0fae00036097bb6c3e5da3510dc/scripts/merge_branches.py#L115)):*
> ```Python
>     run_command(f"git checkout origin/{dst_branch} {' '.join(FILES_TO_PRESERVE) }")
> 
>     run_command("git status -s | grep -E \"^UU|^AA|^UD\" | awk '{ print $2 }' | tee /tmp/conflicts")
> ```

tried adding something like this, but shahak pointed out that we don't always want this to silently succeed...
maybe gate this behind a boolean argument to the script? `--delete-from-dst` or something, defaulting to `False`?



_Code quote:_
```Python
|^UD
```

PR #7947, opened 2025-07-13
l1: return dummy answers on height < start_height

giladchase:
This is introduced in order to tighten the coupling with the batcher, to
accommodate its re-execution-from-genesis state, which ignores l1
interaction.
Note that this will hide batcher startup errors, for example off-by-one
errors.

Suggested alternative: return a dedicated error and let the Batcher
decide if this is an Error or not, based on whatever special state _it_
is in.


giladchase (commented):
Moving this back to draft, i wanna check if validate/get-txs are really called in reexecution or not

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/7947#-:-OV6khsI8qllRCyibW40:b7azyso)* status: 0 of 3 files reviewed, all discussions resolved (waiting on @elintul and @matan-starkware)

PR #7726, opened 2025-07-02
Victor/remove rejected transactions from cende

victorkstarkware:

PR #8168, opened 2025-07-23
apollo_consensus_orchestrator: couple transaction with receipt for cende

victorkstarkware:
### What changed?

- Created a new `InternalTransactionWithReceipt` struct that pairs a transaction with its execution info
- Modified `BlobParameters` to use `transactions_with_execution_infos` instead of separate `transactions` and `execution_infos` fields
- Updated the `AerospikeBlob::from_parameters` method to unpack the new structure
- Refactored the transaction processing in `SequencerConsensusContext` to match transactions with their execution infos and create the paired structure
- Updated test code in `central_objects_test.rs` to use the new structure
- `central_blob.json` was updated. Previously, it had a non-matching number of transactions and receipts (there should be one receipt for each transaction).

### Motivation

Previously, transactions and their receipts were kept in separate `Vec<>`'s when sent to cende (with an assumption that the same index in both `Vec<>`'s correspond to the same transaction). This assumption was broken due to pre-sending, which introduced the following bug in cende:

https://github.com/starkware-libs/sequencer/pull/7726

To remedy this, a new struct `InternalTransactionWithReceipt` that pairs a transaction with its execution info was created, enforcing the correspondence of a transaction and its receipt for cende.

### What didn't change

`AerospikeBlob` still uses separate `Vec<>`'s as this struct mirrors the pythonic logic. Changing the pythonic logic is performance-critical and thus out of scope here.


victorkstarkware (commented):
Made the following changes:
* First create a HashMap out of `transactions` and use it here to get the transaction (now time complexity is linear)
* In the previous version, a fail in matching txs with receipts would panic. Now, it gracefully propagates a `ConsensusError::Other` 

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/8168#-:-OWzMxO45tHI9ZH2UF-m:b-5zyrpd)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @alonh5)

dafnamatsry (changes requested):
> Reviewed 3 of 4 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/8168)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @alonh5)

___
*[`crates/apollo_consensus_orchestrator/src/sequencer_consensus_context.rs` line 465 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/8168#-OXYEkHy4TA17cg1L5fE:-OXYEkHy4TA17cg1L5fF:b9o29sl) ([raw file](https://github.com/starkware-libs/sequencer/blob/e158be3f7d8faae77558e4c3530f99bd3e4a40a5/crates/apollo_consensus_orchestrator/src/sequencer_consensus_context.rs#L465)):*
> ```XML
>         // and the value is the transaction itself.
>         let transactions_hash_map = transactions
>             .iter()
> ```

Can you use `into_iter()` to avoid the `clone()`?



_Code quote:_
```XML
.iter()
```
___
*[`crates/apollo_consensus_orchestrator/src/sequencer_consensus_context.rs` line 478 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/8168#-OXYFiSg2605sk7vmkZi:-OXYFiSg2605sk7vmkZj:bd6eevq) ([raw file](https://github.com/starkware-libs/sequencer/blob/e158be3f7d8faae77558e4c3530f99bd3e4a40a5/crates/apollo_consensus_orchestrator/src/sequencer_consensus_context.rs#L478)):*
> ```XML
>                 error!("{e}");
>                 ConsensusError::Other(e.to_string())
>             })?;
> ```

Consider using a simple for loop here:



_Suggestion:_
```XML
        // A hash map of (possibly failed) transactions, where the key is the transaction hash
        // and the value is the transaction itself.
        let transactions_hash_map = HashMap::new();
        for tx in trensactions.into_iter().flatten() {
            if map.insert(tx.tx_hash(), tx).is_some() {
                return Err("Duplicate transactions found with the same tx_hash: {key:?}");
            }
        })
```
___
*[`crates/apollo_consensus_orchestrator/src/sequencer_consensus_context.rs` line 491 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/8168#-OXYHwDTDvrQ0wRv9kQw:-OXYHwDU2rmPA09rj8XG:b4hc7ug) ([raw file](https://github.com/starkware-libs/sequencer/blob/e158be3f7d8faae77558e4c3530f99bd3e4a40a5/crates/apollo_consensus_orchestrator/src/sequencer_consensus_context.rs#L491)):*
> ```XML
>             .execution_infos
>             .into_iter()
>             .map(|(tx_hash, execution_info)| match transactions_hash_map.get(&tx_hash) {
> ```

Try using `remove` instead of `get` to get an owned value and avoid the`clone()`.



_Code quote:_
```XML
transactions_hash_map.get(&tx_hash)
```

dafnamatsry (approved):
:lgtm:

> @dafnamatsry reviewed 1 of 4 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/8168#-:-OYz9feO3BpkB2klDxsr:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @alonh5)

PR #8546, opened 2025-08-12
apollo_rpc: use the actual compiled class hash when declering a class in tests

AvivYossef-starkware:


noaov1 (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/8546)* status: all files reviewed, 1 unresolved discussion (waiting on @AvivYossef-starkware and @meship-starkware)

___
*[`crates/apollo_rpc/src/v0_8/execution_test.rs` line 1648 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/8546#-OXTR3bDDhl99R48L9us:-OXTdfHtAb6JnpQ_WDv2:bc2f1of) ([raw file](https://github.com/starkware-libs/sequencer/blob/59e001e21773f7a053d2de6b19d35a20988b4015/crates/apollo_rpc/src/v0_8/execution_test.rs#L1648)):*
<details><summary><i>Previously, AvivYossef-starkware wrote…</i></summary><blockquote>

No, because they used the default value,but I wanted to change it
</blockquote></details>

I think it's best to avoid calculating the casm hash if not needed. Why did you want to change it?

AvivYossef-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/8546)* status: all files reviewed, 1 unresolved discussion (waiting on @meship-starkware and @noaov1)

___
*[`crates/apollo_rpc/src/v0_8/execution_test.rs` line 1648 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/8546#-OXTR3bDDhl99R48L9us:-OXTgoSP290gvwwA28rq:b-ks0nfw) ([raw file](https://github.com/starkware-libs/sequencer/blob/59e001e21773f7a053d2de6b19d35a20988b4015/crates/apollo_rpc/src/v0_8/execution_test.rs#L1648)):*

I think it's best to avoid calculating the casm hash if not needed. Why did you want to change it?
</blockquote></details>

Because the default value is treated as a cairo zero contract. 
I can use a dummy value, but it's risky because if we change the state reader to use the class manager, it might return the actual hash. Since there's only one contract, I thought that it was okay.

meship-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/8546)* status: all files reviewed, 1 unresolved discussion (waiting on @noaov1)

PR #8602, opened 2025-08-17
blockifier_reexecution: add 0.13.6 block for reexecution

AvivYossef-starkware:

PR #8615, opened 2025-08-17
blockifier_reexecution: add block 1,600,000 to CI reexecution set

avi-starkware:
Include block 1600000 in `block_numbers_for_reexecution.json` and extend `test_block_reexecution` cases accordingly, so CI downloads and reexecutes this block offline.


avi-starkware:
Covered by #8602

PR #8677, opened 2025-08-19
apollo_l1_gas_price: do not print headers in eth_to_strk_orcale startup

guy-starkware:


Itay-Tsabary-Starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/8677)* status: 0 of 2 files reviewed, 2 unresolved discussions

___
*[`crates/apollo_l1_gas_price/src/eth_to_strk_oracle.rs` line 144 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/8677#-OY1i8dG3jNAQQIiFwUX:-OY1i8dG3jNAQQIiFwUY:b-udb2c0) ([raw file](https://github.com/starkware-libs/sequencer/blob/dbecf067411f2220352c0a56662d4a19b540bbfa/crates/apollo_l1_gas_price/src/eth_to_strk_oracle.rs#L144)):*
> ```XML
>                 .url_header_list
>                 .as_ref()
>                 .map_or_else(Vec::new, |v| v.iter().map(|uh| uh.safe_print()).collect()),
> ```

Please add an explicit comment that we avoid debugging the entire data as it contains sensitive information.



_Code quote:_
```XML
                .url_header_list
                .as_ref()
                .map_or_else(Vec::new, |v| v.iter().map(|uh| uh.safe_print()).collect()),
```
___
*[`crates/apollo_config/src/converters.rs` line 159 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/8677#-OY1hRZJ8-tlh1crndI4:-OY1hRZJ8-tlh1crndI5:b-d805n6) ([raw file](https://github.com/starkware-libs/sequencer/blob/dbecf067411f2220352c0a56662d4a19b540bbfa/crates/apollo_config/src/converters.rs#L159)):*
> ```XML
>     pub fn safe_print(&self) -> String {
>         self.url.to_string()
>     }
> ```

Consider implementing `Display` for this struct, in which only `Display`ing the `url` part.

If not, please rename this fn: it does not `print` but returns a string, and the context is unclear here -- what is "safe"? 

WDYT about `get_url`? The conversion to `String` can be made (explicitly or implicitly) at the callsite.



_Code quote:_
```XML
    /// Print the URL but not the headers, so we don't leak keys when logging.
    pub fn safe_print(&self) -> String {
        self.url.to_string()
    }
```

PR #8924, opened 2025-08-31
scripts: move the update config and restart script to a new dir "prod" for prod modifing scripts

guyf-starkware:


guyf-starkware:
Too many changes in the stack diverge. Will re-do this on a clean stack

PR #9013, opened 2025-09-03
starknet_committer: cached storage

yoavGrs:


dorimedini-starkware (approved):
> @dorimedini-starkware reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9013)* status: all files reviewed, 2 unresolved discussions (waiting on @TzahiTaub and @yoavGrs)

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/9013#-OZK35gxESuIqpRgdTqa:-OZhU2kj9dQ0iSiOYZKr:bvm2u1d) (no related file):*
<details><summary><i>Previously, yoavGrs wrote…</i></summary><blockquote>

WDYT about adding counters for`get` callings and for hitting the cache?
</blockquote></details>

sounds like a good idea, but probably out of scope.
need to find out how apollo tracks metrics, maybe we should plan a dashboard for the committer
___
*[`crates/starknet_patricia_storage/src/map_storage.rs` line 75 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9013#-OZEZZVzAYU7AOrsYAfF:-OZhUUIY2pxINvPQdt9V:bvmcijv) ([raw file](https://github.com/starkware-libs/sequencer/blob/cf4d1227a8f89242f3cc480f0e45c3a68ba8b648/crates/starknet_patricia_storage/src/map_storage.rs#L75)):*
<details><summary><i>Previously, yoavGrs wrote…</i></summary><blockquote>

We assume that many keys will never be read.
Let's update the value but not insert new entry.
</blockquote></details>

what is the benefit? memory conservation?
because `check-and-update` seems no faster than `set`

TzahiTaub (changes requested):
> @TzahiTaub reviewed 3 of 5 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9013)* status: all files reviewed, 4 unresolved discussions (waiting on @dorimedini-starkware and @yoavGrs)

___
*[`crates/starknet_patricia_storage/src/map_storage.rs` line 75 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9013#-OZEZZVzAYU7AOrsYAfF:-OZhyFcy0LrjsbyLEUI4:bnpqsl8) ([raw file](https://github.com/starkware-libs/sequencer/blob/cf4d1227a8f89242f3cc480f0e45c3a68ba8b648/crates/starknet_patricia_storage/src/map_storage.rs#L75)):*

what is the benefit? memory conservation?
because `check-and-update` seems no faster than `set`
</blockquote></details>

I think @yoavGrs means that we don't want to cache every written value because that's says nothing on if we will want to read it in the future, so he doesn't want to possibly evacuate items from the cache for this. An item is "worthy" of getting into the cache only if we tried to read it. As the doc of the struct says "/// Only getter methods are cached."
___
*[`crates/starknet_patricia_storage/src/map_storage.rs` line 78 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9013#-OZEZy4uEU3Z35RUPhlZ:-OZi-f560c2K9t-pPoDS:b-vwt0mw) ([raw file](https://github.com/starkware-libs/sequencer/blob/cf4d1227a8f89242f3cc480f0e45c3a68ba8b648/crates/starknet_patricia_storage/src/map_storage.rs#L78)):*
<details><summary><i>Previously, yoavGrs wrote…</i></summary><blockquote>

Done.
</blockquote></details>

But we don't read from the cache at all in this implementation, we're just doing more work than working with the storage directly. We need to get all the values for the keys that exist in the cache, and mget from storage+ update cache only the keys that don't exist in the cache.
Something like this? (need to make the `mget` signature to accept `&[&DbKey]` instead of `&[DbKey]` to avoid copying the keys.

```
    fn mget(&mut self, keys: &[&DbKey]) -> PatriciaStorageResult<Vec<Option<DbValue>>> {
        let mut results = Vec::with_capacity(keys.len());
        let mut keys_to_fetch = Vec::new();
        let mut indices_to_fetch = Vec::new();

        for (index, key) in keys.iter().enumerate() {
            if let Some(cached_value) = self.cache.get(key) {
                results.push(Some(cached_value.clone()));
            } else {
                results.push(None); // Placeholder for now
                keys_to_fetch.push(*key);
                indices_to_fetch.push(index);
            }
        }

        if !keys_to_fetch.is_empty() {
            let fetched_values = self.storage.mget(&keys_to_fetch)?;
            fetched_values.into_iter().enumerate().filter_map(|(i, v)| v.map(|v| (i, v))).for_each(
                |(i, v)| {
                    let key = keys_to_fetch[i];
                    self.cache.put(key.clone(), v.clone());
                    results[indices_to_fetch[i]] = Some(v);
                },
            );
        }

        Ok(results)
    }
```
___
*[`crates/starknet_patricia_storage/src/map_storage_test.rs` line 43 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/9013#-OZiI-BkBkujB0xAmxPH:-OZiI-Bl5xlPzzsDBnQM:b-al1kui) ([raw file](https://github.com/starkware-libs/sequencer/blob/437abef82c9160d9b0bb898a0460d15570ffd8a6/crates/starknet_patricia_storage/src/map_storage_test.rs#L43)):*
> ```XML
>     // storage = {1:1, 3:3}
>     assert!(storage.get(&key_2.clone()).unwrap().is_none());
> }
> ```

I would be happy to have a test for the cache speed up (in another PR). Have some MockSlowStorage that calls `sleep` for every `get` and `mget` (can be the same for both or the mget proportional to the number of items, but a lot less than the number of items * get sleep time), and verify that mget for, e.g., 10 storage times that some of are in the cache, is faster than calling mget on the storage directly.

dorimedini-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9013)* status: all files reviewed, 3 unresolved discussions (waiting on @TzahiTaub and @yoavGrs)

___
*[`crates/starknet_patricia_storage/src/map_storage.rs` line 78 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9013#-OZEZy4uEU3Z35RUPhlZ:-OZj-BVo991wMUvZIvsP:b-6r3f40) ([raw file](https://github.com/starkware-libs/sequencer/blob/cf4d1227a8f89242f3cc480f0e45c3a68ba8b648/crates/starknet_patricia_storage/src/map_storage.rs#L78)):*

But we don't read from the cache at all in this implementation, we're just doing more work than working with the storage directly. We need to get all the values for the keys that exist in the cache, and mget from storage+ update cache only the keys that don't exist in the cache.
Something like this? (need to make the `mget` signature to accept `&[&DbKey]` instead of `&[DbKey]` to avoid copying the keys.

```
    fn mget(&mut self, keys: &[&DbKey]) -> PatriciaStorageResult<Vec<Option<DbValue>>> {
        let mut results = Vec::with_capacity(keys.len());
        let mut keys_to_fetch = Vec::new();
        let mut indices_to_fetch = Vec::new();

        for (index, key) in keys.iter().enumerate() {
            if let Some(cached_value) = self.cache.get(key) {
                results.push(Some(cached_value.clone()));
            } else {
                results.push(None); // Placeholder for now
                keys_to_fetch.push(*key);
                indices_to_fetch.push(index);
            }
        }

        if !keys_to_fetch.is_empty() {
            let fetched_values = self.storage.mget(&keys_to_fetch)?;
            fetched_values.into_iter().enumerate().filter_map(|(i, v)| v.map(|v| (i, v))).for_each(
                |(i, v)| {
                    let key = keys_to_fetch[i];
                    self.cache.put(key.clone(), v.clone());
                    results[indices_to_fetch[i]] = Some(v);
                },
            );
        }

        Ok(results)
    }
```
</blockquote></details>

right... good catch!

yoavGrs (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9013)* status: 26 of 28 files reviewed, 2 unresolved discussions (waiting on @dorimedini-starkware and @TzahiTaub)

___
*[`crates/starknet_patricia_storage/src/map_storage.rs` line 78 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9013#-OZEZy4uEU3Z35RUPhlZ:-OZsWEJn9WunxbIFU5uV:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/cf4d1227a8f89242f3cc480f0e45c3a68ba8b648/crates/starknet_patricia_storage/src/map_storage.rs#L78)):*

right... good catch!
</blockquote></details>

Done.
___
*[`crates/starknet_patricia_storage/src/map_storage.rs` line 94 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/9013#-OZtBwVT1v44q-Xu08hF:-O_C9z9B7u5CsS69eJqp:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/493fa7eb9c9a6c89d0246e0d153403f9c7116dcf/crates/starknet_patricia_storage/src/map_storage.rs#L94)):*

There shouldn't be two clones for the value, take the ownership of the fetched_values. I think this should work, and if it doesn't, something similar should.
</blockquote></details>

Done.
___
*[`crates/starknet_patricia_storage/src/map_storage_test.rs` line 43 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/9013#-OZiI-BkBkujB0xAmxPH:-O_CAAJdBFigbkbMmPqo:blbhy6u) ([raw file](https://github.com/starkware-libs/sequencer/blob/437abef82c9160d9b0bb898a0460d15570ffd8a6/crates/starknet_patricia_storage/src/map_storage_test.rs#L43)):*

I would be happy to have a test for the cache speed up (in another PR). Have some MockSlowStorage that calls `sleep` for every `get` and `mget` (can be the same for both or the mget proportional to the number of items, but a lot less than the number of items * get sleep time), and verify that mget for, e.g., 10 storage times that some of are in the cache, is faster than calling mget on the storage directly.
</blockquote></details>

See here
https://reviewable.io/reviews/starkware-libs/sequencer/9013#-OZK35gxESuIqpRgdTqa
___
*[`crates/starknet_patricia_storage/src/map_storage.rs` line 75 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/9013#-OZtBMMe5Uw9Zb0a_psR:-O_C86KnCogLR1mbBgrP:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/493fa7eb9c9a6c89d0246e0d153403f9c7116dcf/crates/starknet_patricia_storage/src/map_storage.rs#L75)):*
> ```XML
> 
>     fn mget(&mut self, keys: &[&DbKey]) -> PatriciaStorageResult<Vec<Option<DbValue>>> {
>         let mut values = vec![None; keys.len()]; // The None values are placeholder.
> ```

Done.

TzahiTaub (changes requested):
> @TzahiTaub reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9013)* status: 26 of 28 files reviewed, 3 unresolved discussions (waiting on @dorimedini-starkware and @yoavGrs)

___
*[`crates/starknet_patricia_storage/src/map_storage.rs` line 94 at r7](https://reviewable.io/reviews/starkware-libs/sequencer/9013#-O_CB9ZrDKVw7eLEwp3R:-O_CB9ZrDKVw7eLEwp3S:b47n8ic) ([raw file](https://github.com/starkware-libs/sequencer/blob/3b27b0dc6df0727d8b1b015f7f6de05b6a225e39/crates/starknet_patricia_storage/src/map_storage.rs#L94)):*
> ```XML
>             .zip(fetched_values.into_iter())
>             .for_each(|((index, key), value)| {
>                 self.cache.put((*key).clone(), value.clone());
> ```

I think this can also be removed



_Suggestion:_
```XML
*key
```

TzahiTaub (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9013)* status: all files reviewed, 1 unresolved discussion (waiting on @yoavGrs)

TzahiTaub (approved):
> @TzahiTaub reviewed 27 of 27 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9013)* status: all files reviewed, 1 unresolved discussion (waiting on @yoavGrs)

dorimedini-starkware (approved):
> @dorimedini-starkware reviewed 1 of 6 files at r6, 27 of 27 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9013)* status: all files reviewed, 1 unresolved discussion (waiting on @yoavGrs)

TzahiTaub (approved):
> @TzahiTaub reviewed 27 of 27 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9013)* status: all files reviewed, 1 unresolved discussion (waiting on @yoavGrs)

PR #9171, opened 2025-09-10
starknet_os_flow_tests: verify state diff

dorimedini-starkware:


dorimedini-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9171)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on @nimrod-starkware and @TzahiTaub)

___
*[`crates/starknet_os_flow_tests/src/test_manager.rs` line 535 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9171#-OZnRklWAK8BtFTtUvNl:-OZnRklWAK8BtFTtUvNm:b-4xrm4y) ([raw file](https://github.com/starkware-libs/sequencer/blob/cb5b1d74d56c9253e93d1aa6e41b086b9b3e37cf/crates/starknet_os_flow_tests/src/test_manager.rs#L535)):*
> ```XML
>         OsTestOutput {
>             runner_output: os_output,
>             decompressed_state_diff: initial_state.nontrivial_diff_maps(decompressed_state_diff),
> ```

Blocking self: I don't like this approach because trivial diffs are probably added due to implementation detail of `as_state_maps`.
when computing the decompressed state diff it would be better to check the actual output of the OS without filtering trivial diffs: if the OS outputs trivial diffs we want to fail the test...



_Code quote:_
```XML
decompressed_state_diff,
```

dorimedini-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9171)* status: 2 of 6 files reviewed, 3 unresolved discussions (waiting on @nimrod-starkware and @TzahiTaub)

___
*[`crates/starknet_committer/src/block_committer/input.rs` line 139 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/9171#-O_gHP0r5nbQrXlrvgN-:-O_gWWWL-rbzGmZhd9dK:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/67b8acbcf90eda303994a73ce9ea685a3f1242b7/crates/starknet_committer/src/block_committer/input.rs#L139)):*

Do you want to return True only iff the storage update for contract address in `self` is equal (and not subset) of the `other`contract address's changes? Please doc this in any case in the function.
</blockquote></details>

Done.
___
*[`crates/starknet_os_flow_tests/src/state_trait.rs` line 15 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/9171#-O_g9mcNCgDMq8lmqaxm:-O_gXuRj8_7se_1_FvHt:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/67b8acbcf90eda303994a73ce9ea685a3f1242b7/crates/starknet_os_flow_tests/src/state_trait.rs#L15)):*

I don't think we should bypass an Err here and in the rest.
</blockquote></details>

Done.
___
*[`crates/starknet_os_flow_tests/src/test_manager.rs` line 204 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/9171#-O_gR0xdAl0KXQ66eP8b:-O_gbleB4I42m-w5topR:b-p1lzr1) ([raw file](https://github.com/starkware-libs/sequencer/blob/67b8acbcf90eda303994a73ce9ea685a3f1242b7/crates/starknet_os_flow_tests/src/test_manager.rs#L204)):*

Can't we have trivial storage updates in the actual?
</blockquote></details>

apparently not, I [added a test case](https://reviewable.io/reviews/starkware-libs/sequencer/9406) to check this
___
*[`crates/starknet_os_flow_tests/src/test_manager.rs` line 462 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/9171#-O_gKjQz8rPYUaZ34k6g:-O_gYGZGDTH1ew27y6M_:b-8xbrjg) ([raw file](https://github.com/starkware-libs/sequencer/blob/67b8acbcf90eda303994a73ce9ea685a3f1242b7/crates/starknet_os_flow_tests/src/test_manager.rs#L462)):*

I don't know if we will need this in the future, but the conversion from the blockifier's `CommitmentStateDiff` to the committer `StateDiff` seems redundant. We can either impl extend for `CommitmentStateDiff` or have a common trait for these.
</blockquote></details>

this type is needed in `commit_state_diff`, what are you suggesting?

TzahiTaub (approved):
:lgtm:

> @TzahiTaub reviewed 4 of 4 files at r6.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9171#-:-O_hs9-J5Wy9LaOlb1cF:bnfp4nl)* status: all files reviewed (commit messages unreviewed), 1 unresolved discussion (waiting on @dorimedini-starkware and @nimrod-starkware)

___
*[`crates/starknet_os_flow_tests/src/test_manager.rs` line 462 at r4](https://reviewable.io/reviews/starkware-libs/sequencer/9171#-O_gKjQz8rPYUaZ34k6g:-O_hG6XF6zsuS2JvdGis:b-fahi28) ([raw file](https://github.com/starkware-libs/sequencer/blob/67b8acbcf90eda303994a73ce9ea685a3f1242b7/crates/starknet_os_flow_tests/src/test_manager.rs#L462)):*

this type is needed in `commit_state_diff`, what are you suggesting?
</blockquote></details>

This is out of scope of this PR, but if it's only relevant to the flow tests, then we can just impl 
`pub fn extend(&mut self, other: CommitmentStateDiff)` (instead of the current where `other: Self`, or a generic impl for both if we want). But if it might be used in prod, as the blockifier and committer will communicate directly without serialization of the statediff in between, we might combine the structs (I'm not sure the IndexMap is necessary in the blockifier)  or have a trait of the fields, so both can be used interchangeably.

TzahiTaub (approved):
> @TzahiTaub reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9171)* status: all files reviewed, 1 unresolved discussion (waiting on @dorimedini-starkware and @nimrod-starkware)

PR #9186, opened 2025-09-11
apollo_dashboard: Add quotes to log query if not present

ron-starkware:


matanl-starkware (changes requested):
> @matanl-starkware reviewed 1 of 1 files at r1, 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9186)* status: all files reviewed, 1 unresolved discussion (waiting on @asmaastarkware and @dafnamatsry)

___
*[`crates/apollo_dashboard/src/dashboard.rs` line 192 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/9186#-OZt4YMMEAEmpDRgDeMl:-OZt4YMMEAEmpDRgDeMm:bnnbndo) ([raw file](https://github.com/starkware-libs/sequencer/blob/372b6efe151fdd366a8855bc41046c372011dcfe/crates/apollo_dashboard/src/dashboard.rs#L192)):*
> ```XML
> 
>         // Add quotes if none are present.
>         if !query.contains('"') {
> ```

I'm not sure that's good enough.
What if someone puts 'SEARCH("Some string")' ? Will that work?
Maybe you should always check for prefix and suffix?

ron-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9186)* status: all files reviewed, 1 unresolved discussion (waiting on @asmaastarkware, @dafnamatsry, and @matanl-starkware)

___
*[`crates/apollo_dashboard/src/dashboard.rs` line 192 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/9186#-OZt4YMMEAEmpDRgDeMl:-OZtD4xjD9CiBgs6BBYq:bkbkb0g) ([raw file](https://github.com/starkware-libs/sequencer/blob/372b6efe151fdd366a8855bc41046c372011dcfe/crates/apollo_dashboard/src/dashboard.rs#L192)):*

I'm not sure that's good enough.
What if someone puts 'SEARCH("Some string")' ? Will that work?
Maybe you should always check for prefix and suffix?
</blockquote></details>

Yes, that's the reason I did it this way. If someone for example does `log_query("cluster=\"foo\""))` then I don't want to add quotes to this because then it'll just be a string in Google Logs and won't work like the user expects.
I only want to add quotation marks if there are no quotation marks present, since if there are any present, the query is a 'complicated' query and adding quotation marks will fail it.
For example if someone wants to search for `"Decision Reached" OR "Decision Not Reached"`, then he has to himself put the quotation marks since Rust won't let not add them. In this case, I don't want to add any additional ones as well.
If someone does `log_query("SEARCH(\"Some string\")")` as in your example, we don't want my function to add any quotation marks (and it won't in this case). Did I understand correctly the example you provided?

matanl-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9186)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @asmaastarkware and @dafnamatsry)

PR #9484, opened 2025-10-01
starknet_os_flow_tests: migrate test_os_logic and some small refactors

dorimedini-starkware:


dorimedini-starkware (commented):
good catch! this parametrization was added after I migrated the test...
I'll add the private keys parametrization now.
the os_test_variant is unecessary; it's just for converage. I'll update the coverage PR ([later in the stack](https://reviewable.io/reviews/starkware-libs/sequencer/9556))

as for your questions:
1. this, and other improvements (e.g. reduce boilerplate when adding declare/deploy txs), are a good idea, but the stack is already huge; we can do this later :)
2. correct, no empty block is added. do you think it's important?

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9484#-:-OcAMTBpDjbX1qh5qkhD:b-q92xov)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @AvivYossef-starkware, @meship-starkware, and @Yoni-Starkware)

___
*[`crates/starknet_os_flow_tests/src/tests.rs` line 495 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9484#-Oc0DABk6IY5if-KiMSM:-OcAKs8H06Go5YakLtEb:b-ylkdkw) ([raw file](https://github.com/starkware-libs/sequencer/blob/65aee29b1889489c90144f600f4f9bf2dbfae943/crates/starknet_os_flow_tests/src/tests.rs#L495)):*
<details><summary><i>Previously, AvivYossef-starkware wrote…</i></summary><blockquote>

I don't think that we add an empty block here, like in `split_txs_into_blocks_in_multi_blocks`
</blockquote></details>

yes, this is why the expected_block_number has the `-1`
___
*[`crates/starknet_os_flow_tests/src/tests.rs` line 682 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9484#-Oc0SJyDD9elA5ML1OQC:-OcALuJvBjzVc7N04UJ8:b-ulamj4) ([raw file](https://github.com/starkware-libs/sequencer/blob/65aee29b1889489c90144f600f4f9bf2dbfae943/crates/starknet_os_flow_tests/src/tests.rs#L682)):*
<details><summary><i>Previously, AvivYossef-starkware wrote…</i></summary><blockquote>

Why is it needed?
</blockquote></details>

seems it really doesn't test anything interesting (was like this in the original test), removed
___
*[`crates/starknet_os_flow_tests/src/tests.rs` line 699 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9484#-Oc0RSdvDppKoGbn5ueQ:-OcALz042XaTessALF0w:b-7ropln) ([raw file](https://github.com/starkware-libs/sequencer/blob/65aee29b1889489c90144f600f4f9bf2dbfae943/crates/starknet_os_flow_tests/src/tests.rs#L699)):*
<details><summary><i>Previously, AvivYossef-starkware wrote…</i></summary><blockquote>

Why is it needed?
</blockquote></details>

ditto
___
*[`crates/starknet_os_flow_tests/src/tests.rs` line 516 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9484#-Oc0F3C0B9ebSkoCgtDn:-OcALB7hCsfuJL-p_0-N:bxyp4en) ([raw file](https://github.com/starkware-libs/sequencer/blob/65aee29b1889489c90144f600f4f9bf2dbfae943/crates/starknet_os_flow_tests/src/tests.rs#L516)):*
> ```XML
>         contract_addresses[0],
>         "test_call_contract",
>         &[**contract_addresses[0], selector_from_name("send_message").0, Felt::ONE, Felt::from(85)],
> ```

how about like this? better?

dorimedini-starkware (commented):
good catch! this parametrization was added after I migrated the test...
I'll add the private keys parametrization now.
the os_test_variant is unecessary; it's just for converage. I'll update the coverage PR ([later in the stack](https://reviewable.io/reviews/starkware-libs/sequencer/9556))

as for your questions:
1. this, and other improvements (e.g. reduce boilerplate when adding declare/deploy txs), are a good idea, but the stack is already huge; we can do this later :)
2. correct, no empty block is added. do you think it's important?

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9484#-:-OcAn0w04WL9ebNeGXl_:b-q92xov)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @meship-starkware and @Yoni-Starkware)

PR #9512, opened 2025-10-05
shahak/remove_store_sierras_and_casms -> main-v0.14.1

ShahakShama:


guyf-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9512)* status: 0 of 6 files reviewed, 1 unresolved discussion

___
*[`crates/apollo_central_sync/src/lib.rs` line 605 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9512#-ObaIBjH8lSoyNJl2hnb:-ObaIBjH8lSoyNJl2hnc:b-djlh0) ([raw file](https://github.com/starkware-libs/sequencer/blob/9ff5109e9477deb089c3d7e7bb6ba470e3cb82d0/crates/apollo_central_sync/src/lib.rs#L605)):*
> ```XML
>             }
>         }
>         if !self.config.store_sierras_and_casms {
> ```

1. This could have been the bug from before (if the config is off but there were non backward compatible classes it wouldn't have written).
2. Now it will always perform the write regardless of whether this block has non backwards compatible classes. Was this your intention? If so why do we even check for the backward comptability?

PR #9533, opened 2025-10-09
apollo_class_manager: Use raw classes to avoid unwraps and conversions

ron-starkware:


matanl-starkware (changes requested):
I don't really see the big difference.
Can you explain why this code is an improvement over the previous one?

> @matanl-starkware reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9533#-:-ObN9UK-AOuj3YfzuZAy:b16pezz)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ron-starkware)

ron-starkware (commented):
By moving the conversions to the client, the server uses raw types end-to-end. Without unwraps we won't have panics in the server. Conversions to or from `starknet_api` types are done at the client level and the server uses the raw classes same as ClassStorage and the compiler.  Also, if in the future we'll want to change types or versions in `starknet_api`, we'll only need to change the typed client wrapper. What do you think? If you believe it doesn't actually provide any benefit we can remove the TODO.

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9533#-:-ObWqxs35GPYmIwEqKaH:b-pyzv36)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ron-starkware)

matanl-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9533)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ron-starkware)

PR #9537, opened 2025-10-09
apollo_class_manager: Clone value instead of whole class

ron-starkware:


ron-starkware (commented):
I kept the clone for the compiler call (~10 lines below the change in class_manager.rs) since it requires ownership of the class. 

> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9537#-:-Ob6XoD64gM7o_1hKyV_:bhuq7jw)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @matanl-starkware)

matanl-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9537)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @ron-starkware)

___
*[`crates/apollo_compile_to_casm_types/src/lib.rs` line 128 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9537#-ObNLaLr0StvZO_shMwJ:-ObNLaLr0StvZO_shMwK:b-kn58jx) ([raw file](https://github.com/starkware-libs/sequencer/blob/7a0f03fb134c308a141b353100a9845dcffec1c7/crates/apollo_compile_to_casm_types/src/lib.rs#L128)):*
> ```XML
>     fn try_from(class: &RawClass) -> Result<Self, Self::Error> {
>         // Deserialize from the underlying JSON value without cloning the wrapper.
>         serde_json::from_value(class.0.clone())
> ```

The heavy part of the clone is still happening here. 
The wrapper is not something to consider.
Your solution doesn't solve the "double clone" problem.



_Code quote:_
```XML
class.0.clone()
```

ron-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9537)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @matanl-starkware and @ron-starkware)

___
*[`crates/apollo_compile_to_casm_types/src/lib.rs` line 128 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9537#-ObNLaLr0StvZO_shMwJ:-ObWdrYhCnD24iBPUnFW:b-4nf1ed) ([raw file](https://github.com/starkware-libs/sequencer/blob/7a0f03fb134c308a141b353100a9845dcffec1c7/crates/apollo_compile_to_casm_types/src/lib.rs#L128)):*

The heavy part of the clone is still happening here. 
The wrapper is not something to consider.
Your solution doesn't solve the "double clone" problem.
</blockquote></details>

I pushed a possible solution that refactors `SerializedClass` to hold an `Arc (Arc<serde_json::Value>)`. This way, cloning is cheap. `try_unwrap` is used to avoid cloning if possible.

matanl-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9537)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @ron-starkware)

___
*[`crates/apollo_compile_to_casm_types/src/lib.rs` line 128 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9537#-ObNLaLr0StvZO_shMwJ:-ObaaYCwBu0OcvjDGVLc:b-gytaqu) ([raw file](https://github.com/starkware-libs/sequencer/blob/7a0f03fb134c308a141b353100a9845dcffec1c7/crates/apollo_compile_to_casm_types/src/lib.rs#L128)):*

I pushed a possible solution that refactors `SerializedClass` to hold an `Arc (Arc<serde_json::Value>)`. This way, cloning is cheap. `try_unwrap` is used to avoid cloning if possible.
</blockquote></details>

Use Arc::unwrap_or_clone.
Can you check if we have a flow that Arc will be cloned?

ron-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9537)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @matanl-starkware and @ron-starkware)

___
*[`crates/apollo_compile_to_casm_types/src/lib.rs` line 128 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/9537#-ObNLaLr0StvZO_shMwJ:-Obaxflc92jS1fad78SA:b3tlqof) ([raw file](https://github.com/starkware-libs/sequencer/blob/7a0f03fb134c308a141b353100a9845dcffec1c7/crates/apollo_compile_to_casm_types/src/lib.rs#L128)):*

Use Arc::unwrap_or_clone.
Can you check if we have a flow that Arc will be cloned?
</blockquote></details>

1. Fixed.
2. There do exist flows like that. For example, in `CachedClassStorage` for cache population after read, there is `self.classes.set(class_id, class.clone())` and `class` is returned. So in this case we have two `RawClass` instances with one `Arc`, so here the inner value will actually be cloned.

matanl-starkware (approved):
> @matanl-starkware reviewed 1 of 2 files at r1, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/9537)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ron-starkware)

___
*[`crates/apollo_compile_to_casm_types/src/lib.rs` line 60 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/9537#-ObcBFzFEm7WQABUl7_E:-ObcBFzFEm7WQABUl7_F:b91rfsp) ([raw file](https://github.com/starkware-libs/sequencer/blob/b53e16b04c86855fbb31e34cbaf86b6fa090a650/crates/apollo_compile_to_casm_types/src/lib.rs#L60)):*
> ```XML
> 
>     pub fn size(&self) -> RawClassResult<usize> {
>         Ok(serde_json::to_string_pretty(&*self.0)?.len())
> ```

I know this was changed to `to_string`. Expect conflict.



_Code quote:_
```XML
to_string_pretty
```

PR #13787, opened 2026-04-16
deployment: if ubuntu official repo is slow, fallback to azure

idan-starkware:


matanl-starkware (commented):
Image

Given this diff (check against my local sources file), are you sure this is what you wanted?

matanl-starkware (approved):

PR #13682, opened 2026-04-06
apollo_consensus_orchestrator: validate builder address in ProposalInit against config

asmaastarkware:
> [!NOTE]
> **Medium Risk**
> Touches consensus proposal validation logic and can cause proposals to be rejected if the configured builder address is incorrect or mismatched across nodes.
> 
> **Overview**
> Adds **builder address** to `ProposalInit` validation and rejects proposals whose `init.builder` doesn’t match the node’s configured `static_config.builder_address`.
> 
> Plumbs the configured builder address through `SequencerConsensusContext` into `ProposalInitValidation`, and updates validation tests to include the new field.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4abd64a43a8b79275803aae5863beb5f9dfb467a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #13799, opened 2026-04-19
apollo_consensus_orchestrator: validate builder address in ProposalInit against config

matanl-starkware:
## Summary
- Rebased from PR #13682 on top of `matanl/validate-starknet-version`
- Adds `builder_address: ContractAddress` to `ProposalInitValidation`
- Validates in `is_proposal_init_valid()` that `init.builder` matches `static_config.builder_address`, rejecting the proposal before block execution if it doesn't
- Plumbs `self.config.static_config.builder_address` into both `ProposalInitValidation` construction sites in `SequencerConsensusContext`

## Test plan
- Existing test fixture updated with `builder_address: ContractAddress::default()` to match the default proposer address used in tests
- All 16 existing `validate_proposal` tests pass


matanl-starkware:
Decided to drop this.
Currently, the builder address is fixed for all (trusted) nodes.
In the future, each proposer has its own address and this validation doesnt hold.

PR #13850, opened 2026-04-23
blockifier_test_utils: fix ETXTBSY race in cairo package download

avi-starkware:
`verify_cairo1_package` double-checks `cairo1_package_exists` — once
without the lock (fast path) and once after acquiring it. The predicate
only tested `.exists()` on the two binaries, so a concurrent process
could observe the package as "ready" while `tar` was still mid-write:

1. Process A holds the download lock, `tar` is extracting the archive.
2. `tar` extracts sequentially: first binary is closed, second binary
   is created (so `.exists()` is true) and being written to.
3. Process B calls `cairo1_package_exists`, sees both files, skips the
   lock.
4. Process B forks + execs the second binary → Linux returns
   `ETXTBSY` because the file is still open for writing by `tar`.

Fix with the same commit-marker pattern used in `compile_cache.rs`:
write `.download_complete` after `tar` finishes, and require it in
`cairo1_package_exists`. Until the marker exists, concurrent callers
fall through to the lock and wait for the writer.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/crates/blockifier_test_utils/src/cairo_compile.rs
+++ b/crates/blockifier_test_utils/src/cairo_compile.rs
@@ -34,6 +34,13 @@ fn starknet_sierra_compile_binary_path(version: &String) -> PathBuf {
     cairo1_package_dir(version).join("cairo/bin/starknet-sierra-compile")
 }
 
+/// Commit marker written after `tar` extraction fully completes. Required because
+/// a file `.exists()` while `tar` is still writing to it, and exec'ing such a file
+/// returns `ETXTBSY`.
+fn cairo1_package_complete_marker(version: &String) -> PathBuf {
+    cairo1_package_dir(version).join(".download_complete")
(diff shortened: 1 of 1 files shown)


dorimedini-starkware (approved):
> @dorimedini-starkware reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/13850)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on avi-starkware).

dorimedini-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/13850)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on avi-starkware).

PR #13674, opened 2026-04-06
ci,scripts: install LLVM before cargo tools

avi-starkware:
Reorder bootstrap and install_build_tools.sh so LLVM 19 is installed
before cargo tools. This is needed because install_cargo_tools.sh will
install starknet-native-compile which requires LLVM to build.

Safe change: LLVM installation does not depend on Rust.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

---

> [!NOTE]
> **Medium Risk**
> Touches CI/build installation order; low code risk but can break developer/CI environments if the new sequencing has untested assumptions or OS-specific differences.
> 
> **Overview**
> Reorders CI/bootstrap and local build tooling installation so `scripts/dependencies.sh` (LLVM 19 + related native deps) runs *before* `install_cargo_tools.sh`, ensuring cargo-installed tools that compile native code (e.g., `starknet-native-compile`) have LLVM available.
> 
> In the `bootstrap` composite action, `setup_native_deps` is now executed before installing Rust to match the updated dependency ordering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1efda68af5a4f2bf3e6990707d726698c542de38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #14060, opened 2026-05-17
ci,scripts: install LLVM before cargo tools

avi-starkware:
Reorder bootstrap and install_build_tools.sh so LLVM 19 is installed
before cargo tools. This is needed because install_cargo_tools.sh will
install starknet-native-compile which requires LLVM to build.

Safe change: LLVM installation does not depend on Rust.

---

Re-do of #13674, which was accidentally merged into `avi/compiler-verify-fn`
instead of `main`. The change itself (squash commit `08764c4529`) is identical;
this PR cherry-picks that commit onto `main`.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.github/actions/bootstrap/action.yml
+++ b/.github/actions/bootstrap/action.yml
@@ -12,10 +12,10 @@ runs:
     - name: Setup Namespace Cache
       uses: ./.github/actions/namespace_cache
 
+    - name: Install cairo native.
+      uses: ./.github/actions/setup_native_deps
+
     - name: Install rust.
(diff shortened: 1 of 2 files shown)


dorimedini-starkware (approved):
> @dorimedini-starkware reviewed 2 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14060)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on avi-starkware).

PR #14068, opened 2026-05-18
apollo_node: spawn a blocking thread when writing to stdout in tests

dorimedini-starkware:


dorimedini-starkware (commented):
> @dorimedini-starkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14068)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on matanl-starkware and ShahakShama).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/14068#-OsvCuM2CX98PAefXVcR:-OsvCuM2CX98PAefXVcS:b2rrd7j) (no related file):*
from claude:
```
● Root cause: write_annotated_stdout_line is a blocking call (acquires std::sync::Mutex, calls flush()) called directly on a tokio worker thread inside tokio::spawn. With 5+ nodes starting at the same time, all their pipe tasks compete for the
  global stdout mutex and each block a worker on flush(). When enough workers are blocked, the monitoring endpoint's futures stop being scheduled. await_alive polls 2500 times over 4+ minutes and sees nothing — hence the misleading "should be
  alive" panic.

  The 4-minute gap between node-internal log timestamps and CI timestamps is the tell: the node's stdout pipe filled up because the pipe task was slow to drain it (workers blocked), so the node's own logging calls blocked, and all that buffered
  output only flushed when the node was eventually killed after await_alive gave up.

  What the fix does: moves the blocking stdout write to spawn_blocking so it runs on tokio's dedicated blocking thread pool rather than occupying an async worker. The async runtime stays free to schedule the monitoring endpoint.
```

matanl-starkware (approved):
> @matanl-starkware reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14068)* status: all files reviewed, 1 unresolved discussion (waiting on ShahakShama).

dorimedini-starkware (commented):
> @dorimedini-starkware resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14068)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ShahakShama).

PR #14069, opened 2026-05-18
apollo_integration_tests: fix non-interpolating .expect() messages and log typos

ron-starkware:
`.expect()` is an ordinary method on Option/Result that takes a `&str` and prints it verbatim on panic — it does not invoke `format_args!`. So named arguments inside the braces never get substituted:

    Before: thread '...' panicked at 'Service {service:?} does not exist in the running executables map.'
    After:  thread '...' panicked at 'Service Batcher does not exist in the running executables map.'

The panic still fires at the right moment for the right reason; only the diagnostic value is lost, which defeats the whole point of embedding the variable. The fix is to wrap the call as `.unwrap_or_else(|| panic!(...))` (or `|err| panic!("...: {err}")` on a Result), since `panic!` IS a format macro and the placeholders interpolate normally.

.expect() placeholder fixes (panic message would have been useless before):
  - apollo_storage/src/header.rs                                 ({block_number})
  - apollo_consensus_manager/src/consensus_manager.rs            ({height})
  - apollo_consensus/src/manager.rs                              (was {self.height} -- field access also invalid in fmt; corrected to local {height})
  - apollo_storage/src/serialization/serializers_test.rs         (7 sites; one referenced an out-of-scope {casm_file}, corrected to {bin_file_name})
  - starknet_os/.../state_diff_encryption/utils.rs               ({public_key}, {sn_public_key})
  - apollo_integration_tests/src/integration_test_manager.rs     ({service:?}, {node_idx})
  - apollo_batcher/src/batcher.rs                                ({new_latest_height})

No behavior change beyond what panic messages display; cargo check --tests passes on all touched crates.

### DIFF
--- a/crates/apollo_batcher/src/batcher.rs
+++ b/crates/apollo_batcher/src/batcher.rs
@@ -1428,7 +1428,9 @@ impl Batcher {
                     .storage_reader
                     .global_root(new_latest_height)
                     .expect("Failed to get global root from storage.")
-                    .expect("Global root is not set for height {new_latest_height}.");
+                    .unwrap_or_else(|| {
+                        panic!("Global root is not set for height {new_latest_height}.")
+                    });
                 assert_eq!(
                     global_root, stored_global_root,
                     "The given global root does not match the stored global root for height \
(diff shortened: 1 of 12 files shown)


matanl-starkware (approved):
> @matanl-starkware reviewed 12 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14069)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ron-starkware).

PR #14110, opened 2026-05-19
apollo_deployments: delete dead code

asaf-sw:
## Description

Delete `crates/apollo_deployments/src/addresses.rs` entirely and remove its `mod` declaration from `lib.rs`.

The file contained a single constant, `PEER_IDS` (a 40-element array of libp2p peer-id strings annotated with `#[allow(dead_code)]`), which has never been referenced anywhere in the codebase.

---
_Generated by [Claude Code](https://claude.ai/code/session_01G59fSPfUm7ZD91gXqN4GJW)_


asaf-sw:
**This was a mistake on Claude's part, here's the reasoning:**

sequencer-devops references files produced by the apollo_deployments crate's resources/ directory across ~235 YAML overlay files (e.g., committer.yaml, core.yaml, gateway.yaml, l1.yaml, mempool.yaml, sierra-compiler.yaml). These point at  
generated JSON templates like replacer_deployment_*.json and testing_secrets.json by path — not at [addresses.rs](http://addresses.rs) or [lib.rs](http://lib.rs) directly, and not as a Rust dependency.

PR #14396, opened 2026-06-05
starknet_transaction_prover: retry storage-proof RPC fetches with backoff

l-henri:
## Problem

Proving a virtual block fetches a storage proof per 100-key chunk of touched state via `RpcStorageProofsProvider::fetch_single_proof` (`crates/starknet_transaction_prover/src/running/storage_proofs.rs`). Against a public RPC node, a burst of these gets rate-limited. The call is a bare `get_storage_proof().await?` with **no retry**, so a single HTTP 429/5xx aborts a prove that has already run for minutes.

The failure is also opaque: a throttling node returns a non-JSON body, which `starknet-rs` fails to deserialize and surfaces as `ProviderError::Other(..)` carrying `expected value at line 1 column 1` — not a structured JSON-RPC error.

Field data: a controlled replay of 150 small requests against a public node returned 137× HTTP 429; one anywhere in input collection killed the whole prove.

## Fix

Retry transient `get_storage_proof` failures with exponential backoff:

- 5 attempts, 0.5s → 8s (doubling, capped), configurable via `STARKNET_PROVER_RPC_MAX_RETRIES`.
- `is_retryable_rpc_error()` retries rate-limit / 5xx / transport / non-JSON-parse failures.
- **Deterministic structured JSON-RPC errors (block-not-found, etc.) are explicitly NOT retried** — detected via the same `JsonRpcClientError::JsonRpcError` downcast already used in `impl From<ProviderError> for ProofProviderError`, so real errors still fail fast.
- A `tracing::warn!` is emitted per retry with attempt / backoff / error context.

No new dependencies (uses `tokio::time` + `tracing`, both already in the crate).

## Testing

- `cargo check -p starknet_transaction_prover` passes. (Compiled on the `PRIVACY-0.14.2-RC.6` line, where this was found; main's error model — the `From<ProviderError>` downcast — and deps are identical, and the patch applies to `main` unchanged.)
- Not yet exercised against a live throttling node; backoff behavior under real 429s is unverified.

### DIFF
--- a/crates/starknet_transaction_prover/src/running/storage_proofs.rs
+++ b/crates/starknet_transaction_prover/src/running/storage_proofs.rs
@@ -41,7 +41,8 @@ use crate::running::virtual_block_executor::VirtualBlockExecutionData;
 /// `class_hashes.len() + contract_addresses.len() + total_storage_keys`.
 const MAX_KEYS_PER_REQUEST: usize = 100;
 
-/// Default number of attempts for a single `get_storage_proof` call before giving up.
+/// Default number of retries for a single `get_storage_proof` call, *on top of* the
+/// initial attempt — so the default `5` allows up to 6 calls (1 + 5) before giving up.
 /// Proving a virtual block fetches a storage proof per chunk of touched state; a public
 /// node will rate-limit a burst of these. Without retry, a single transient failure
 /// anywhere in input collection aborts a multi-minute prove. Overridable via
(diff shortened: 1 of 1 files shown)


l-henri:
Henri writing, not an agent.
This is obviously an automated PR. It does solve a problem I'm facing. I figured I'd open the PR in case it's useful to others. I'm happy to discuss this further with maintainers and do what's needed to get the PR in, if it's useful.

l-henri:
Update: addressed the Bugbot review and verified the crate locally.

**Review fixes (3 commits on top of the original):**
- `1ddef22` — also retry rate-limits that arrive as a *structured* JSON-RPC error (e.g. "too many requests"), not just as a non-JSON HTTP body. Addresses Bugbot's "JSON-RPC rate limits not retried". Also corrected the `STARKNET_PROVER_RPC_MAX_RETRIES` docs (it's retries on top of the initial attempt: default 5 → up to 6 calls, ~15.5s of backoff).
- `b2a5e5e` — tightened the rate-limit match to unambiguous wording only (`too many requests`, `rate limit`, `throttled`), after Bugbot flagged that loose tokens like `limit exceeded` / `429` would wrongly retry deterministic errors (gas/query-limit messages, block numbers in text).
- `64069a6` — fixed `clippy::as_conversions` on the backoff log line (`delay.as_millis() as u64` → `u64::try_from(...).unwrap_or(u64::MAX)`), which was failing `code_style`.

`Cursor Bugbot` and `code_style` are green. Verified locally: `cargo clippy -p starknet_transaction_prover --lib -- -D warnings` is clean, and the crate's tests pass (`starknet_transaction_prover_ci`: 67 passed, 0 failed).

The remaining red checks (`run-tests`, `run-integration-tests`, `starknet_transaction_prover_ci`, `merge-gatekeeper`) all fail at the `google-github-actions/auth` step — secrets aren't passed to fork-triggered workflows — not on anything in the diff. Happy to rebase or have a maintainer re-run them in the upstream context, whatever's easiest to get this mergeable.

PR #7315, opened 2025-06-16
feat(apollo_signature_manager): define a local key store service variant

elintul:
This is supposed to be used in all test types (unit through
system/integration), not in prod.

In the effort of prep.-ing the service as a node component, and since
design is still missing re how would this service be built with a remote
KMS (and also no adequate infra. support of dependency injection yet) --
I chose to temporarily use this (local KMS) variant across the node.

PR #14472, opened 2026-06-13
apollo_signature_manager: delete dead code (unsure, review carefully)

asaf-sw:
> [!CAUTION]
> REVIEW WITH CARE! THIS PR REQUIRES CAREFUL HUMAN REVIEW.
> If you find this to be a false positive **comment in detail why this code should be kept** and **close the PR**.

## What this removes

- The unused private associated function `LocalKeyStore::_new`.
- Its now-unused `starknet_crypto::get_public_key` import.
- The now-unused `starknet-crypto` dependency (Cargo.toml + Cargo.lock).

## Why it appears dead

- `_new` is the real-private-key constructor for `LocalKeyStore` — it derives the public key from a private key.
- It has **zero callers** anywhere in this workspace or in either sibling repo (`starkware-industries/sequencer-devops`, `starkware-industries/starkware`).
- It is the only user of the `get_public_key` import, which in turn is the crate's only use of the `starknet-crypto` dependency (cargo-machete flagged `starknet-crypto` as unused once `_new` was removed).
- Every construction of `LocalKeyStore` — including the production path `LocalKeyStoreSignatureManager::new` in `lib.rs` — goes through `LocalKeyStore::new_for_testing()` instead.
- The `_` name prefix suppresses the rustc `dead_code` lint, so the compiler does not flag it, but the function is genuinely unreachable.

## Verification (all pass after removal)

- `cargo build -p apollo_signature_manager` — zero `dead_code`/`unused` warnings.
- `cargo build -p apollo_signature_manager --tests` — zero warnings.
- `cargo clippy -p apollo_signature_manager --all-targets` — clean.
- `cargo machete` — clean (the dep removal addresses the initial `code_style` failure).
- `SEED=0 cargo test -p apollo_signature_manager` — passes.

## What a human must verify

`_new` looks like **intentional scaffolding**. Per PR #7315, the `LocalKeyStore` variant was introduced as a temporary stand-in while the remote-KMS design is still pending, with `new_for_testing()` used "across the node" in the interim. `_new` may be deliberately kept to wire up real private-key loading later. **Please confirm whether `_new` is imminently needed for that work** — if so, keep it (and comment why) and close this PR.

> [!CAUTION]
> REVIEW WITH CARE! THIS PR REQUIRES CAREFUL HUMAN REVIEW.
> If you find this to be a false positive **comment in detail why this code should be kept** and **close the PR**.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2554,7 +2554,6 @@ dependencies = [
  "expect-test",
  "hex",
  "starknet-core",
- "starknet-crypto 0.8.1",
  "starknet-types-core",
  "starknet_api",
  "thiserror 1.0.69",
--- a/crates/apollo_signature_manager/Cargo.toml
+++ b/crates/apollo_signature_manager/Cargo.toml
@@ -18,7 +18,6 @@ async-trait.workspace = true
 blake2.workspace = true
 digest.workspace = true
 starknet-core.workspace = true
-starknet-crypto.workspace = true
 starknet-types-core.workspace = true
 starknet_api.workspace = true
 thiserror.workspace = true
--- a/crates/apollo_signature_manager/src/signature_manager.rs
+++ b/crates/apollo_signature_manager/src/signature_manager.rs
@@ -19,7 +19,6 @@ use starknet_api::crypto::utils::{
 };
 use starknet_core::crypto::{ecdsa_sign, ecdsa_verify, EcdsaVerifyError};
 use starknet_core::types::Felt;
-use starknet_crypto::get_public_key;
 use thiserror::Error;
 
 use crate::blake_utils::blake2s_to_felt;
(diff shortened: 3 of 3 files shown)


noamsp-starkware (changes requested):
Thanks for the careful flagging — correct read of the facts, but I'm going to keep _new and close the PR. Detailed why:

_new is intentional scaffolding, not accidental dead code. LocalKeyStore was introduced in #7315 as a temporary stand-in while the production/remote-KMS keystore design is still pending. Today, the production path (LocalKeyStoreSignatureManager::new → LocalKeyStore::new_for_testing()) signs with a hardcoded test private key. _new is the only constructor that builds the keystore from a real private key (deriving the pubkey via get_public_key), so it's exactly the building block needed to close that gap when production key loading lands — see the open TODO(Elin) in lib.rs.

The _ prefix is a deliberate signal that the function is intentionally unused for now (it suppresses the dead_code lint on purpose). Keeping it costs only the small starknet-crypto dependency; removing it would drop the documented real-key construction path.

I confirmed that this code is going to be used and just isn't wired up yet, so we're retaining it. Closing as a false positive.

> @noamsp-starkware reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14472#-:-OwIOLlH5zsoTecGt0i0:b-1a10v7)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on asaf-sw).

PR #14493, opened 2026-06-15
apollo_consensus_orchestrator: fix inverted round comparison discarding queued proposals

matanl-starkware:
## Summary

Fixes **H-20** from the Sequencer Security Report (June 2026): *Backwards Comparison of Queued Proposals' Rounds Discards Future Proposals and Permanently Halts Consensus Liveness*.

In `SequencerConsensusContext::set_height_and_round`, the comparison between `self.current_round` and a queued proposal's round (`entry.key()`) was inverted:

- `Ordering::Less` (queued proposal is for a **future** round) → deleted the proposal instead of keeping it.
- `Ordering::Greater` (queued proposal is for a **past** round) → returned `Ok(())` early instead of cleaning it up.

The early-return path is the dangerous one: a stale past-round entry sits at the front of the `BTreeMap` and poisons the queue. Every subsequent round advancement hits it and returns early, so the current/future-round proposals are never validated — a permanent consensus liveness halt for the height.

## Fix

Correct the match arms:
- `Greater` (past round) → `entry.remove()` and keep scanning.
- `Equal` (current round) → take and process.
- `Less` (future round) → `break`, preserving it for later.

## Test

Added `skipped_round_does_not_block_future_proposal`: queues proposals for rounds 1 and 2 while at round 0, advances directly to round 2 (round 1 skipped), and asserts the stale round-1 proposal is dropped while the round-2 proposal is still validated.

Verified against the same test binary: the buggy version **hangs** (liveness halt — `fin_receiver` never resolves, killed by timeout, exit 124); the fixed version passes in 0.01s.

### DIFF
--- a/crates/apollo_consensus_orchestrator/src/sequencer_consensus_context.rs
+++ b/crates/apollo_consensus_orchestrator/src/sequencer_consensus_context.rs
@@ -1090,14 +1090,17 @@ impl ConsensusContext for SequencerConsensusContext {
         let mut to_process = None;
         while let Some(entry) = self.queued_proposals.first_entry() {
             match self.current_round.cmp(entry.key()) {
-                std::cmp::Ordering::Less => {
+                // The queued proposal is for a past round; drop it and keep scanning.
+                std::cmp::Ordering::Greater => {
                     entry.remove();
                 }
+                // The queued proposal is for the current round; take it and stop.
                 std::cmp::Ordering::Equal => {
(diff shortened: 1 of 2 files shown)


dafnamatsry (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14493)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

PR #14501, opened 2026-06-15
apollo_consensus_orchestrator: add per-attempt timeout to cende recorder client

matanl-starkware:
## Problem

On Mainnet, isolated round>0 events trace to the proposer aborting its proposal with `CendeWriteError("Writing blob to Aerospike didn't return in time")`, which leaves validators with a truncated stream (`ProposalPartFailed`) and forces a round change.

Root cause: the cende recorder client is built with `reqwest::Client::new()` — **no per-request timeout**. `RetryTransientMiddleware` only retries attempts that *return* a transient error (connection refused, 5xx, …); it cannot preempt a request that is slow/hung (connection open, no response). So a single stalled request blocks until the proposer's build deadline, surfacing as the outer `tokio::time::timeout` firing — `CendeWriteError("…didn't return in time")` — with **zero retries actually attempted**. The retry policy only protected against the case it almost never needs (fast-failing errors), not the real failure mode (a slow recorder).

## Fix

Bound each attempt with `reqwest`'s per-request `.timeout()`, derived from the existing `max_retry_interval_ms` (no new config). A hang now fails fast as a timeout error, which `RetryTransientMiddleware` classifies as transient and retries — with room for multiple attempts within `max_retry_duration_secs`.

## Testing

- `cargo build -p apollo_consensus_orchestrator` ✅
- Note: the `testing`-feature test build is already broken on `main-v0.14.3` (`CentralTransactionExecutionInfo` lacks `Deserialize`/`PartialEq`), unrelated to this change, so the cende tests could not be run.

### DIFF
--- a/crates/apollo_consensus_orchestrator/src/cende/mod.rs
+++ b/crates/apollo_consensus_orchestrator/src/cende/mod.rs
@@ -152,9 +152,20 @@ impl CendeAmbassador {
                 .recorder_url
                 .join(RECORDER_GET_LATEST_RECEIVED_BLOCK_PATH)
                 .expect("Failed to construct get latest received block URL"),
-            client: ClientBuilder::new(reqwest::Client::new())
-                .with(RetryTransientMiddleware::new_with_policy(retry_policy))
-                .build(),
+            // Bound each attempt by the max retry interval. Without a per-attempt timeout
+            // `RetryTransientMiddleware` only retries attempts that *return* a transient error, so
+            // a request that hangs against a slow recorder would block until the build deadline
+            // instead of being retried. Capping an attempt at the backoff's upper bound turns such
+            // a hang into a timeout error the retry policy can act on, while leaving room for
+            // multiple attempts within `max_retry_duration_secs`.
+            client: ClientBuilder::new(
+                reqwest::Client::builder()
+                    .timeout(cende_config.max_retry_interval_ms)
+                    .build()
+                    .expect("Failed to build cende recorder client"),
+            )
+            .with(RetryTransientMiddleware::new_with_policy(retry_policy))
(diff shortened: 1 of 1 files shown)


dafnamatsry (changes requested):
> @dafnamatsry reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14501)* status: all files reviewed, 1 unresolved discussion (waiting on matanl-starkware).

___
*[`crates/apollo_consensus_orchestrator/src/cende/mod.rs` line 160 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14501#-OvA9QUe3d5TnuAv0A4r:-OvA9QUe3d5TnuAv0A4s:b3c90j7) ([raw file](https://github.com/starkware-libs/sequencer/blob/2ee970b28bc1390298bf30ae7cefae0ab463f1d1/crates/apollo_consensus_orchestrator/src/cende/mod.rs#L160)):*
> ```XML
>             // instead of being retried. Capping an attempt at the backoff's upper bound turns such
>             // a hang into a timeout error the retry policy can act on, while leaving room for
>             // multiple attempts within `max_retry_duration_secs`.
> ```

The recorder can be slow when writing a big blob, and retrying the same block again will fail, since another request is in progress. 
Do we see this happening that we need this change?

dafnamatsry (approved):
> @dafnamatsry resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14501)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

PR #14519, opened 2026-06-16
apollo_node_config: enforce config_manager runs locally (no remote client)

matanl-starkware:
## What

Reject any node config where the **config manager** is set to a remote-capable execution mode (`Remote` or `LocalExecutionWithRemoteEnabled`), in `SequencerNodeConfig::cross_member_validations`. Only `LocalExecutionWithRemoteDisabled` and `Disabled` are now accepted.

## Why (Security report H-5)

`MempoolCommunicationWrapper::handle_request` calls `update_dynamic_config()` on **every** mempool request, which does:

```rust
self.config_manager_client.get_mempool_dynamic_config().await
    .expect("Should be able to get mempool dynamic config");
```

The client is local or remote depending on `config.components.config_manager.execution_mode` (via the `create_client!` macro). If it were a **remote** (HTTP) client, any transient network failure / timeout / restart would make the `.expect()` panic and crash the mempool — a full DoS (report H-5).

The config manager is a **local infrastructure component**: every node runs its own instance, and consumers reach it over an in-process channel. All deployment presets already set it to `local_with_remote_disabled()`, and `components.rs` already `panic!`s on remote modes — but that's a deep startup panic and config validation never caught the misconfiguration.

This change closes the gap at the validation layer:
- Guarantees the `create_client!` macro can only ever take the local arm → the config-manager client is always an in-process channel → **H-5's remote-RPC crash path is unreachable**.
- Converts the latent deep panic into an early, descriptive `ConfigError` at config load (validation runs on the startup path via `config_utils.rs`).
- Stays consistent with existing `components.rs` semantics (allows `LocalExecutionWithRemoteDisabled` / `Disabled`).

## Tests

Added `config_manager_remote_is_rejected` and `config_manager_local_with_remote_enabled_is_rejected` in `config_test.rs`. Both pass; crate builds clean and formatted.

### DIFF
--- a/crates/apollo_node_config/src/config_test.rs
+++ b/crates/apollo_node_config/src/config_test.rs
@@ -209,6 +209,51 @@ fn validation_only_with_mempool_p2p_enabled_fails() {
     assert!(format!("{err:?}").contains("mempool_p2p"), "Unexpected error: {err:?}");
 }
 
+// The config manager is a local infrastructure component; running it remotely would make its
+// consumers (e.g. the mempool) reach it over a network RPC that can fail mid-request. Both
+// remote-capable execution modes must be rejected at validation time so the client is always local.
+#[test]
+fn config_manager_remote_is_rejected() {
+    // `config_manager_config` is None so the per-component "set iff running locally" check passes
+    // (remote is not running locally) and we reach the config_manager-specific validation.
+    let config = SequencerNodeConfig {
+        components: ComponentConfig {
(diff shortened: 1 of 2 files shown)


ron-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14519)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

PR #14520, opened 2026-06-16
papyrus_base_layer: re-point live provider on URL cycle so L1 failover works

asaf-sw:
L1 endpoint failover advanced the internal URL index but never rebuilt the live alloy provider, so all subsequent calls kept hitting the failed endpoint. This fix tracks the active URL in a `node_url` field and rebuilds the provider on every cycle.

- Add private `node_url: Sensitive<Url>` field to `EthereumBaseLayerContract`; `get_url()` returns it; `set_provider_url` updates it and rebuilds the provider; `cycle_provider_url` delegates to `set_provider_url`
- Add `test_cycle_provider_url_repoints_live_provider` and `test_cycle_wraps_to_primary_through_full_list` to `base_layer_test.rs`
- Add `test_exhaust_from_non_primary_index_returns_last_error` to `cyclic_base_layer_wrapper_test.rs`
- Fix `anvil_base_layer.rs` struct literal broken by the now-private field (replaced with `new_with_provider` call)

### DIFF
--- a/crates/papyrus_base_layer/src/base_layer_test.rs
+++ b/crates/papyrus_base_layer/src/base_layer_test.rs
@@ -4,6 +4,7 @@ use alloy::providers::mock::Asserter;
 use alloy::providers::{Provider, ProviderBuilder};
 use alloy::rpc::types::{Block, BlockTransactions, Header as AlloyRpcHeader};
 use pretty_assertions::assert_eq;
+use url::Url;
 
 use crate::eth_events::{create_l1_event_data, felt_max_u256, u256_exceeds_felt};
 use crate::ethereum_base_layer_contract::{
@@ -100,6 +101,37 @@ async fn get_gas_price_and_timestamps() {
     // Roughly e ** (BLOB_GAS / eip7691::BLOB_GASPRICE_UPDATE_FRACTION_PECTRA)
     let expected_pectra_blob_calc = 7;
     assert_eq!(header.blob_fee, expected_pectra_blob_calc);
+}
+
+#[tokio::test]
+async fn test_cycle_wraps_to_primary_through_full_list() {
+    let primary_url = Url::parse("http://primary-endpoint.test/").unwrap();
+    let secondary_url = Url::parse("http://secondary-endpoint.test/").unwrap();
+    let tertiary_url = Url::parse("http://tertiary-endpoint.test/").unwrap();
+    let config = EthereumBaseLayerConfig {
+        ordered_l1_endpoint_urls: vec![
+            primary_url.clone().into(),
+            secondary_url.clone().into(),
+            tertiary_url.clone().into(),
+        ],
+        ..Default::default()
+    };
+    let mut base_layer = EthereumBaseLayerContract::new(config);
+
+    // The live provider starts on the primary (first) endpoint.
+    assert_eq!(base_layer.get_url().await.unwrap().expose_secret(), primary_url);
+
+    // First cycle moves to the secondary endpoint.
+    base_layer.cycle_provider_url().await.unwrap();
+    assert_eq!(base_layer.get_url().await.unwrap().expose_secret(), secondary_url);
+
+    // Second cycle moves to the tertiary endpoint.
+    base_layer.cycle_provider_url().await.unwrap();
+    assert_eq!(base_layer.get_url().await.unwrap().expose_secret(), tertiary_url);
+
+    // Third cycle wraps back to the primary endpoint.
+    base_layer.cycle_provider_url().await.unwrap();
+    assert_eq!(base_layer.get_url().await.unwrap().expose_secret(), primary_url);
 }
 
 #[test]
--- a/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
+++ b/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
@@ -344,3 +344,56 @@ async fn pass_through_cycle_provider_url() {
     let result = wrapper.cycle_provider_url().await;
     assert!(result.is_ok());
 }
+
+// Verifies that when an operation begins on a non-primary endpoint and every attempt fails, the
+// wrapper cycles through the entire URL list exactly once and returns the last error. The operation
+// starts on the tertiary endpoint, and cycling advances tertiary -> primary -> secondary ->
+// tertiary, at which point the wrapper detects it wrapped back to the start URL and returns the
+// error.
+#[tokio::test]
+async fn test_exhaust_from_non_primary_index_returns_last_error() {
+    // Setup: model a 3-URL list where the operation starts on the tertiary (non-primary) endpoint.
+    let primary_url = Sensitive::new(Url::parse("http://primary_endpoint").unwrap())
+        .with_redactor(to_safe_string);
+    let secondary_url = Sensitive::new(Url::parse("http://secondary_endpoint").unwrap())
+        .with_redactor(to_safe_string);
+    let tertiary_url = Sensitive::new(Url::parse("http://tertiary_endpoint").unwrap())
+        .with_redactor(to_safe_string);
+
+    // The wrapper calls get_url once for start_url, then per failed attempt calls get_url (current)
+    // + cycle_provider_url + get_url (new). With a tertiary start and all three attempts failing,
+    // the get_url return sequence is: start=tertiary; then current/new pairs as the index advances
+    // tertiary -> primary -> secondary -> tertiary.
+    let url_sequence = [
+        tertiary_url.clone(),
+        tertiary_url.clone(),
+        primary_url.clone(),
+        primary_url,
+        secondary_url.clone(),
+        secondary_url,
+        tertiary_url,
+    ];
+    const NUM_ATTEMPTS: usize = 3;
+    let num_get_url_calls = url_sequence.len();
+    // Each mocked call pops the next queued return, so the return order is explicit per call.
+    let mut get_url_returns = url_sequence.into_iter();
+    let mut error_returns = (0..NUM_ATTEMPTS).map(MockError::Numbered);
+
+    let mut base_layer = MockBaseLayerContract::new();
+    base_layer
+        .expect_get_url()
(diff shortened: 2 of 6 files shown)


asaf-sw:
* **#14521** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14521?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14520** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14520?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14520?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/papyrus_base_layer/src/base_layer_test.rs
```diff
@@ -102,6 +103,66 @@ async fn get_gas_price_and_timestamps() {
     assert_eq!(header.blob_fee, expected_pectra_blob_calc);
 }
 
+// Offline unit test (never issues a request): verifies that cycling the provider URL re-points the
```
ShahakShama: This "offline unit test" comment isn't needed IMO — it's clear from the test itself. Same for the comment on the test below it.

crates/papyrus_base_layer/src/base_layer_test.rs
```diff
@@ -102,6 +103,66 @@ async fn get_gas_price_and_timestamps() {
...
+// *live* provider, not merely an internal index. A mock can't cover this because the behavior under
+// test lives inside `EthereumBaseLayerContract::cycle_provider_url` itself, so we use the real
+// contract with non-routable fake URLs and only inspect URL-routing state.
+#[tokio::test]
+async fn test_cycle_provider_url_repoints_live_provider() {
```
ShahakShama: This test and `test_cycle_wraps_to_primary_through_full_list` below are nearly identical — AFAICS the only difference is the number of endpoints (2 vs 3). Please merge them into a single test, keeping whichever assertion is stronger.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
```diff
@@ -344,3 +344,51 @@ async fn pass_through_cycle_provider_url() {
...
+    // the get_url return sequence is: start=tertiary; then current/new pairs as the index advances
+    // tertiary -> primary -> secondary -> tertiary.
+    let url_sequence = [
+        tertiary_url.clone(),
+        tertiary_url,
+        primary_url.clone(),
+        primary_url,
+        secondary_url.clone(),
+        secondary_url,
+        Sensitive::new(Url::parse("http://tertiary_endpoint").unwrap())
```
ShahakShama: Why rebuild the tertiary URL here instead of reusing `tertiary_url`? You could `.clone()` it at its first use a few lines up and move the original into this slot, avoiding the duplicated `Url::parse` + redactor setup.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
```diff
@@ -344,3 +344,51 @@ async fn pass_through_cycle_provider_url() {
...
+    let num_url_calls_made = Arc::new(AtomicUsize::new(0));
+    let num_url_calls_made_clone = num_url_calls_made.clone();
+
+    let mut base_layer = MockBaseLayerContract::new();
+    base_layer.expect_get_url().times(url_sequence.len()).returning(move || {
+        let url_index = num_url_calls_made_clone.fetch_add(1, Ordering::Relaxed);
+        Ok(url_sequence[url_index].clone())
+    });
+    // Every attempt fails: one per endpoint in the 3-URL list.
+    base_layer.expect_get_proved_block_at().times(3).returning(move |_| Err(MockError::MockError));
```
ShahakShama: Since the wrapper's contract is to return the *last* error after cycling back to the start URL, returning the same `MockError::MockError` on every attempt means the test can't actually verify *which* error is returned. Consider returning a distinct error per attempt so the assertion proves it's the last one. A TODO referencing this would be fine for now if you'd rather not do it in this PR.

crates/papyrus_base_layer/src/base_layer_test.rs
asaf-sw: Done.

crates/papyrus_base_layer/src/base_layer_test.rs
asaf-sw: Done.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
asaf-sw: Done.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
asaf-sw: Fixed (no TODO).

crates/papyrus_base_layer/src/base_layer_test.rs
asaf-sw: Removed both `// Offline unit test …` comment blocks.

crates/papyrus_base_layer/src/base_layer_test.rs
asaf-sw: Merged — removed `test_cycle_provider_url_repoints_live_provider` and kept `test_cycle_wraps_to_primary_through_full_list` (3 endpoints, a superset of the 2-endpoint case); it still asserts the initial endpoint is the primary before cycling.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
asaf-sw: Done — `tertiary_url` is now bound once and cloned; removed the duplicate `Url::parse`.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
asaf-sw: Done — added a `MockError::Numbered(usize)` variant so each attempt returns a distinct error; the test now asserts the wrapper returns the last one (`Numbered(2)`).

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
```diff
@@ -344,3 +344,55 @@ async fn pass_through_cycle_provider_url() {
...
+        secondary_url,
+        tertiary_url,
+    ];
+    let num_url_calls_made = Arc::new(AtomicUsize::new(0));
+    let num_url_calls_made_clone = num_url_calls_made.clone();
+    let num_attempt_calls_made = Arc::new(AtomicUsize::new(0));
+    let num_attempt_calls_made_clone = num_attempt_calls_made.clone();
+
+    let mut base_layer = MockBaseLayerContract::new();
+    base_layer.expect_get_url().times(url_sequence.len()).returning(move || {
```
ShahakShama: Instead of hand-rolling an `AtomicUsize` + indexing into `url_sequence`, consider driving the return values with mockall's sequencing — e.g. chain per-call expectations with `mockall::Sequence`, or queue the returns so each call pops the next URL. That drops the `num_url_calls_made` atomic and the manual index bookkeeping, and makes the call order an explicit part of the expectation rather than something reconstructed from a shared counter.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
```diff
@@ -344,3 +344,55 @@ async fn pass_through_cycle_provider_url() {
...
+        let url_index = num_url_calls_made_clone.fetch_add(1, Ordering::Relaxed);
+        Ok(url_sequence[url_index].clone())
+    });
+    // Every attempt fails with a distinct numbered error so we can prove the last one is returned.
+    base_layer.expect_get_proved_block_at().times(3).returning(move |_| {
```
ShahakShama: The attempt count `3` is repeated here and on `expect_cycle_provider_url().times(3)` below, and the final assertion hard-codes `Numbered(2)`. Extract it to a named constant (e.g. `const NUM_ATTEMPTS: usize = 3;`) and write the assertion as `Numbered(NUM_ATTEMPTS - 1)` — that ties the three together and documents that the last attempt's index is `count - 1`, so the test stays correct if the attempt count ever changes.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
asaf-sw: Done — dropped both `AtomicUsize` counters; `get_url` and `get_proved_block_at` now pop from queued iterators (`url_sequence.into_iter()` and `(0..NUM_ATTEMPTS).map(MockError::Numbered)`), so each call's return is explicit per call. I went with the queue rather than `mockall::Sequence`: the essential coverage (each endpoint tried once, last error returned) is already pinned by `times(...)` + the distinct-error assertion, so enforcing cross-method call order would mainly add brittleness without protecting a real contract.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
asaf-sw: Done — extracted `const NUM_ATTEMPTS: usize = 3;` and used it for both `times(NUM_ATTEMPTS)` calls and the assertion `Numbered(NUM_ATTEMPTS - 1)`.

ShahakShama (approved):
:lgtm:

> @ShahakShama reviewed 6 files and all commit messages, made 1 comment, and resolved 6 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14520#-:-OvPmGo25v5JihBXNG1G:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on asaf-sw).

PR #14521, opened 2026-06-16
papyrus_base_layer: add periodic retry of primary endpoint after failover

asaf-sw:
## Summary
After `CyclicBaseLayerWrapper` fails over from the primary (first) L1 endpoint to a backup, it stays on that backup. This adds an opportunistic, time-based retry of the primary so a recovered primary gets used again.

## Behavior
- On the next L1 access once `retry_primary_interval_seconds` have elapsed since leaving the primary, the live provider is reverted to the primary endpoint.
- The clock measures time since leaving the primary — cycling between backups does not delay it.
- No-op while already on the primary; opportunistic (no background task and no health probe before switching).

## Config
- New `base_layer_config.retry_primary_interval_seconds` (default `60`); node config schema and deployment app-configs regenerated.

## Tests
Offline unit tests covering the retry trigger (interval elapsed / not elapsed / already on primary), error propagation, deterministic timing via `tokio::time::pause`, and that backup→backup cycling does not delay the primary retry.

### DIFF
--- a/crates/apollo_base_layer_tests/src/anvil_base_layer.rs
+++ b/crates/apollo_base_layer_tests/src/anvil_base_layer.rs
@@ -267,6 +267,15 @@ impl BaseLayerContract for AnvilBaseLayer {
     async fn cycle_provider_url(&mut self) -> Result<(), Self::Error> {
         unimplemented!("Anvil base layer is tied to a an Anvil server, url is fixed.")
     }
+
+    async fn reset_provider_url_to_primary(&mut self) -> Result<(), Self::Error> {
+        unimplemented!("Anvil base layer is tied to a an Anvil server, url is fixed.")
+    }
+
+    // Anvil is tied to a single fixed URL, so it is always on its only (primary) endpoint.
+    async fn is_at_primary(&self) -> Result<bool, Self::Error> {
+        Ok(true)
+    }
 }
 
 /// Converts a given [L1 handler transaction](starknet_api::transaction::L1HandlerTransaction)
--- a/crates/apollo_node/resources/config_schema.json
+++ b/crates/apollo_node/resources/config_schema.json
@@ -24,6 +24,11 @@
     "privacy": "Private",
     "value": "https://mainnet.infura.io/v3/YOUR_INFURA_API_KEY"
   },
+  "base_layer_config.retry_primary_interval_seconds": {
+    "description": "The interval (seconds) after which the next base-layer access retries the primary (first) endpoint.",
+    "privacy": "Public",
+    "value": 60
+  },
   "base_layer_config.starknet_contract_address": {
     "description": "Starknet contract address in ethereum.",
     "privacy": "Public",
--- a/crates/papyrus_base_layer/src/base_layer_test.rs
+++ b/crates/papyrus_base_layer/src/base_layer_test.rs
@@ -134,6 +134,35 @@ async fn test_cycle_wraps_to_primary_through_full_list() {
     assert_eq!(base_layer.get_url().await.unwrap().expose_secret(), primary_url);
 }
 
+#[tokio::test]
+async fn test_reset_provider_url_to_primary_repoints_live_provider() {
+    let primary_url = Url::parse("http://primary-endpoint.test/").unwrap();
+    let secondary_url = Url::parse("http://secondary-endpoint.test/").unwrap();
+    let tertiary_url = Url::parse("http://tertiary-endpoint.test/").unwrap();
+    let config = EthereumBaseLayerConfig {
+        ordered_l1_endpoint_urls: vec![
+            primary_url.clone().into(),
+            secondary_url.clone().into(),
+            tertiary_url.clone().into(),
+        ],
+        ..Default::default()
+    };
+    let mut base_layer = EthereumBaseLayerContract::new(config);
+
+    // Cycle twice to land on the tertiary endpoint.
+    base_layer.cycle_provider_url().await.unwrap();
+    base_layer.cycle_provider_url().await.unwrap();
+    assert_eq!(base_layer.get_url().await.unwrap().expose_secret(), tertiary_url);
+
+    // Reset to primary.
+    base_layer.reset_provider_url_to_primary().await.unwrap();
+    assert_eq!(base_layer.get_url().await.unwrap().expose_secret(), primary_url);
+
+    // Calling reset again when already on primary is a no-op.
+    base_layer.reset_provider_url_to_primary().await.unwrap();
+    assert_eq!(base_layer.get_url().await.unwrap().expose_secret(), primary_url);
+}
+
 #[test]
 fn create_l1_event_data_rejects_out_of_range_inputs() {
     let oversized = felt_max_u256() + U256::from(1_u8);
--- a/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs
+++ b/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs
@@ -1,4 +1,5 @@
 use std::ops::RangeInclusive;
+use std::time::Duration;
 
 use apollo_config::secrets::Sensitive;
 use async_trait::async_trait;
@@ -15,11 +16,26 @@ pub mod cyclic_base_layer_wrapper_test;
 #[derive(Debug)]
 pub struct CyclicBaseLayerWrapper<B: BaseLayerContract + Send + Sync> {
     base_layer: B,
+    retry_primary_interval: Duration,
+    last_primary_retry: tokio::time::Instant,
 }
 
 impl<B: BaseLayerContract + Send + Sync> CyclicBaseLayerWrapper<B> {
-    pub fn new(base_layer: B) -> Self {
-        Self { base_layer }
+    pub fn new(base_layer: B, retry_primary_interval: Duration) -> Self {
+        Self { base_layer, retry_primary_interval, last_primary_retry: tokio::time::Instant::now() }
+    }
+
+    // Retries the primary endpoint once the interval has elapsed since we left it. Does nothing
+    // while already on the primary, so the timer is untouched until a failover moves us off it.
+    async fn retry_primary_if_due(&mut self) -> Result<(), B::Error> {
+        if self.base_layer.is_at_primary().await? {
+            return Ok(());
+        }
+        if self.last_primary_retry.elapsed() >= self.retry_primary_interval {
+            self.last_primary_retry = tokio::time::Instant::now();
+            self.base_layer.reset_provider_url_to_primary().await?;
+        }
+        Ok(())
     }
 
     // Check the result of a function call to the base layer. If it fails, cycle the URL and signal
@@ -38,11 +54,21 @@ impl<B: BaseLayerContract + Send + Sync> CyclicBaseLayerWrapper<B> {
         let Ok(current_url) = current_url_result else {
             return Some(Err(current_url_result.expect_err("result is checked at let-else")));
         };
+        // Record whether we are about to leave the primary, before cycling away from it.
+        let is_at_primary_result = self.base_layer.is_at_primary().await;
+        let Ok(was_at_primary) = is_at_primary_result else {
+            return Some(Err(is_at_primary_result.expect_err("result is checked at let-else")));
+        };
         // Otherwise, cycle the URL so we can try again. Return error in case it fails to cycle.
         let cycle_url_result = self.base_layer.cycle_provider_url().await;
         let Ok(()) = cycle_url_result else {
             return Some(Err(cycle_url_result.expect_err("result is checked at let-else")));
         };
+        // Restart the retry-primary clock only when this failover leaves the primary, so the wait
+        // is measured from when we left it; cycling between backups must not push the retry out.
+        if was_at_primary {
+            self.last_primary_retry = tokio::time::Instant::now();
+        }
         // Get the new URL (return error in case it fails to get it).
         let new_url_result = self.base_layer.get_url().await;
         let Ok(new_url) = new_url_result else {
@@ -69,6 +95,7 @@ impl<B: BaseLayerContract + Send + Sync> BaseLayerContract for CyclicBaseLayerWr
         &mut self,
         l1_block: L1BlockNumber,
     ) -> Result<BlockHashAndNumber, Self::Error> {
+        self.retry_primary_if_due().await?;
         let start_url = self.base_layer.get_url().await?;
         loop {
             let result = self.base_layer.get_proved_block_at(l1_block).await;
@@ -79,6 +106,7 @@ impl<B: BaseLayerContract + Send + Sync> BaseLayerContract for CyclicBaseLayerWr
     }
 
     async fn latest_l1_block_number(&mut self) -> Result<L1BlockNumber, Self::Error> {
+        self.retry_primary_if_due().await?;
         let start_url = self.base_layer.get_url().await?;
         loop {
             let result = self.base_layer.latest_l1_block_number().await;
@@ -92,6 +120,7 @@ impl<B: BaseLayerContract + Send + Sync> BaseLayerContract for CyclicBaseLayerWr
         &mut self,
         block_number: L1BlockNumber,
     ) -> Result<Option<L1BlockReference>, Self::Error> {
+        self.retry_primary_if_due().await?;
         let start_url = self.base_layer.get_url().await?;
         loop {
             let result = self.base_layer.l1_block_at(block_number).await;
@@ -106,6 +135,7 @@ impl<B: BaseLayerContract + Send + Sync> BaseLayerContract for CyclicBaseLayerWr
         block_range: RangeInclusive<L1BlockNumber>,
         event_identifiers: &'a [&'a str],
     ) -> Result<Vec<L1Event>, Self::Error> {
+        self.retry_primary_if_due().await?;
         let start_url = self.base_layer.get_url().await?;
         loop {
             let result = self.base_layer.events(block_range.clone(), event_identifiers).await;
@@ -119,6 +149,7 @@ impl<B: BaseLayerContract + Send + Sync> BaseLayerContract for CyclicBaseLayerWr
         &mut self,
         block_number: L1BlockNumber,
     ) -> Result<Option<L1BlockHeader>, Self::Error> {
+        self.retry_primary_if_due().await?;
         let start_url = self.base_layer.get_url().await?;
         loop {
             let result = self.base_layer.get_block_header(block_number).await;
@@ -128,6 +159,8 @@ impl<B: BaseLayerContract + Send + Sync> BaseLayerContract for CyclicBaseLayerWr
         }
     }
 
+    // Takes &self so it cannot cycle or retry endpoints; callers needing resilience use the &mut
+    // self methods.
     async fn get_block_header_immutable(
         &self,
         block_number: L1BlockNumber,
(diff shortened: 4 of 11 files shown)


asaf-sw:
* **#14521** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14521?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14521?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#14520** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14520?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

matanl-starkware (changes requested):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14521)* status: 0 of 10 files reviewed, 2 unresolved discussions (waiting on asaf-sw and ShahakShama).

___
*[`crates/apollo_node/resources/config_schema.json` line 28 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-OvKQvgH4RXJJ23l-F-k:-OvKQvgH4RXJJ23l-F-l:b-q1y6wr) ([raw file](https://github.com/starkware-libs/sequencer/blob/d0a1a60300f077f8c4c016dc1fd9fad425f8f143/crates/apollo_node/resources/config_schema.json#L28)):*
> ```JSON
>   },
>   "base_layer_config.primary_failback_interval_seconds": {
>     "description": "The interval (seconds) after which the contract attempts to revert to the primary (first) endpoint.",
> ```

which contract?



_Code quote:_
```JSON
the contract
```
___
*[`crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs` line 94 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-OvKSHL-Bfrb_XKPGwKR:-OvKSHL-Bfrb_XKPGwKS:b-lc8bs) ([raw file](https://github.com/starkware-libs/sequencer/blob/d0a1a60300f077f8c4c016dc1fd9fad425f8f143/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs#L94)):*
> ```XML
>             }
>         }
>     }
> ```

We should switch to primary iff we've verified the primary endpoint is working.
We would like to avoid triggering alerts or jamming the already-working endpoint.
IMO the check should run asynchronously to the ongoing requests.



_Code quote:_
```XML
    async fn get_proved_block_at(
        &mut self,
        l1_block: L1BlockNumber,
    ) -> Result<BlockHashAndNumber, Self::Error> {
        self.failback_to_primary_if_due().await?;
        let start_url = self.base_layer.get_url().await?;
        loop {
            let result = self.base_layer.get_proved_block_at(l1_block).await;
            if let Some(result) = self.cycle_url_on_error(&start_url, result).await {
                return result; // Could return a success or an error.
            }
        }
    }
```

crates/papyrus_base_layer/src/base_layer_test.rs
```diff
@@ -163,6 +163,38 @@ async fn test_cycle_wraps_to_primary_through_full_list() {
     assert_eq!(base_layer.get_url().await.unwrap().expose_secret(), primary_url);
 }
 
+// Offline unit test (never issues a request): verifies that reset_provider_url_to_primary
```
ShahakShama: Same note as the last PR: this "offline unit test" comment isn't needed IMO — it's clear from the test itself.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
```diff
@@ -56,7 +57,7 @@ async fn cycle_get_proved_block_at(#[case] num_failing_calls: usize) {
         num_url_calls_made_clone2.fetch_add(1, Ordering::Relaxed);
         Ok(())
     });
-    let mut wrapper = CyclicBaseLayerWrapper::new(base_layer);
+    let mut wrapper = CyclicBaseLayerWrapper::new(base_layer, Duration::from_secs(3600));
```
ShahakShama: `Duration::from_secs(3600)` is repeated across all the test cases in this file (~13 times). Please extract it to a shared test constant (e.g. `const TEST_FAILBACK_INTERVAL: Duration = Duration::from_secs(3600);`) so the intent is named and there's one place to change it.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs
```diff
@@ -119,6 +137,7 @@ impl<B: BaseLayerContract + Send + Sync> BaseLayerContract for CyclicBaseLayerWr
         &mut self,
         block_number: L1BlockNumber,
     ) -> Result<Option<L1BlockHeader>, Self::Error> {
+        self.failback_to_primary_if_due().await?;
```
ShahakShama: Every `&mut self` method here now gets `failback_to_primary_if_due()` + the cycle-on-error loop, but `get_block_header_immutable` below delegates straight through with neither. I want to make sure this isn't an actual bug — can you explain why it's safe for this method to skip the failback/cycling? Is it related to the fact that this one is `&self` while the rest are `&mut self`? Once it's clear, please also capture the reason in a one-line comment on the method.

crates/apollo_node/resources/config_schema.json
```diff
@@ -24,6 +24,11 @@
     "privacy": "Private",
     "value": "https://mainnet.infura.io/v3/YOUR_INFURA_API_KEY"
   },
+  "base_layer_config.primary_failback_interval_seconds": {
```
ShahakShama: Naming nit: I'm not sure "failback" is the right word here — maybe "retry"? This name repeats across the source, so if it changes it should change everywhere.

crates/apollo_base_layer_tests/src/anvil_base_layer.rs
```diff
@@ -267,6 +267,10 @@ impl BaseLayerContract for AnvilBaseLayer {
     async fn cycle_provider_url(&mut self) -> Result<(), Self::Error> {
         unimplemented!("Anvil base layer is tied to a an Anvil server, url is fixed.")
     }
+
+    async fn reset_provider_url_to_primary(&mut self) -> Result<(), Self::Error> {
+        Ok(())
```
ShahakShama: The two sibling methods (`set_provider_url`, `cycle_provider_url`) `unimplemented!()` here — should `reset_provider_url_to_primary` do the same instead of silently returning `Ok(())`? If the no-op is intentional because the failback path in `CyclicBaseLayerWrapper` calls this, please state that in a one-line comment.

asaf-sw:
*[`crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs` line 94 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-OvKSHL-Bfrb_XKPGwKR:-OvKoJzR2MaL6Rh4VNpz:b-fa44nk) ([raw file](https://github.com/starkware-libs/sequencer/blob/d0a1a60300f077f8c4c016dc1fd9fad425f8f143/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs#L94)):*

We should switch to primary iff we've verified the primary endpoint is working.
We would like to avoid triggering alerts or jamming the already-working endpoint.
IMO the check should run asynchronously to the ongoing requests.
</blockquote></details>

That would be wasteful and more complex. This solution here is opportunistic, the next time we need to access the L1 source, if enough time has passed we try to revert to the original, if not we cycle as before.

asaf-sw:
*[`crates/apollo_node/resources/config_schema.json` line 28 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-OvKQvgH4RXJJ23l-F-k:-OvKr8wz01IOaDh4B5n5:b-m9v8tv) ([raw file](https://github.com/starkware-libs/sequencer/blob/d0a1a60300f077f8c4c016dc1fd9fad425f8f143/crates/apollo_node/resources/config_schema.json#L28)):*

which contract?
</blockquote></details>

Will fix comment in next push.

crates/papyrus_base_layer/src/base_layer_test.rs
asaf-sw: Removed the comment.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper_test.rs
asaf-sw: Extracted to `const TEST_RETRY_PRIMARY_INTERVAL: Duration = Duration::from_secs(3600);` and replaced all usages.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs
asaf-sw: Not a bug — it takes `&self`, so it cannot cycle/retry endpoints (that needs `&mut self`); it is an internal helper, and callers needing resilience go through the `&mut self` methods. Added a one-line comment saying so.

crates/apollo_node/resources/config_schema.json
asaf-sw: Renamed throughout to `retry_primary_*` (the config is now `retry_primary_interval_seconds`, plus `retry_primary_if_due`, etc.); `failback` no longer appears in the codebase. Also reworded the description to drop the ambiguous "contract" wording (per Matan).

crates/apollo_base_layer_tests/src/anvil_base_layer.rs
asaf-sw: Changed to `unimplemented!()` to match the sibling `set_provider_url`/`cycle_provider_url` methods.

ShahakShama (approved):
:lgtm:

> @ShahakShama reviewed 10 files and all commit messages, made 1 comment, and resolved 5 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-:-OvOsCH01L2byzd8ukfs:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on matanl-starkware).

matanl-starkware (changes requested):
> @matanl-starkware made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14521)* status: all files reviewed, 1 unresolved discussion (waiting on asaf-sw).

___
*[`crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs` line 94 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-OvKSHL-Bfrb_XKPGwKR:-OvPMcHC71M88ZyGlNsG:b-avq8ia) ([raw file](https://github.com/starkware-libs/sequencer/blob/d0a1a60300f077f8c4c016dc1fd9fad425f8f143/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs#L94)):*
<details><summary><i>Previously, asaf-sw wrote…</i></summary><blockquote>

That would be wasteful and more complex. This solution here is opportunistic, the next time we need to access the L1 source, if enough time has passed we try to revert to the original, if not we cycle as before.
</blockquote></details>

Long shot, but let me ask.
Say we have a list of 5 URLs, only the last is working.
Every minute, you reset to primary, then have to iterate (and fail) 0-3 until you finally return to the single working endpoint.
Why do you think my suggestion is more complex?

asaf-sw:
*[`crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs` line 94 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-OvKSHL-Bfrb_XKPGwKR:-OvPX9-kAFE0fVA-GtSK:be9ba5m) ([raw file](https://github.com/starkware-libs/sequencer/blob/d0a1a60300f077f8c4c016dc1fd9fad425f8f143/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs#L94)):*

Long shot, but let me ask.
Say we have a list of 5 URLs, only the last is working.
Every minute, you reset to primary, then have to iterate (and fail) 0-3 until you finally return to the single working endpoint.
Why do you think my suggestion is more complex?
</blockquote></details>

The code is more complex and there's more room for error because there's another thread. I think this a premature optimization. If we see the current suggested solution isn't good enough then we can revisit.
The likelihood of what you described is pretty low and we technically have only 3 sources in production.
Also, if we're at the 5th option but the 2nd is the first healthy one in the list, we want to use the 2nd, so this yields the desired behavior.
We can also lengthen the interval, say 2m or even 5m. What do you think?

ShahakShama (approved):
> @ShahakShama made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14521)* status: all files reviewed, 1 unresolved discussion (waiting on matanl-starkware).

___
*[`crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs` line 94 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-OvKSHL-Bfrb_XKPGwKR:-OvPYoMF-HEMZHgzFSco:bb6mt9b) ([raw file](https://github.com/starkware-libs/sequencer/blob/d0a1a60300f077f8c4c016dc1fd9fad425f8f143/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs#L94)):*
<details><summary><i>Previously, asaf-sw wrote…</i></summary><blockquote>

The code is more complex and there's more room for error because there's another thread. I think this a premature optimization. If we see the current suggested solution isn't good enough then we can revisit.
The likelihood of what you described is pretty low and we technically have only 3 sources in production.
Also, if we're at the 5th option but the 2nd is the first healthy one in the list, we want to use the 2nd, so this yields the desired behavior.
We can also lengthen the interval, say 2m or even 5m. What do you think?
</blockquote></details>

:+1:

matanl-starkware (approved):
:lgtm:

> @matanl-starkware partially reviewed 10 files, made 4 comments, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-:-OvPePBM-sF5hpshyRda:bnfp4nl)* status: 9 of 10 files reviewed, 1 unresolved discussion (waiting on asaf-sw and ShahakShama).

___
*[`crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs` line 94 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-OvKSHL-Bfrb_XKPGwKR:-OvPZO7B3nBGERlJ43w7:b-yk1btq) ([raw file](https://github.com/starkware-libs/sequencer/blob/d0a1a60300f077f8c4c016dc1fd9fad425f8f143/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs#L94)):*

:+1:
</blockquote></details>

OK
Lets start with this. It's certainly better than what we have right now.
___
*[`crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs` line 20 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-OvPdL1tD9FwyqO41lQR:-OvPdL1tD9FwyqO41lQS:b-b8zdod) ([raw file](https://github.com/starkware-libs/sequencer/blob/1fc0f4b4b391ed24bfdc13048517c3043f3572e0/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs#L20)):*
> ```XML
>     base_layer: B,
>     retry_primary_interval: Duration,
>     last_primary_retry: std::time::Instant,
> ```

Consider using tokio::time::Instant for better testability



_Code quote:_
```XML
std::time::Instant
```
___
*[`crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs` line 35 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-OvPeCSG550bxUP4_ehX:-OvPeCSH25pT8SwXTEzB:bz2ji0a) ([raw file](https://github.com/starkware-libs/sequencer/blob/1fc0f4b4b391ed24bfdc13048517c3043f3572e0/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs#L35)):*
> ```XML
>             self.last_primary_retry = std::time::Instant::now();
>             self.base_layer.reset_provider_url_to_primary().await?;
>         }
> ```

No need for this fuss if already at primary



_Code quote:_
```XML
        if self.last_primary_retry.elapsed() >= self.retry_primary_interval {
            self.last_primary_retry = std::time::Instant::now();
            self.base_layer.reset_provider_url_to_primary().await?;
        }
```

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs
```diff
@@ -15,11 +16,24 @@ pub mod cyclic_base_layer_wrapper_test;
...
+
+    // If the retry-primary interval has elapsed, attempt to revert to the primary endpoint so a
+    // recovered primary is used again. Resets the timer regardless, to keep a fixed cadence and
+    // avoid thrashing.
+    async fn retry_primary_if_due(&mut self) -> Result<(), B::Error> {
+        if self.last_primary_retry.elapsed() >= self.retry_primary_interval {
+            self.last_primary_retry = std::time::Instant::now();
+            self.base_layer.reset_provider_url_to_primary().await?;
+        }
+        Ok(())
```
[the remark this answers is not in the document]
asaf-sw: Fix: reset the timer whenever `cycle_url_on_error` cycles the URL, so the wait always restarts from the failover. Combined with the fix for the *_No need for this fuss if already at primary_* comment, the clock now measures "time since we left the primary".

asaf-sw (commented):
> @asaf-sw made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14521)* status: 5 of 11 files reviewed, all discussions resolved (waiting on matanl-starkware and ShahakShama).

___
*[`crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs` line 20 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-OvPdL1tD9FwyqO41lQR:-OvPiKRRC8Pq2lhIdOMw:bibzn4y) ([raw file](https://github.com/starkware-libs/sequencer/blob/1fc0f4b4b391ed24bfdc13048517c3043f3572e0/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs#L20)):*

Consider using tokio::time::Instant for better testability
</blockquote></details>

Done and added actual tests for the timing transition instead of only the zero/large-interval boundaries.
___
*[`crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs` line 35 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14521#-OvPeCSG550bxUP4_ehX:-OvPiUk90zhskgsfvXiV:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/1fc0f4b4b391ed24bfdc13048517c3043f3572e0/crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs#L35)):*

No need for this fuss if already at primary
</blockquote></details>

Done.

crates/papyrus_base_layer/src/cyclic_base_layer_wrapper.rs
```diff
@@ -43,6 +59,9 @@ impl<B: BaseLayerContract + Send + Sync> CyclicBaseLayerWrapper<B> {
         let Ok(()) = cycle_url_result else {
             return Some(Err(cycle_url_result.expect_err("result is checked at let-else")));
         };
+        // Restart the retry-primary clock on each failover, so the wait is measured from when we
+        // left the primary rather than from the last periodic tick.
+        self.last_primary_retry = tokio::time::Instant::now();
```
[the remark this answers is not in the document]
asaf-sw: Done.
Fixed and added tests to verify correct behavior.

matanl-starkware (approved):
> @matanl-starkware reviewed 7 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14521)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on asaf-sw).

PR #14590, opened 2026-06-22
apollo_l1_events: bound catch-up commit-block backlog with a cap and metric

asaf-sw:
## What
Bounds the L1 events provider's startup catch-up `commit_block_backlog` (previously an unbounded `Vec`) with a configurable cap and adds an observability gauge.

## Why (L-16)
During startup catch-up, every commit-block arriving above the provider's height is appended to `Catchupper::commit_block_backlog` and only drained once L2 sync reaches the target height. There was no cap or backpressure. If sync stalls or lags persistently while the batcher keeps committing, the backlog grows proportionally to the sync gap × inbound commit rate, consuming memory. This is a degraded-mode resource-hygiene issue (the feed is the internal, trusted batcher — not attacker-controlled), rated Low. See analysis: `docs/security-review/L-16.md`.

### Unbounded-growth scenario
1. Node boots far behind the chain tip; catch-up begins syncing toward `target_height`.
2. State sync stalls or repeatedly retries while `commit_block` notifications keep arriving and are appended unconditionally.
3. The backlog accumulates one entry per buffered commit with no cap, growing memory until sync finally catches up (or the node is exhausted).

## Chosen bound + behavior-at-limit
- New config `l1_events_provider_config.max_commit_block_backlog_len` (default `1_000_000` — only tens of MB worst case, comfortably covers any legitimate startup gap).
- On reaching the cap (or on a non-sequential height that would tear the gapless run), `add_commit_block_to_backlog` **abandons the in-memory backlog and defers to L2 sync**: it clears the buffer, latches `backlog_overflowed`, extends the sync target to the tip, and returns `Ok`. While latched, further tip commits are no longer buffered — only the sync target is raised. The buffered commits are redundant with what L2 sync re-delivers as it drives `current_height` up to the tip, so discarding them costs latency (until sync unstalls), not correctness. Result: bounded memory, no gap, no panic. The latch clears when catch-up completes and the backlog drains.
- New gauge `l1_message_provider_commit_block_backlog_len`, set on each push, reset to 0 on abandon, and reset again when the backlog drains at catch-up completion — lets ops observe a stalling sync before memory pressure.

### Why not a hard error (addressing review)
An earlier version returned a `CatchUpBacklogOverflow` error at the cap. That does **not** provide backpressure: the batcher swallows `commit_block` errors and advances anyway (`batcher.rs`, pinned by `decision_reached_return_success_when_l1_commit_block_fails`). The rejected height would be dropped downstream, and the *next* tip commit would then trip the `"Heights should be sequential."` assert and panic the provider into a crash loop — the opposite of the intended degraded-but-recoverable behavior. Abandoning to sync closes that loop within `apollo_l1_events`, without changing the batcher's "always advance" contract.

## Tests
- `backlog_below_cap_accepts_all_commits` (under cap)
- `backlog_fills_exactly_to_cap_without_error` (exactly-at-cap, gapless)
- `commit_after_overflow_does_not_panic` — regression: overflow (swallowed like the batcher does) then a follow-up tip commit must not panic; asserts the backlog is abandoned and the sync target extended
- `commits_after_overflow_stay_bounded_and_track_tip` — repeated post-overflow commits keep the backlog empty and follow the tip
- `catch_up_completes_after_overflow_via_sync` — after overflow, `Equal` sync commits re-drive the range to caught-up / `Pending`
- `non_sequential_backlog_commit_abandons_instead_of_panicking` — the gap trigger hits the same recovery path
- `backlog_len_metric_tracks_push_and_drain` (gauge increments on push, resets on drain)

All `apollo_l1_events` tests pass (88); clippy clean; config schema regenerated.

### DIFF
--- a/crates/apollo_l1_events/src/catchupper.rs
+++ b/crates/apollo_l1_events/src/catchupper.rs
@@ -2,12 +2,14 @@ use std::sync::atomic::{AtomicU64, AtomicU8, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 
-use apollo_l1_events_types::SharedL1EventsProviderClient;
+use apollo_l1_events_types::{L1EventsProviderResult, SharedL1EventsProviderClient};
 use apollo_state_sync_types::communication::SharedStateSyncClient;
 use indexmap::IndexSet;
 use starknet_api::block::BlockNumber;
 use starknet_api::transaction::TransactionHash;
 use tracing::{debug, warn};
+
+use crate::metrics::L1_MESSAGE_PROVIDER_COMMIT_BLOCK_BACKLOG_LEN;
 
 // When the Provider gets a commit_block that is too high, it starts catching up.
 // The commit is rejected by the provider, so it must use sync to catch up to the height of the
@@ -30,6 +32,14 @@ pub struct Catchupper {
     // Keep track of sync task for health checks and logging status.
     pub sync_task_handle: SyncTaskHandle,
     pub n_sync_health_check_failures: Arc<AtomicU8>,
+    /// Cap on `commit_block_backlog` length. Exceeding it abandons the in-memory backlog (rather
+    /// than dropping a single entry, which would tear a gap) and defers to L2 sync; see
+    /// `add_commit_block_to_backlog`.
+    pub max_commit_block_backlog_len: usize,
+    /// Set once the backlog is abandoned during a catch-up (cap hit, or a non-sequential height
+    /// that would tear the gapless run). While set, tip commits are no longer buffered: L2 sync
+    /// owns the whole range and only its target is extended. Cleared when catch-up completes.
+    pub backlog_overflowed: bool,
 }
 
 impl Catchupper {
@@ -41,6 +51,7 @@ impl Catchupper {
         l1_events_provider_client: SharedL1EventsProviderClient,
         sync_client: SharedStateSyncClient,
         sync_retry_interval: Duration,
+        max_commit_block_backlog_len: usize,
     ) -> Self {
         Self {
             sync_retry_interval,
@@ -49,6 +60,8 @@ impl Catchupper {
             sync_client,
             sync_task_handle: SyncTaskHandle::NotStartedYet,
             n_sync_health_check_failures: Default::default(),
+            max_commit_block_backlog_len,
+            backlog_overflowed: false,
             // This is overriden when starting the sync task (e.g., when provider starts
             // catching up).
             target_height: Default::default(),
@@ -68,17 +81,47 @@ impl Catchupper {
         &mut self,
         committed_txs: IndexSet<TransactionHash>,
         height: BlockNumber,
-    ) {
-        assert!(
-            self.commit_block_backlog
-                .last()
-                .is_none_or(|commit_block| commit_block.height.unchecked_next() == height),
-            "Heights should be sequential."
-        );
+    ) -> L1EventsProviderResult<()> {
+        // Already abandoned this catch-up's backlog: L2 sync owns the whole range now. Don't
+        // rebuild the buffer; just keep extending the sync target so the task drives the provider
+        // up to the latest tip height.
+        if self.backlog_overflowed {
+            self.update_target_height(height);
+            return Ok(());
+        }
+
+        let is_sequential = self
+            .commit_block_backlog
+            .last()
+            .is_none_or(|commit_block| commit_block.height.unchecked_next() == height);
+        let is_full = self.commit_block_backlog.len() >= self.max_commit_block_backlog_len;
+
+        // Two runtime-reachable conditions force us off the in-memory fast path: the backlog hit
+        // its cap (a stalled/lagging L2 sync let the tip race ahead), or a non-sequential height
+        // would tear a hole in the gapless run. Dropping any single entry would leave a permanent
+        // gap that corrupts the drain-time sequential invariant (and previously panicked the
+        // provider, since the batcher swallows the error and keeps advancing). Instead, abandon the
+        // backlog entirely and let the authoritative L2 sync re-drive every height up to the tip --
+        // bounded memory, no gap, no panic. The buffered commits are redundant: sync re-delivers
+        // the same heights, just with more latency while it is stalled.
+        if is_full || !is_sequential {
+            warn!(
+                "Catch-up commit-block backlog abandoned at height {height} (cap {}, sequential: \
+                 {is_sequential}); deferring to L2 sync to re-drive the range. L2 sync is likely \
+                 stalled or lagging.",
+                self.max_commit_block_backlog_len
+            );
+            self.backlog_overflowed = true;
+            self.commit_block_backlog.clear();
+            L1_MESSAGE_PROVIDER_COMMIT_BLOCK_BACKLOG_LEN.set_lossy(0_usize);
+            self.update_target_height(height);
+            return Ok(());
+        }
 
         debug!("Adding future commit-block to backlog at height: {height}");
-        self.commit_block_backlog
-            .push(CommitBlockBacklog { height, committed_txs: committed_txs.clone() });
+        self.commit_block_backlog.push(CommitBlockBacklog { height, committed_txs });
+        L1_MESSAGE_PROVIDER_COMMIT_BLOCK_BACKLOG_LEN.set_lossy(self.commit_block_backlog.len());
+        Ok(())
     }
 
     /// Spawns async task that produces and sends commit block messages to the provider, according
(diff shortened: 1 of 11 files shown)


asaf-sw:
* **#14590** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14590?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14590?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/apollo_l1_events/src/catchupper.rs
```diff
@@ -68,17 +76,34 @@ impl Catchupper {
...
             self.commit_block_backlog
                 .last()
                 .is_none_or(|commit_block| commit_block.height.unchecked_next() == height),
             "Heights should be sequential."
         );
 
+        // Bound growth on a stalled/lagging L2 sync. We must NOT drop entries: the backlog is a
+        // gapless, strictly-sequential run and a hole would corrupt the drain-time sequentiality
+        // assert and silently skip an L1-handler commit. Fail loudly instead.
+        if self.commit_block_backlog.len() >= self.max_commit_block_backlog_len {
```
noamsp-starkware: This returns an error without dropping, which is correct in isolation -- but the only caller (batcher `commit_block`, `batcher.rs:1123`) swallows the error and advances. So height `T+cap` is dropped downstream, and the *next* commit (`T+cap+1`) trips the `"Heights should be sequential."` assert at line 84 and panics the provider. The cap therefore turns bounded memory into a guaranteed crash rather than backpressure. See the review summary for options.

crates/apollo_l1_events/src/l1_events_provider.rs
```diff
@@ -432,7 +434,7 @@ impl L1EventsProvider {
             Equal => self.apply_commit_block(committed_txs, Default::default()),
             // We're still syncing, backlog it, it'll get applied later.
             Greater => {
-                self.catchupper.add_commit_block_to_backlog(committed_txs, new_height);
+                self.catchupper.add_commit_block_to_backlog(committed_txs, new_height)?;
```
noamsp-starkware, line 437: Propagating with `?` looks like backpressure, but the batcher discards this error (`batcher.rs:1136-1142`, returns `Ok`) and does not retry the height. Nothing here holds or re-sends the rejected commit, so the sequential invariant the backlog relies on is broken as soon as the cap is hit.

noamsp-starkware (changes requested):
The cap addition is well-motivated and the metric/tests are useful, but the recovery story at the limit does not hold: **hitting the cap converts an unbounded-memory condition into a guaranteed panic of the L1 events provider**, which is the opposite of the stated "degraded-but-recoverable" goal.

**Why.** The design assumes returning `CatchUpBacklogOverflow` gives backpressure. Its only real caller, the batcher, swallows it: `batcher.rs:1123-1142` logs the error, increments `BATCHER_L1_EVENTS_PROVIDER_ERRORS`, and returns `Ok(())`. This is deliberate and pinned by the test `decision_reached_return_success_when_l1_commit_block_fails` (`batcher_test.rs:1723`). So the batcher never retries the rejected height and keeps advancing.

**Sequence at the cap** (backlog full = `[T .. T+cap-1]`):
1. Commit `T+cap` is rejected. Batcher swallows it -> height `T+cap` is **dropped**, the exact outcome this PR says it prevents.
2. Next tip commit `T+cap+1` arrives, still `Greater`, calls `add_commit_block_to_backlog`.
3. The pre-existing `"Heights should be sequential."` assert fires (`last` = `T+cap-1`, `.unchecked_next()` = `T+cap` != `T+cap+1`) -> **panic**, crashing the provider.

The gap the PR aims to prevent happens anyway, just at push-time as a crash rather than at drain-time. After the panic, catch-up restarts from the provider height; if L2 sync is still stalled (the condition that filled the backlog), it refills and panics again -> crash loop until sync happens to catch up. There is no self-healing path from the cap condition itself.

**To make it actually recoverable, close the loop with the caller instead of returning into a void:**
- Batcher treats `CatchUpBacklogOverflow` as retryable (hold + re-send the height) so no gap is created and memory stays bounded -- a real change to the current "always Ok" contract; or
- On overflow, abandon the backlog and let sync re-drive from `current_height`, making the drain/push paths tolerate a re-synced range.
- Either way, the two asserts (`"Heights should be sequential."` and the drain-time one) are now reachable from a runtime degraded state and must become recoverable errors per the repo rule against panicking on data reachable from external/runtime input.

**Minor:**
- PR body says "The error surfaces to the batcher (which already handles `commit_block` errors)" -- "handles" here means "discards"; worth correcting since the safety argument rests on it.
- None of the four tests exercise the *post-overflow next-commit* path; a test that pushes one more height after overflow would immediately surface the panic.

crates/apollo_l1_events/src/catchupper.rs
asaf-sw: Agreed, fixed. This no longer returns `CatchUpBacklogOverflow` (the variant is removed). On overflow it abandons the backlog and defers to L2 sync, returning `Ok`, so there's no error for the batcher to swallow and no dropped height: sync re-drives the whole range (including `T+cap`) via the `Equal` path. The `"Heights should be sequential."` assert on the next push is replaced by the same graceful abandon. Covered by `commit_after_overflow_does_not_panic`.

crates/apollo_l1_events/src/l1_events_provider.rs
asaf-sw: Right, the `?` here no longer propagates a rejection (there isn't one to discard). `add_commit_block_to_backlog` now returns `Ok` in the overflow case after clearing the buffer and raising the sync target, so the sequential invariant is preserved by handing the range to L2 sync rather than relying on the batcher to hold/re-send. Nothing is left un-sent, sync re-delivers the heights authoritatively.

asaf-sw (commented):
You're exactly right, and I confirmed the whole path - the batcher swallows `commit_block` errors (`batcher.rs`, `other_err` arm → `BATCHER_L1_EVENTS_PROVIDER_ERRORS.increment(1)` → falls through to `Ok(())`, pinned by `decision_reached_return_success_when_l1_commit_block_fails`), so the rejected height is dropped and the next tip commit trips `"Heights should be sequential."`. I reproduced it as a failing test first (`commit_after_overflow_does_not_panic`) -
it panicked at `catchupper.rs:80` exactly as you described.

**Fix - option 2 (abandon the backlog, re-drive from L2 sync).** On overflow *or* a non-sequential height, `add_commit_block_to_backlog` no longer returns an error into the void. It clears the buffer, latches `backlog_overflowed`, and extends the sync target to the tip; while latched, further tip commits only raise the target. L2 sync then re-drives `current_height` up via the authoritative `Equal` path. The buffered commits are redundant with what sync re-delivers, so discarding them costs latency (until sync unstalls), not correctness - bounded memory, no gap, no panic, and a real self-healing path from the cap condition.

I chose this over option 1 (batcher retry) deliberately: option 1 reverses the "batcher always advances" contract that `decision_reached_return_success_when_l1_commit_block_fails` pins and couples block production to L1-provider health. Option 2 keeps the change inside `apollo_l1_events`.

**Asserts.** The first assert (`"Heights should be sequential."`) is gone - it's now the `is_sequential` check that routes to the same graceful abandon path. The drain-time assert I kept: after this change it's no longer reachable from the degraded state (on abandon the backlog is cleared, so at drain it's either empty or gapless-by-construction), so it now guards only a genuine internal invariant. Happy to downgrade it to a log+abandon too if you'd prefer belt-and-suspenders.

**Minor.** Corrected the PR body - it now describes the abandon-to-sync behavior; no error surfaces to the batcher at all, so the "handles = discards" wording is gone.

**Tests** (all red-green - each fails on the pre-fix code):
- `commit_after_overflow_does_not_panic` - the exact post-overflow next-commit path you flagged.
- `commits_after_overflow_stay_bounded_and_track_tip` - memory stays bounded across many overflowing commits.
- `catch_up_completes_after_overflow_via_sync` - proves recovery: sync re-drives to `Pending`.
- `non_sequential_backlog_commit_abandons_instead_of_panicking` - the gap trigger.

> @asaf-sw made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14590#-:-Ox0hmDDCvNe6nirdF3u:blthmm1)* status: 0 of 12 files reviewed, 2 unresolved discussions (waiting on noamsp-starkware).

noamsp-starkware (approved):
:lgtm:

> @noamsp-starkware reviewed 12 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14590#-:-Ox54QQrCJr_aDjjpkIh:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on asaf-sw).

gkaempfer:

---

PR #14582, opened 2026-06-21
starknet_patricia: introduce LeafSource enum for filled-tree leaf handling

yoavGrs:
Replace the mutually-exclusive (Option<leaf_modifications>, leaf_input_map, leaf_output_map)
parameters threaded through compute_filled_tree_rec with a single LeafSource enum, so the two
leaf-handling modes can no longer be combined nonsensically and callers no longer pass dummy
empty maps for the unused mode.

The enum holds its maps directly (no per-field Arc): the single shared Arc<LeafSource> already
provides the cross-task ownership, so create reclaims the leaf output map via Arc::into_inner
after the recursion.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

PR #14603, opened 2026-06-23
starknet_patricia: use std::sync::Mutex in filled-tree traversal

gkaempfer:
## Summary

Performance follow-up to #14582 (introduced `LeafSource` enum).

### The issue

All three lock operations in the filled-tree traversal were using `tokio::sync::Mutex`, but **none of them cross an `.await` point while holding the lock**:

| Function | Lock use | `.await` inside lock? |
|---|---|---|
| `write_to_output_map` | take-or-write a `Mutex<Option<T>>` | No |
| `remove_mutex_and_option_from_output_map` | take each value | No |
| `get_or_compute_leaf` (ComputeLeaves branch) | `.take()` the input, drop guard, *then* `L::create().await` | No |

`tokio::sync::Mutex` is designed for holding locks across yield points. Even for an uncontended acquisition it goes through async-runtime machinery (creates a `MutexLockFuture`, registers a waker, polls once). That overhead is unnecessary here.

### The fix

Replace `tokio::sync::Mutex` with `std::sync::Mutex`. Because no guard is held across any `.await`, this is correct. The three helper functions (`write_to_output_map`, `remove_mutex_and_option_from_output_map`, `remove_arc_mutex_and_option_from_output_map`) also become plain `fn` instead of `async fn`, which makes the callsites cleaner.

### Expected impact

The filled-tree traversal runs on every state-commitment update (storage writes, contract deployments). A typical block touches thousands of leaves, each requiring O(log N ≈ 251) node visits — each visit previously incurred a `tokio::sync::Mutex` lock/unlock cycle. Removing the async-scheduler overhead from every one of these acquisitions should reduce CPU time in the state-commitment hot path.

## Test plan
- [x] All 107 existing `starknet_patricia` tests pass unchanged

---
_Generated by [Claude Code](https://claude.ai/code/session_01FndKFAhehesa98Vq9PUWiH)_

### DIFF
--- a/crates/starknet_patricia/src/patricia_merkle_tree/filled_tree/tree.rs
+++ b/crates/starknet_patricia/src/patricia_merkle_tree/filled_tree/tree.rs
@@ -1,14 +1,13 @@
 use std::collections::HashMap;
 use std::fmt::Debug;
 use std::future::Future;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 
 use async_recursion::async_recursion;
 use starknet_api::hash::HashOutput;
 use starknet_patricia_storage::db_object::{DBObject, HasStaticPrefix};
 use starknet_patricia_storage::errors::SerializationResult;
 use starknet_patricia_storage::storage_trait::DbHashMap;
-use tokio::sync::Mutex;
 
 use crate::db_layout::NodeLayoutFor;
 use crate::patricia_merkle_tree::filled_tree::errors::FilledTreeError;
@@ -99,14 +98,16 @@ impl<L: Leaf + 'static> FilledTreeImpl<L> {
 
     /// Writes the hash and data to the output map. The writing is done in a thread-safe manner with
     /// interior mutability to avoid thread contention.
-    async fn write_to_output_map<T: Debug>(
+    fn write_to_output_map<T: Debug>(
         output_map: &HashMap<NodeIndex, Mutex<Option<T>>>,
         index: NodeIndex,
         output: T,
     ) -> FilledTreeResult<()> {
         match output_map.get(&index) {
             Some(node) => {
-                let mut node = node.lock().await;
+                let mut node = node
+                    .lock()
+                    .expect("Output map mutex does not expect to panic at locking the mutex");
                 match node.take() {
                     Some(existing_node) => Err(FilledTreeError::DoubleUpdate {
                         index,
@@ -123,7 +124,7 @@ impl<L: Leaf + 'static> FilledTreeImpl<L> {
     }
 
     // Removes the `Arc` from the map and unwraps the `Mutex` and `Option` from the values.
-    async fn remove_arc_mutex_and_option_from_output_map<V>(
+    fn remove_arc_mutex_and_option_from_output_map<V>(
         output_map: Arc<HashMap<NodeIndex, Mutex<Option<V>>>>,
         panic_if_empty_placeholder: bool,
     ) -> FilledTreeResult<HashMap<NodeIndex, V>> {
@@ -132,7 +133,9 @@ impl<L: Leaf + 'static> FilledTreeImpl<L> {
 
         let mut hash_map_out = HashMap::new();
         for (key, value) in output_map {
-            let mut value = value.lock().await;
+            let mut value = value
+                .lock()
+                .expect("Output map mutex does not expect to panic at locking the mutex");
             match value.take() {
                 Some(unwrapped_value) => {
                     hash_map_out.insert(key, unwrapped_value);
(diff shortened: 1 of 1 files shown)


yoavGrs:
* **#14603** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14603?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14603?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

nimrod-starkware (changes requested):
> @nimrod-starkware reviewed all commit messages and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14603)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on gkaempfer and yoavGrs).

___
*[`crates/starknet_patricia/src/patricia_merkle_tree/filled_tree/tree.rs` line 108 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/14603#-OvrwBjz6S3Nd08JZzOO:-OvrwBjz6S3Nd08JZzOP:bgb518t) ([raw file](https://github.com/starkware-libs/sequencer/blob/7884e7ce68dc2932a1e5e4d90b0ae94db6be2dad/crates/starknet_patricia/src/patricia_merkle_tree/filled_tree/tree.rs#L108)):*
> ```XML
>         match output_map.get(&index) {
>             Some(node) => {
>                 let mut node = node.lock().unwrap();
> ```

Does this `unwrap` should never panic because we never call this method with the same index?
Please replace to expect(...) or even consider to return an error with a new error variant.
___
*[`crates/starknet_patricia/src/patricia_merkle_tree/filled_tree/tree.rs` line 172 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/14603#-Ovrwvp1DTm1Q6NX8UWu:-Ovrwvp1DTm1Q6NX8UWv:b-kk2v9n) ([raw file](https://github.com/starkware-libs/sequencer/blob/7884e7ce68dc2932a1e5e4d90b0ae94db6be2dad/crates/starknet_patricia/src/patricia_merkle_tree/filled_tree/tree.rs#L172)):*
> ```XML
>             .ok_or(FilledTreeError::MissingLeafInput(index))?
>             .lock()
>             .unwrap()
> ```

same as above, why do expect this not to panic? 
Please replace to expect(...) or even consider to return an error with a new error variant.



_Code quote:_
```XML
            .unwrap()
```

yoavGrs (commented):
> @yoavGrs made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14603)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on nimrod-starkware).

___
*[`crates/starknet_patricia/src/patricia_merkle_tree/filled_tree/tree.rs` line 108 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/14603#-OvrwBjz6S3Nd08JZzOO:-OvsRnqs5-PkMAhx3wLf:b-5nwe9o) ([raw file](https://github.com/starkware-libs/sequencer/blob/7884e7ce68dc2932a1e5e4d90b0ae94db6be2dad/crates/starknet_patricia/src/patricia_merkle_tree/filled_tree/tree.rs#L108)):*

Does this `unwrap` should never panic because we never call this method with the same index?
Please replace to expect(...) or even consider to return an error with a new error variant.
</blockquote></details>

`lock` returns `Err(PoisonError)` only if there is a thread that panics with that mutex locked.
___
*[`crates/starknet_patricia/src/patricia_merkle_tree/filled_tree/tree.rs` line 172 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/14603#-Ovrwvp1DTm1Q6NX8UWu:-OvsT8Fy6KT26nV_H8Pi:bb73xfv) ([raw file](https://github.com/starkware-libs/sequencer/blob/7884e7ce68dc2932a1e5e4d90b0ae94db6be2dad/crates/starknet_patricia/src/patricia_merkle_tree/filled_tree/tree.rs#L172)):*

same as above, why do expect this not to panic? 
Please replace to expect(...) or even consider to return an error with a new error variant.
</blockquote></details>

Same

nimrod-starkware (approved):
:lgtm:

> @nimrod-starkware reviewed 1 file and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14603#-:-OvsriSp4ohcmmBUf7WU:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gkaempfer).

gkaempfer:

---

PR #14612, opened 2026-06-23
blockifier: record pre-migration compiled class hash as initial read

yoavGrs:
When migrating a class's compiled class hash, `set_compiled_class_hash_migration`
wrote the post-migration (v2) value into the write cache without first recording
the pre-block (v1) value as an initial read. Since `get_compiled_class_hash`
resolves the write cache before the initial reads, the v2 write shadowed the
pre-block value, so it was never captured. This made `to_state_diff` panic in
`strict_subtract_mappings` (the migrated write had no matching initial read to
subtract against) and left the OS without the pre-block class-trie leaf needed to
replay the block.

Force-read the pre-migration compiled class hash into the initial reads before
applying the migration write, and add a unit test covering the fix.

### DIFF
--- a/crates/blockifier/src/state/compiled_class_hash_migration.rs
+++ b/crates/blockifier/src/state/compiled_class_hash_migration.rs
@@ -6,6 +6,10 @@ use crate::blockifier::transaction_executor::CompiledClassHashV2ToV1;
 use crate::state::cached_state::CachedState;
 use crate::state::state_api::{State, StateReader, StateResult};
 
+#[cfg(test)]
+#[path = "compiled_class_hash_migration_test.rs"]
+mod compiled_class_hash_migration_test;
+
 pub trait CompiledClassHashMigrationUpdater {
     fn set_compiled_class_hash_migration(
         &mut self,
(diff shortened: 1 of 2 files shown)


yoavGrs:
* **#14612** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14612?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14612?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

Yoni-Starkware (changes requested):
> @Yoni-Starkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14612)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on yoavGrs).

___
*[`crates/blockifier/src/state/compiled_class_hash_migration.rs` line 36 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14612#-OvpRP5K1AFlF09O9Smz:-OvpRP5K1AFlF09O9Sn-:b1wxrb) ([raw file](https://github.com/starkware-libs/sequencer/blob/cdbf3ab07b4b6752981a686ccd176d0419e489ac/crates/blockifier/src/state/compiled_class_hash_migration.rs#L36)):*
> ```XML
> 
>             // Capture the pre-block (v1) hash as an initial read before the v2 write shadows it.
>             self.get_compiled_class_hash(*class_hash)?;
> ```

Every migrated class should have triggered a call to `get_compiled_class_hash` before this point. See here: https://github.com/starkware-libs/sequencer/blob/a5cbd5360136a52a6097a17fabd9882a9433f7c8/crates/blockifier/src/utils.rs#L126

Also, it's clear from the function API - it gets the previous compiled class hash, so not sure why this read is needed



_Code quote:_
```XML
            // Capture the pre-block (v1) hash as an initial read before the v2 write shadows it.
            self.get_compiled_class_hash(*class_hash)?;
```

PR #14600, opened 2026-06-23
starknet_patricia: thread the leaf output map separately from LeafSource

yoavGrs:
Move leaf_index_to_leaf_output out of the LeafSource::ComputeLeaves variant and
pass it to compute_filled_tree_rec as its own Arc, mirroring filled_tree_output_map.
This lets create reclaim the leaf output map via the existing
remove_arc_mutex_and_option_from_output_map helper instead of unwrapping the shared
LeafSource Arc and matching a single variant with an unreachable! arm.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

PR #14616, opened 2026-06-24
starknet_patricia: avoid async locking when reclaiming output maps

gkaempfer:
## Summary

Follow-up performance optimization to #14600 (`starknet_patricia: thread the leaf output map separately from LeafSource`).

### Optimization

In `remove_arc_mutex_and_option_from_output_map`, after `Arc::into_inner()` succeeds the caller holds **exclusive ownership** of the `HashMap` — no other reference to the `Arc` exists. This means the `Mutex` inside can never be contended at that point, yet the previous code still called `value.lock().await` on every entry, incurring tokio-scheduler overhead for each node in the tree.

**Changes:**
- Replace `value.lock().await` / `value.take()` with `value.into_inner()`, extracting the inner value synchronously without acquiring the lock.
- Make the function non-`async` (the only async operation was the now-removed `lock().await`).
- Pre-allocate the output `HashMap` with `with_capacity(output_map.len())` to avoid incremental rehashing.

The function is called twice per `create` invocation (once for the filled-tree output map, once for the leaf output map) and once per `create_with_existing_leaves` invocation. Both maps can contain hundreds to thousands of entries for large Patricia-Merkle tree updates, so the per-entry async overhead compounds.

### Expected impact

- Eliminates O(n) tokio async-lock acquisitions at the end of each tree computation, where n is the number of modified nodes.
- Avoids O(n) HashMap rehash operations during output-map construction.
- All 107 existing tests pass.

---
_Generated by [Claude Code](https://claude.ai/code/session_012bGFCWpsYc5cWrkGyYGapB)_

### DIFF
--- a/crates/starknet_patricia/src/patricia_merkle_tree/filled_tree/tree.rs
+++ b/crates/starknet_patricia/src/patricia_merkle_tree/filled_tree/tree.rs
@@ -70,7 +70,7 @@ enum LeafSource<L: Leaf> {
     /// `leaf_index_to_leaf_output`.
     ComputeLeaves {
         leaf_index_to_leaf_input: HashMap<NodeIndex, Mutex<Option<L::Input>>>,
-        leaf_index_to_leaf_output: HashMap<NodeIndex, Mutex<Option<L::Output>>>,
+        leaf_index_to_leaf_output: Arc<HashMap<NodeIndex, Mutex<Option<L::Output>>>>,
     },
 }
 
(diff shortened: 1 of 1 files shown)


yoavGrs (commented):
Activity.

> @yoavGrs made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14616#-:-OzKukzU0kqNgkUA3WuY:bm0o6m2)* status: 0 of 1 files reviewed, all discussions resolved.

PR #14633, opened 2026-06-25
apollo_batcher,apollo_batcher_config: shorter tx polling interval for proposal validation

matanl-starkware:
The block-builder loop polls its transaction provider and sleeps
tx_polling_interval_millis whenever the provider returns no transactions.
This is tuned high in prod (200ms) to avoid hammering the mempool while
proposing, but the same value was also used when validating a proposal,
where the provider reads from a local in-process channel fed by the consensus
stream. A validator could therefore wait up to a full interval (~200ms) before
picking up already-streamed transactions, even though the channel poll is free.

Split the knob: add validate_tx_polling_interval_millis (default 10ms) and use
it on the validate path, leaving tx_polling_interval_millis for the propose
path. Prod only overrides the propose value, so validators fall through to the
10ms default with no deployment change.

### DIFF
--- a/crates/apollo_batcher/src/batcher.rs
+++ b/crates/apollo_batcher/src/batcher.rs
@@ -532,7 +532,7 @@ impl Batcher {
                     tx_polling_interval_millis: self
                         .config
                         .dynamic_config
-                        .tx_polling_interval_millis,
+                        .validate_tx_polling_interval_millis,
                 },
                 self.config.dynamic_config.native_classes_whitelist.clone(),
                 Box::new(tx_provider),
--- a/crates/apollo_batcher_config/src/config.rs
+++ b/crates/apollo_batcher_config/src/config.rs
@@ -321,8 +321,16 @@ pub struct BatcherDynamicConfig {
     /// Number of transactions in each request from the tx_provider.
     pub n_concurrent_txs: usize,
     /// Time to wait (in milliseconds) between transaction requests when the previous request
-    /// returned no transactions.
+    /// returned no transactions. Applies when proposing, where the provider polls the mempool.
+    /// Kept intentionally high: a coarse interval makes txs accumulate between polls so ordering
+    /// is decided by tip/priority fee, rather than letting a tx win inclusion purely by arriving
+    /// a few ms earlier (i.e. by sitting geographically close to our node). It also reduces
+    /// mempool polling load.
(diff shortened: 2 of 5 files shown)


matanl-starkware:
* **#14633** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14633?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14633?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ShahakShama (approved):
:lgtm:

> @ShahakShama reviewed 5 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14633#-:-OvyeLSQ208loFkDVO55:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on dafnamatsry and matanl-starkware).

___
*[`crates/apollo_batcher_config/src/config.rs` line 328 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14633#-Ovye2s30gnwTA9svdtc:-Ovye2s30gnwTA9svdtd:b-92v85b) ([raw file](https://github.com/starkware-libs/sequencer/blob/715a25b55cf6e1f99d3627b2d906cb09c0b9ac31/crates/apollo_batcher_config/src/config.rs#L328)):*
> ```XML
>     /// Same as `tx_polling_interval_millis`, but applied when validating a proposal. The validate
>     /// provider reads from a local in-process channel fed by the consensus stream, so polling is
>     /// cheap and this can be much shorter to reduce the delay before streamed txs are executed.
> ```

This is not the full story. The full story is that in proposer flow we want to encourage users to pay for tip in order to be included first and not to just sit geographically close to our node. I think it's worth adding to the comment

matanl-starkware (commented):
> @matanl-starkware reviewed 5 files and all commit messages, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14633)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dafnamatsry).

gkaempfer:

---

PR #14642, opened 2026-06-25
apollo_propeller: O(1) incremental size check in create_message_batch

gkaempfer:
## What

Performance follow-up to #14533 on the `apollo_propeller` connection handler.

`create_message_batch` called `batch.encoded_len()` — which traverses the entire batch — inside its fill loop, making outbound batch assembly **O(N²)** in the number of units per batch. It also cloned each candidate unit before deciding whether it fits, discarding the clone on rejection.

### 1. O(1) incremental size check (`create_message_batch`)

Replace the repeated `encoded_len()` calls with an **O(1) incremental running total**: compute each candidate's protobuf overhead before touching the queue, update the counter on acceptance, and `break` without any clone on rejection. Batch assembly is now **O(N)**.

`PropellerUnitBatch` is a single `repeated PropellerUnit batch = 1` field, which proto3 encodes as one independent record per element:

```
1 byte (tag 0x0A)  +  varint_len(unit.encoded_len())  +  unit.encoded_len()
```

There is no element-count prefix, so appending a unit never changes the cost of the units already in the batch — which is what makes the running total exact. The increment is computed via `prost::encoding::encoded_len_varint` and matches `ProtoBatch::encoded_len()` exactly.

Units are moved out of the queue with `pop_front()` on acceptance rather than cloned-then-maybe-discarded, and every unit is handled uniformly by the loop: the first unit is identified by `batch.batch.is_empty()` rather than being popped and measured separately before the loop. An oversized first unit is still included (with the advisory `warn!`), since dropping it would block the queue forever.

### 2. Bound inbound buffering

`poll_inner` carried a comment saying it reads from the wire "only when the unsent buffer is empty", but had no such check — it polled the inbound substreams unconditionally. Added the missing `unsent_units.is_empty()` guard, so a peer sending batches faster than the engine drains the bounded channel from #14533 can no longer grow `unsent_units` without bound.

### 3. Waker clone elision

Skip the `Waker::clone()` in `poll` when the stored waker already refers to the same task (`will_wake`), avoiding an unnecessary atomic increment on every steady-state poll.

## Impact

- Outbound batch creation: O(N²) → O(N) in units-per-batch
- Eliminates O(N) unnecessary deep clones per batch-fill cycle
- Inbound buffering is now actually bounded by the engine channel, as the code always claimed
- Waker path: one atomic refcount saved per `poll` call under steady state

## Tests

5 unit tests in `handler_test.rs` covering `create_message_batch`:

- Empty queue
- Single unit fits
- Single unit over the limit (still included, advisory warn only)
- All units fit
- Batch stops at the size limit, with the queue remainder preserved

A `varint_boundary` test was added and then removed: it exercised a single unit whose own length prefix crosses 127→128, which is not the case that was asked about. The unit count itself is never encoded, so appending a unit never widens a shared varint inside the budget. The varint that *does* grow with the unit count is the length delimiter `PropellerCodec::encode` prepends to the whole batch, and that sits outside `max_wire_message_size` — `ProtoCodec` compares the limit against the body on both encode and decode. `create_message_batch` now carries a `TODO(Shahak)` recording that gap.

All existing tests continue to pass: 166 unit tests + 4 e2e tests.

### DIFF
--- a/crates/apollo_propeller/src/handler.rs
+++ b/crates/apollo_propeller/src/handler.rs
@@ -239,6 +239,10 @@ impl Handler {
     }
 
     /// Create a batch of messages from the send queue that fits within max_wire_message_size.
+    ///
+    /// `max_wire_message_size` bounds the protobuf body (i.e. `ProtoBatch::encoded_len()`),
+    /// matching the codec's length check. The framed wire bytes are up to ~5 bytes larger due
+    /// to the length delimiter prepended by `PropellerCodec::encode`.
     fn create_message_batch(
         send_queue: &mut VecDeque<ProtoUnit>,
         max_wire_message_size: usize,
--- a/crates/apollo_propeller/src/handler.rs
+++ b/crates/apollo_propeller/src/handler.rs
@@ -18,12 +18,17 @@ use libp2p::swarm::handler::{
     StreamUpgradeError,
 };
 use libp2p::swarm::{Stream, SubstreamProtocol};
+use prost::encoding::encoded_len_varint;
 use prost::Message;
 use tracing::{error, trace, warn};
 
 use crate::config::Config;
 use crate::protocol::{PropellerCodec, PropellerProtocol};
 use crate::PropellerUnit;
+
+#[cfg(test)]
+#[path = "handler_test.rs"]
+mod handler_test;
 
 /// Events that the handler can send to the behaviour.
 #[derive(Debug)]
@@ -242,19 +247,27 @@ impl Handler {
             return ProtoBatch { batch: Vec::new() };
         }
 
-        let mut batch = ProtoBatch { batch: vec![send_queue.pop_front().unwrap()] };
-        if batch.encoded_len() > max_wire_message_size {
+        let first = send_queue.pop_front().unwrap();
+        let mut batch = ProtoBatch { batch: vec![first] };
+        let mut batch_encoded_len = batch.encoded_len();
+        if batch_encoded_len > max_wire_message_size {
             warn!("Propeller unit size exceeds max wire message size, sending will fail");
         }
 
         while let Some(msg) = send_queue.front() {
-            batch.batch.push(msg.clone());
-            if batch.encoded_len() <= max_wire_message_size {
-                send_queue.pop_front();
-            } else {
-                batch.batch.pop();
+            let msg_encoded_len = msg.encoded_len();
+            // Per-item cost in a protobuf `repeated` message field: 1-byte tag (field 1,
+            // LEN wire type = 0x0A) + varint-encoded length prefix + the message bytes.
+            // Tracking this incrementally avoids re-encoding the entire batch on every
+            // iteration, dropping the loop from O(N²) to O(N).
+            let msg_encoded_len_u64 =
+                u64::try_from(msg_encoded_len).expect("message encoded length fits in u64");
+            let item_len = 1 + encoded_len_varint(msg_encoded_len_u64) + msg_encoded_len;
+            if batch_encoded_len + item_len > max_wire_message_size {
                 break;
             }
+            batch_encoded_len += item_len;
+            batch.batch.push(send_queue.pop_front().unwrap());
         }
 
         batch
@@ -550,7 +563,11 @@ impl ConnectionHandler for Handler {
     > {
         let result = self.poll_inner(cx);
         // Store the waker so on_behaviour_event / on_connection_event can wake us.
-        self.waker = Some(cx.waker().clone());
+        // Skip the clone when the stored waker already refers to the same task — avoids
+        // an atomic increment on every poll in the common steady-state case.
+        if self.waker.as_ref().is_none_or(|w| !w.will_wake(cx.waker())) {
+            self.waker = Some(cx.waker().clone());
+        }
         result
     }
 
--- a/crates/apollo_propeller/src/handler.rs
+++ b/crates/apollo_propeller/src/handler.rs
@@ -502,9 +502,12 @@ impl Handler {
         }
 
         // Read from the wire only when the unsent buffer is empty (the previous batch has been
-        // fully delivered).
-        for inbound_substream in self.inbound_substream.iter_mut() {
-            Self::poll_single_inbound_substream(inbound_substream, &mut self.unsent_units, cx);
+        // fully delivered). This prevents unbounded accumulation: a malicious peer cannot force
+        // memory growth by sending batches faster than the engine drains the channel.
+        if self.unsent_units.is_empty() {
+            for inbound_substream in self.inbound_substream.iter_mut() {
+                Self::poll_single_inbound_substream(inbound_substream, &mut self.unsent_units, cx);
+            }
         }
 
         // Send newly-decoded units to the engine channel.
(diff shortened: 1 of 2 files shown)


gkaempfer:
Keeping active — awaiting review from @ShahakShama and @sirandreww-starkware.

---

ShahakShama (changes requested):
> @ShahakShama reviewed 2 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14642)* status: all files reviewed, 4 unresolved discussions (waiting on gkaempfer and sirandreww-starkware).

___
*[`crates/apollo_propeller/src/handler_test.rs` line 19 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/14642#-Ozp2Z1ZD4mzMJhTs_vO:-Ozp2Z1ZD4mzMJhTs_vP:b-qlwyrw) ([raw file](https://github.com/starkware-libs/sequencer/blob/f338832250887342fa5573170790e094306a7c9b/crates/apollo_propeller/src/handler_test.rs#L19)):*
> ```XML
>     let unit_encoded_len = unit.encoded_len();
>     let unit_encoded_len_u64 = u64::try_from(unit_encoded_len).expect("encoded length fits in u64");
>     1 + prost::encoding::encoded_len_varint(unit_encoded_len_u64) + unit_encoded_len
> ```

Why do we need the +1? Describe in a comment
___
*[`crates/apollo_propeller/src/handler_test.rs` line 70 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/14642#-Ozp3BVw1k7lg7XJ9_7x:-Ozp3BVw1k7lg7XJ9_7y:b-qhtyt8) ([raw file](https://github.com/starkware-libs/sequencer/blob/f338832250887342fa5573170790e094306a7c9b/crates/apollo_propeller/src/handler_test.rs#L70)):*
> ```XML
>     let unit = make_proto_unit(20);
>     let single_item_cost = item_batch_cost(&unit);
>     let max_size = 2 * single_item_cost; // fits exactly 2
> ```

Change this to 2*single_item_cost + 1
___
*[`crates/apollo_propeller/src/handler_test.rs` line 83 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/14642#-Ozp3XDg3Tl5Sk9mV7US:-Ozp3XDg3Tl5Sk9mV7UT:b2o1eci) ([raw file](https://github.com/starkware-libs/sequencer/blob/f338832250887342fa5573170790e094306a7c9b/crates/apollo_propeller/src/handler_test.rs#L83)):*
> ```XML
> 
> #[test]
> fn test_create_message_batch_packed_maximally() {
> ```

These 3 tests (this and the 2 below it) don't add anything that the existing test dont already check
___
*[`crates/apollo_propeller/src/handler_test.rs` line 149 at r5](https://reviewable.io/reviews/starkware-libs/sequencer/14642#-Ozp3hsc2KWpTH6R_Lz8:-Ozp3hsc2KWpTH6R_Lz9:b-b6p70j) ([raw file](https://github.com/starkware-libs/sequencer/blob/f338832250887342fa5573170790e094306a7c9b/crates/apollo_propeller/src/handler_test.rs#L149)):*
> ```XML
>         .sum();
>     assert_eq!(batch.encoded_len(), expected_len);
> }
> ```

Do we encode in varint at the start the amount of units? If yes, add a test where adding the last item causes the varint to grow in 1 byte and that causes not enough space

gkaempfer:
Addressing all four discussions from @ShahakShama (commit 4e0941f):

**Line 19 — +1 comment**
Added: `// 1-byte tag (field 1, LEN wire type 0x0A) + varint-encoded item length + item bytes.`

**Line 70 — 2×cost+1**
Changed to `2 * single_item_cost + 1` and updated the comment to "fits exactly 2, 3rd would exceed".

**Lines 83+ — redundant tests**
Removed `test_create_message_batch_packed_maximally` and `test_create_message_batch_tight_boundary` — agreed these overlap with the existing stop/all-fit tests.

Kept `test_create_message_batch_encoded_len_matches_proto`: it tests something orthogonal to the behavioral tests — it validates that the incremental formula `1 + varint_len(len) + len` agrees with prost's own `ProtoBatch::encoded_len()`. A formula with a wrong coefficient (e.g., `len + 2` instead of `1 + 1 + len` for small messages) would produce identical packing outcomes but diverge here. Happy to remove it if you still consider it redundant.

**Line 149 — varint growth**
No, protobuf `repeated message` fields don't encode a leading item count. Each item is encoded independently as `[tag][length_varint][data]`, with no prefix for the total.

However, the per-item *length* varint can grow: when a unit's `encoded_len()` crosses the 127→128 boundary, `encoded_len_varint(127) = 1` byte but `encoded_len_varint(128) = 2` bytes, adding one extra byte to that item's cost. Added `test_create_message_batch_varint_boundary` which verifies that a unit at this boundary is correctly excluded when the budget fits a 127-encoded-len item (cost 129) but not a 128-encoded-len one (cost 131).

All 172 tests pass, rustfmt clean.

---

ShahakShama:
Addressed all four review comments — thanks. Pushed as `f662a3a1ae`, on top of `4e0941f405`.

> `handler_test.rs` line 19: Why do we need the +1? Describe in a comment

Done — `item_batch_cost` now documents it: an element of a `repeated` field is encoded as tag + length + payload, and `PropellerUnitBatch::batch` is field 1 with the LEN wire type, so the tag is the single byte `0x0A`. Also hoisted `encoded_len_varint` to the import block now that `item_batch_cost` is its only caller.

> `handler_test.rs` line 70: Change this to 2*single_item_cost + 1

Done. The assertions still hold: a 3rd unit needs a full `single_item_cost`, so the spare byte cannot be used.

> `handler_test.rs` line 83: These 3 tests (this and the 2 below it) don't add anything that the existing test dont already check

All three are gone — `packed_maximally` and `tight_boundary` in `4e0941f405`, and `encoded_len_matches_proto` in `f662a3a1ae`.

> `handler_test.rs` line 149: Do we encode in varint at the start the amount of units? If yes, add a test where adding the last item causes the varint to grow in 1 byte and that causes not enough space

No, there is no leading count. `PropellerUnitBatch` is a single `repeated PropellerUnit batch = 1` field, and proto3 encodes a repeated message field as one independent `tag + varint(len) + payload` record per element — so appending a unit never changes the cost of the units already in the batch. That is exactly what makes the O(1) running total exact.

What *does* grow is each element's own length varint, and the scenario you describe is real in that form: a unit whose `encoded_len()` crosses 127 → 128 costs `1 + 2 + 128 = 131` bytes instead of `1 + 1 + 127 = 129`. `test_create_message_batch_varint_boundary` covers it — with a budget sized for the 129-byte item, the 131-byte boundary unit is excluded and left in the queue.

While here I also simplified `create_message_batch` itself: the first unit is no longer popped and measured ahead of the loop. The loop starts from an empty batch and identifies the first unit via `batch.batch.is_empty()`, which also let the `send_queue.is_empty()` early return go. An oversized first unit is still included with the advisory `warn!`.

Local verification: `scripts/rust_fmt.sh --check`, `cargo clippy -p apollo_propeller --all-targets --all-features`, `scripts/named_todos.py`, and `SEED=0 cargo test -p apollo_propeller` (167 unit + 4 e2e) all pass.

ShahakShama (approved):
:lgtm:

> @ShahakShama reviewed 2 files and all commit messages, made 1 comment, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14642#-:-OzplVjV-_NNoQjQ6v7k:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on sirandreww-starkware).

PR #14666, opened 2026-06-30
apollo_class_manager: recover orphaned class dir to avoid ENOTEMPTY sync deadlock

matanl-starkware:
## Problem

A node on `starknet-testnet-apollo-3` could not sync: central-sync was permanently stuck retrying a single class, failing every attempt with `Class storage error: Directory not empty (os error 39)` (ENOTEMPTY).

Root cause: `FsClassStorage::set_class` writes a class in two non-atomic steps —
1. `write_class_atomically` renames the staged temp dir into the content-addressed `persistent_dir`, then
2. `mark_class_id_as_existent` commits the existence marker to a separate MDBX store.

`contains_class()` uses only the MDBX marker as the source of truth. If the process crashes between (1) and (2) (e.g. a batcher panic killing the process), the class directory is on disk but the marker is missing. On restart `contains_class()` returns `false`, so `set_class` re-runs (1), but `std::fs::rename` refuses to replace a **non-empty** existing directory and fails with `ENOTEMPTY` — every retry, forever. Sync never advances.

## Fix

Extract a shared `rename_to_persistent_dir` helper used by both `write_class_atomically` and `write_deprecated_class_atomically`. Before the rename, if `persistent_dir` already exists, remove it (with a `warn!`) so the rename can proceed. This is safe: callers reach this code only when the existence marker is absent, and the directory is named by the class hash, so an existing directory holds the same class. The write then completes and the marker is committed, restoring filesystem/marker consistency.

## Tests

- `set_class_recovers_from_orphaned_class_dir` and `set_deprecated_class_recovers_from_orphaned_class_dir` reproduce the exact partial-write state and assert recovery.
- Verified as negative controls: with the recovery removed, both fail with `Os { code: 39, kind: DirectoryNotEmpty }` — the production error.
- `cargo test -p apollo_class_manager class_storage` → 12 passed; `rust_fmt.sh` and `clippy -p apollo_class_manager --all-targets` clean.

### DIFF
--- a/crates/apollo_class_manager/src/class_storage.rs
+++ b/crates/apollo_class_manager/src/class_storage.rs
@@ -24,7 +24,7 @@ use apollo_storage::StorageConfig;
 use starknet_api::class_cache::GlobalContractCache;
 use thiserror::Error;
 use tokio::task::AbortHandle;
-use tracing::instrument;
+use tracing::{instrument, warn};
 
 use crate::metrics::{increment_n_classes, record_class_size, CairoClassType, ClassObjectType};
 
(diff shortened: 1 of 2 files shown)


ron-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14666)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

gkaempfer:

Reviewed: filesystem path construction (content-addressed hex paths, no traversal possible), TOCTOU window in `remove_dir_all` → `rename` sequence (confined to sequencer-owned storage, not externally exploitable), and concurrent-writer behavior (content-addressing ensures idempotent end state). No OWASP Top 10 concerns found.

---

PR #14678, opened 2026-07-01
starknet_transaction_prover: hold worker permit for the proof's full lifetime

avi-starkware:
prove_transaction previously held the admission and worker-slot permits in the
handler future and awaited the prover directly. On a client disconnect the handler
future is dropped, releasing both permits — but the heavy proving work runs on an
un-cancellable blocking thread and keeps running, so a burst of connect-then-
disconnect requests could recycle worker slots and drive real proving above the
concurrency cap.

Run the proof in a task that owns both permits instead. Dropping the JoinHandle on
disconnect detaches the task rather than aborting it, so the proof runs to completion
and releases the permits only when it actually finishes, keeping the worker slot
reserved for the work's true lifetime.

### DIFF
--- a/crates/starknet_transaction_prover/src/server/rpc_impl.rs
+++ b/crates/starknet_transaction_prover/src/server/rpc_impl.rs
@@ -77,8 +77,8 @@ impl ProvingRpcServer for ProvingRpcServerImpl {
         transaction: RpcTransaction,
     ) -> RpcResult<ProveTransactionResult> {
         // Admission: cap queue length (running + waiting). Reject with -32005 only when the queue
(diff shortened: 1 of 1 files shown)


avi-starkware:
* **#14678** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14678?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main-v0.14.3`


<h2></h2>

PR #14682, opened 2026-07-01
echonet: run Starknet OS against replayed cende blobs

ron-starkware:
Add an OS-runner background pipeline: for each blob received, defer-and-enqueue an OS execution over the previous block (whose state_commitment_infos live in the next blob's sliding window). Six worker threads drain a bounded queue, shelling out to committer-and-os-cli's `os run-os-stateless`. Blob bodies stay as raw JSON bytes in memory; a new class_fetcher resolves missing CASM from the mainnet FGW with an on-PVC cache. block_commitments computed once at ingest is threaded through the queue task so workers don't re-parse the multi-MB blob.

Report UI is redesigned around a sidenav + metric-tile Alerts grid + per-section CSV export (rebased from earlier UI-overhaul work). Diagnostic sections only render when their count > 0. Adds an OS run failures section + a divider'd OS-runner subgroup in Alerts showing completed/dropped/abandoned/skipped counts and queue/deferred state.

Deploy: force onto n4-standard-8 via nodeSelector so ScaleOps doesn't downsize into an n4-standard-4; Recreate strategy so the old pod releases the PVC before the new one attaches; --block-hash-cli-binary flag now accepts a path.

The associated Rust-side changes (pin syscall gas costs to 0.14.2 and set the OS's effective Starknet version from block_info) live outside this branch.

### DIFF
--- a/echonet/class_fetcher.py
+++ b/echonet/class_fetcher.py
@@ -1,0 +1,289 @@
+"""
+Resolve the CASM (and deprecated v0) classes the OS needs to run a block.
+
+The cende blob only carries *newly declared* contract classes (`compiled_classes`)
+for the block being run. The OS, however, executes against any class touched by
+the block — including pre-existing on-chain ones. We bridge the gap by fetching
+missing classes from the mainnet feeder gateway, caching them on the echonet
+PVC so subsequent blocks don't refetch the same class.
+
+Inputs:
+- `initial_reads.compiled_class_hashes`: `class_hash → compiled_class_hash` for
+  every Cairo 1+ class touched. The OS's `compiled_classes` map is keyed by
+  `compiled_class_hash`, so this map is exactly the fetch list.
+- `initial_reads.class_hashes`: `address → class_hash` for every accessed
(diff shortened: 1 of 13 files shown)


ron-starkware:
Superseded by an 11-PR stack that splits this branch into reviewable pieces (with a few fixes on top — notably recovering the dropped class-hash keys in `class_hashes_to_migrate` and bounding the commitment store when a gap outlives the blob window):

- #14711 deploy: Recreate strategy + pre-scale-down (PVC multi-attach)
- #14712 blob store: raw bytes + tighter retention
- #14713 stash block-hash commitments at ingest
- #14714 OS CLI input builder
- #14715 class fetcher (feeder + PVC cache)
- #14716 state-commitment-info store + last-stored-height endpoint
- #14717 OS runner worker subprocess
- #14718 OS-run worker pool pipeline
- #14719 report: OS-runner outcomes
- #14720 report UI redesign
- #14721 pod sizing for the OS runner

Original branch content preserved at `ron/echonet-os-squashed`.

PR #14602, opened 2026-06-23
apollo_l1_events,apollo_l1_events_config: chunk L1 getLogs range into bounded windows (M-25)

asaf-sw:
## M-25 — Chunk the L1 getLogs range into bounded windows

**Security review finding:** M-25 (Low–Medium). The L1 events scraper built the inclusive range `scraping_start..=latest_l1_block` and asked the base layer for **all** tracked events across that entire range in a single `events()` / `eth_getLogs` call, with no per-iteration cap and no chunking (a `// If this gets too high, send in batches.` comment acknowledged the gap). After downtime, an L1 outage, or with a large startup rewind, that range can span thousands of L1 blocks. The dominant failure mode is a **liveness stall**: the single call is wrapped in a 1 s timeout plus an N+1 per-event header fetch, so an oversized range times out and the scraper spins in its retry loop, never advancing — and (via the cyclic wrapper) misfires the primary-down-since signal. Unbounded `Vec<L1Event>` / `Vec<Event>` materialization is a secondary memory risk.

### What changed

- **New config param** `l1_events_scraper_config.max_blocks_per_fetch: u64` (`apollo_l1_events_config`), default **1000**, validated `>= 1`.
- **Bounded fetch window** (`apollo_l1_events/src/l1_scraper.rs`): `fetch_events` now requests `scraping_start..=min(latest, start + max_blocks_per_fetch - 1)` and returns the `L1BlockReference` for the window end, so the cursor advances by exactly one window per poll and the steady-state loop drains a backlog over successive polls. The finality ceiling and the once-per-poll reorg check are preserved. `initialize` sends the first bounded window via the provider's `initialize()` and lets the steady loop drain the rest via `add_events`, so the provider's initialize-once contract is unchanged. Saturating arithmetic guards a 0 cap even though validation rejects it.
- **Deployment presets updated.** `l1_events_scraper_config.max_blocks_per_fetch: 1000` added to the hand-maintained `l1_events_scraper_config.json` and `replacer_l1_events_scraper_config.json` presets, and `config_schema.json` regenerated. A new required schema param missing from the presets would `MissingParam`-panic the deployed node at startup (CrashLoopBackOff in `system_test_hybrid`) — exactly the bug L-16 (#14590) hit.

### Assumptions (documented in code)

Default 1000 is conservative versus common public-RPC `eth_getLogs` caps (~1k–10k) and the 1 s timeout; operators on private RPCs may raise it.

### Tests

[...]

PR #14683, opened 2026-07-01
papyrus_base_layer: bisect L1 log range on oversized or timed-out getLogs queries

asaf-sw:
## What
Make `EthereumBaseLayerContract::events()` adaptively bisect its block range when a `getLogs` query fails because the range is too large (or too slow) for the provider. `events()` delegates the raw log fetch to a new `get_logs_bisected` helper: on an oversized/timeout error it splits the range in half and retries each half, down to a single-block floor. The existing per-log header-fetch + `parse_event` step is unchanged and runs once on the concatenated logs.

## Why
Public RPC providers cap `getLogs` responses (e.g. "query returned more than 10000 results", response-size limits, block-range limits, HTTP 413) or time out on wide windows. Previously a wide window that tripped such a cap failed every poll with no way to make progress — the scraper would retry the same window forever. Adaptive bisection lets the same window succeed by fetching it in smaller pieces. This is the base-layer follow-up to the review on #14602 (getLogs range chunking): it addresses the "bisect the window before backing off" and "bound event volume" points, which belong at the base-layer boundary where the oversized/timeout error can actually be classified.

## Behavior
- **Returns all events, in order.** Bisection is iterative (explicit stack, no async recursion) and always fetches the low half before the high half, so concatenated logs stay in ascending block order — identical to today's single-call ordering. No events are dropped; the `CalldataValueOutOfRange`-skip semantics are unchanged.
- **Only bisects the oversized/timeout class.** Classification is a whitelist over the *typed* error (`ErrorPayload.message` substrings for the major providers, provider codes only when paired with a range/results message, and HTTP 413), plus provider timeouts. Auth / rate-limit / connection / decoding errors are propagated unchanged (they are the cyclic-wrapper's or the caller's concern, never bisected).
- **Single-block floor.** If a single-block range still fails with an oversized/timeout error, the error is propagated verbatim (never a truncated/partial result), so the scraper's existing retry and the cyclic wrapper's failover behave exactly as before.

## Tests
Offline (`Asserter`-mocked provider): error classification (whitelist vs excluded variants); bisection split structure and single-block-floor propagation; and `bisection_returns_all_logs_in_ascending_block_order` — real non-empty logs across the split boundary asserting all are returned in ascending block order (completeness + ordering, not just drain counts).

`papyrus_base_layer`: 43 tests pass, clippy clean. Trait signature unchanged, so the scraper and cyclic wrapper are untouched.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9990,6 +9990,7 @@ dependencies = [
  "strum",
  "thiserror 1.0.69",
  "tokio",
+ "tower",
  "tracing",
  "url",
  "validator",
--- a/crates/papyrus_base_layer/Cargo.toml
+++ b/crates/papyrus_base_layer/Cargo.toml
@@ -42,3 +42,4 @@ rstest.workspace = true
 starknet-types-core.workspace = true
 starknet_api = { workspace = true, features = ["testing"] }
 tokio = { workspace = true, features = ["test-util"] }
+tower.workspace = true
--- a/crates/papyrus_base_layer/src/ethereum_base_layer_contract.rs
+++ b/crates/papyrus_base_layer/src/ethereum_base_layer_contract.rs
@@ -8,11 +8,12 @@ use alloy::eips::eip7840;
 use alloy::network::Ethereum;
 use alloy::primitives::{Address, U256};
 use alloy::providers::{Provider, ProviderBuilder, RootProvider};
-use alloy::rpc::json_rpc::RpcError;
+use alloy::rpc::json_rpc::{ErrorPayload, RpcError};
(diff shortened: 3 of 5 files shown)


asaf-sw:
* **#14683** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14683?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14683?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #14689, opened 2026-07-02
apollo_staking: bound next-epoch resolution window by the epoch length

matanl-starkware:
## Problem

`EpochCache::try_resolve_epoch_id` resolves heights up to `MIN_EPOCH_LENGTH` (150) blocks past the current epoch's end as "current epoch + 1". When epochs are shorter than `MIN_EPOCH_LENGTH` (e.g. `MockStakingContract`'s 30-block epochs), heights belonging to later epochs are attributed to a stale epoch ID for up to 150 heights, and the epoch cache re-syncs at a node-specific phase. As a result, `override_committee.start_epoch` is evaluated against the stale label and a committee change takes effect at a **different height on each node**, diverging the round-robin proposer schedules (`(height + round) % len` with different `len`) for several minutes.

## Fix

Bound the next-epoch resolution window by `min(MIN_EPOCH_LENGTH, epoch_length)`:
- Short epochs: the window is exactly one epoch, so resolved epoch IDs are always exact and an override activates at the same height on all nodes. Heights further ahead force an epoch re-sync; if the contract view lags, resolution fails (per-height sync fallback) rather than mislabeling.
- Epochs ≥ `MIN_EPOCH_LENGTH`: behavior unchanged.

## Testing

- New regression test `get_committee_short_epochs_override_activates_at_start_epoch`: with 30-block epochs and the epoch cache synced two epochs before the override's `start_epoch`, the last pre-override height gets the default committee and the first override height gets the override committee. Fails without the fix (stale default committee served past the activation epoch).
- Tightened the existing next-epoch bounds coverage to the exact boundary: `E2_H2` is now `start + MIN_EPOCH_LENGTH` (first invalid height) and `get_committee_for_next_epoch` also asserts the last height within bounds resolves.

### DIFF
--- a/crates/apollo_staking/src/staking_manager.rs
+++ b/crates/apollo_staking/src/staking_manager.rs
@@ -96,7 +96,11 @@ impl Epoch {
 
     fn within_next_epoch_min_bounds(&self, height: BlockNumber) -> bool {
         let next_epoch_start_block = BlockNumber(self.start_block.0 + self.epoch_length);
-        range_contains(height, next_epoch_start_block, MIN_EPOCH_LENGTH)
+        // Only heights guaranteed to fall within the next epoch may resolve to `epoch_id + 1`.
+        // The next epoch's length is unknown until it starts, so bound the window by the
+        // smallest known lower bound on epoch length.
+        let next_epoch_min_length = MIN_EPOCH_LENGTH.min(self.epoch_length);
+        range_contains(height, next_epoch_start_block, next_epoch_min_length)
     }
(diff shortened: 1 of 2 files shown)


dafnamatsry (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14689)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

gkaempfer:

---

PR #14710, opened 2026-07-07
apollo_transaction_converter: verify sierra/executable consistency in tx conversion (M-33)

noamsp-starkware:
## Finding
M-33 (Medium, Mempool) — Sequencer Security Report v1.0. `convert_internal_rpc_tx_to_executable_tx` derived Sierra version, program length, and ABI length from class-manager output without consistency checks against the stored compiled class.

## Change
`get_sierra` and `get_executable` are two independent class-manager reads; nothing guarantees they returned artifacts for the same compiled class, yet the Sierra-derived lengths and version are then trusted as authoritative billing / version-gating metadata for the separately-fetched executable.

- Add `verify_class_consistency`: recompute the class hash from the fetched Sierra and the compiled-class hash from the fetched executable, and confirm both are keyed to the tx's `class_hash` / `compiled_class_hash` before building `ClassInfo`. Return a descriptive error on mismatch (never panic).

This is defense-in-depth: under content-addressing + atomic writes a mismatch is not externally attacker-inducible today (it would require a compiler bug or storage corruption), but the check fails closed rather than admitting inconsistent metadata.

## Testing
- New tests: consistent pair converts successfully; Sierra-hash mismatch and compiled-class-hash mismatch each return the expected error.
- `SEED=0 cargo test -p apollo_transaction_converter` → 8 passed.

### DIFF
--- a/crates/apollo_gateway/src/errors.rs
+++ b/crates/apollo_gateway/src/errors.rs
@@ -340,6 +340,22 @@ pub fn transaction_converter_err_to_deprecated_gw_err(
         TransactionConverterError::ProofNotFound { .. } => {
             StarknetError::internal_with_signature_logging("Proof not found", tx_signature, err)
         }
+        // Internal error: the class manager returned artifacts that are inconsistent with the
+        // requested class (e.g. stale cache or partial write), not a fault in the user's tx.
+        TransactionConverterError::SierraClassHashMismatch { .. } => {
+            StarknetError::internal_with_signature_logging(
+                "Inconsistent Sierra class hash",
+                tx_signature,
+                err,
+            )
+        }
+        TransactionConverterError::UnexpectedCairo0Executable { .. } => {
+            StarknetError::internal_with_signature_logging(
+                "Unexpected Cairo 0 executable",
+                tx_signature,
+                err,
+            )
+        }
         TransactionConverterError::StarknetApiError(err) => convert_sn_api_error(err),
     }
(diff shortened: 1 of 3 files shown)


noamsp-starkware:
Addressed the review's main finding (compiled-class-hash branch reusing the client-facing error) in 26ad0bd715:

- The third check in `verify_class_consistency` previously returned `ValidateCompiledClassHashError`, which maps to `InvalidCompiledClassHash` — a client-facing, unlogged "you supplied a bad value" error. But the tx's `compiled_class_hash` is already validated against the freshly-compiled class at ingestion (`convert_rpc_tx_to_internal`), so a mismatch here can only be a class-manager inconsistency — purely internal, exactly the M-33 scenario this check exists to catch.
- Added a dedicated `CompiledClassHashMismatch` variant mapped through `internal_with_signature_logging` (logs at `error!`, returns a generic internal-error code), parallel to `SierraClassHashMismatch` / `UnexpectedCairo0Executable`. Now all three checks in the function behave consistently: log the internal fault, don't blame the client.
- Updated `test_convert_internal_rpc_tx_to_executable_tx_declare_compiled_class_hash_mismatch` to assert the new variant. `SEED=0 cargo test -p apollo_transaction_converter` → 8 passed.

On the two minor notes: the re-hash cost on the batcher/gateway paths is acceptable (declares are rare); the V0 guard is intentional and stays.

PR #14222, opened 2026-05-27
starknet_transaction_prover,tower_ohttp: OHTTP-unlinkable request-id for decapsulated content

avi-starkware:
Tags downstream content logs with a request-id via a new `RequestSpanLayer`
placed below the OHTTP layer. For plaintext it reuses the envelope id from
`RequestLogLayer`; for an OHTTP-decapsulated request (marked with a new
`tower_ohttp::Decapsulated` extension) it mints a FRESH UUID. Any
client-supplied inner `x-request-id` is stripped by `tower_ohttp` at
decapsulation time, so the property holds even if the marker is ever lost
(fail-closed).

The fresh inner id is never echoed back, so the relay-visible envelope id and
the gateway's content-log id share no join key — preserving OHTTP
unlinkability while still giving every request's downstream logs a
correlatable id. Both integration tests exercise the production
`prover_http_middleware!` chain, so a layer reorder that reverts either
property fails CI.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

PR #14743, opened 2026-07-09
starknet_transaction_prover: avoid redundant request-id re-parse in RequestSpanLayer

gkaempfer:
## What

Follow-up perf cleanup to #14222, which added `RequestSpanLayer` (`crates/starknet_transaction_prover/src/server/request_span.rs`) below the OHTTP layer to bind a tracing span with a request-id over every HTTP request handled by the transaction-prover server.

For plaintext (non-OHTTP) requests, `RequestSpanLayer` called `extract_or_generate_request_id(&request)`, which re-parses and re-validates the `x-request-id` header (UTF-8 conversion, length check, byte-by-byte ASCII-safety scan, `String` allocation) on **every single request** — even though the upstream `RequestLogLayer` (`request_log.rs`), earlier in the same middleware chain (`prover_http_middleware!` in `server.rs`), already computed and validated that exact id moments earlier.

## Fix

`RequestLogLayer::call` now stashes the already-validated id in a new `RequestId(String)` request extension right after computing it. `RequestSpanLayer::call` reads that extension first, and only falls back to `extract_or_generate_request_id` when the extension is absent — e.g. when `RequestSpanLayer` is exercised standalone in unit tests without `RequestLogLayer` upstream. The OHTTP-decapsulated branch (fresh envelope-unlinkable id) is untouched — it never reads this extension.

This is safe because request extensions survive `MapRequestBodyLayer` (`http::Request::map` preserves them) and the OHTTP layer's non-OHTTP passthrough route (`inner.oneshot(request)` with the request unchanged), so the extension set by the outer `RequestLogLayer` reaches the inner `RequestSpanLayer` unmodified in the production chain.

## Tests

- Added `plaintext_prefers_request_id_extension_over_header`, which sets the `RequestId` extension and the `x-request-id` header to two different values (a state `RequestLogLayer` never produces in practice) and asserts, via a traced log line emitted inside the span, that the extension's value — not the header's — is what the span carries. This locks in the fast path so a future refactor can't silently regress back to header re-parsing.
- All pre-existing `starknet_transaction_prover` / `tower_ohttp` tests pass (122 passed, 0 regressions; the only failure, `test_compiled_class_v1_to_casm_round_trip`, is a pre-existing, network-dependent Cairo-compiler-download failure unrelated to this change and reproduces identically on the unmodified base branch).
- `cargo clippy -p starknet_transaction_prover -p tower_ohttp --all-targets` is clean.
- `scripts/rust_fmt.sh` applied, no outstanding diffs.

An independent Opus review confirmed correctness (no cross-request leak — the extension lives in per-request `http::Extensions`, not shared service state; OHTTP unlinkability is fully preserved) and flagged two style/coverage suggestions, both addressed in this PR (`map_or_else` instead of `map().unwrap_or_else()`, and the extension-vs-header divergence test above).

---
_Generated by [Claude Code](https://claude.ai/code/session_01F3mvGUJMwN5w18HG58LXjy)_

### DIFF
--- a/crates/starknet_transaction_prover/src/server/request_log.rs
+++ b/crates/starknet_transaction_prover/src/server/request_log.rs
@@ -41,6 +41,13 @@ mod request_log_test;
 /// HTTP header carrying the request id.
 pub const REQUEST_ID_HEADER: &str = "x-request-id";
 
+/// Request extension carrying the id this layer already validated/generated,
+/// so a downstream layer (e.g. `RequestSpanLayer`) can reuse it on the
+/// plaintext path instead of re-parsing and re-validating the header it just
+/// set.
+#[derive(Clone)]
+pub(crate) struct RequestId(pub String);
+
 /// tower [`Layer`] producing [`RequestLogService`].
 #[derive(Clone, Copy, Default)]
 pub struct RequestLogLayer;
@@ -78,6 +85,7 @@ where
         let request_id = extract_or_generate_request_id(&request);
         let id_header_value = request_id_header_value(&request_id);
         request.headers_mut().insert(REQUEST_ID_HEADER, id_header_value.clone());
+        request.extensions_mut().insert(RequestId(request_id.clone()));
         let is_health_probe =
             request.method() == Method::GET && request.uri().path() == HEALTH_PATH;
         let method = request.method().clone();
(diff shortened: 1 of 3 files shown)


avi-starkware (approved):
:lgtm:

> @avi-starkware reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14743#-:-Ox6erjG3yk3u8ri1k-J:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on gkaempfer).

gkaempfer:

---

PR #14717, opened 2026-07-07
echonet: add standalone OS runner worker subprocess

ron-starkware:
echo_center will spawn this per OS run: it loads the task file, builds
the OsCliInput (filling the class maps via class_fetcher), invokes
committer-and-os-cli 'os run-os-stateless', and exits — so captured CLI
output, intermediate JSON dicts, and the lazily imported cairo-lang all
die with the process instead of accumulating in echo_center's heap. On
failure the OS input is dumped under /data/echonet/os_runs/ with a
rolling retention of 10. Adds OsRunnerConfig and the cairo-lang pod
dependency (component-hash derivation for declared classes).

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

PR #14715, opened 2026-07-07
echonet: add class fetcher resolving classes for OS runs from the feeder gateway

ron-starkware:
The cende blob only carries the classes newly declared in its block,
but the OS executes against every class the block touches. Resolve the
rest from the mainnet feeder gateway with an on-PVC cache: Cairo 1 CASM
by compiled_class_hash (zero-valued sentinel entries in
initial_reads.compiled_class_hashes mark Cairo 0 classes), deprecated
v0 classes by class_hash, and Sierra-shaped v0 responses skipped since
the FGW's cairo-lang CASM serialization does not match the Rust schema.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

PR #14835, opened 2026-07-16
echonet: validate block_number as int in os_runner_worker to close path traversal

gkaempfer:
## Summary

Automated security scan of #14717 found that `echonet/os_runner_worker.py` takes `block_number` straight from the `--input-path` JSON payload and interpolates it, unvalidated, into the failed-run dump directory name:

```python
block_number = payload["block_number"]
...
dump_dir = dump_root / f"block_{block_number}_failed"
dump_dir.mkdir(parents=True, exist_ok=True)
(dump_dir / "input.json").write_bytes(input_path.read_bytes())
```

`pathlib` splits on `/` when joining path components, so a malformed `block_number` (e.g. containing `../`) lets the resulting directory/file write escape the intended `os_runs` sandbox. Every other `block_number` ingestion point in `echonet/` (e.g. `echo_center.py`'s blob/write-blob handling) already casts this field with `int(...)`; `os_runner_worker.py` was the one place that didn't, despite `build_os_cli_input`'s own signature declaring `block_number: int`.

- Not currently reachable in production: the caller that would wire `echo_center` to spawn this subprocess hasn't landed yet (per #14717's own commit message: "echo_center *will* spawn this per OS run"). This closes the gap defensively before that wiring PR lands, rather than relying on the future caller to remember to sanitize.

## Fix

`echonet/os_runner_worker.py`: cast `block_number` to `int` right at extraction, matching the established convention elsewhere in `echonet/`.

## Test

`echonet/tests/test_os_runner_worker.py` (new):
- `test_path_traversal_block_number_is_rejected_before_any_side_effect`: feeds a `block_number` of `"../../../../tmp/echonet_poc_pwned"` through the failed-run dump path. Verified this test **fails** against the pre-fix code — the traversal write actually lands at `<log_dir>/tmp/echonet_poc_pwned_failed/input.json`, outside the intended `<log_dir>/echonet/os_runs/` sandbox. With the fix, `int()` raises `ValueError` before any of `build_os_cli_input`, the CLI subprocess, or the dump write are reached.
- `test_valid_block_number_still_succeeds`: regression check that a normal numeric `block_number` (including as a JSON string) still flows through correctly as an `int`.

```
python3 -m unittest echonet.tests.test_os_runner_worker -v
```
both pass with the fix; the traversal test fails on the pre-fix code.

Refs #14717.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Kqhp1aQVT33tneWqKYxdni)_

### DIFF
--- a/echonet/echonet_types.py
+++ b/echonet/echonet_types.py
@@ -164,6 +164,36 @@ class PathsConfig:
 
 
 @dataclass(frozen=True, slots=True)
+class OsRunnerConfig:
+    """
+    Configuration for running the Starknet OS over each received blob, via the
+    block-hash CLI binary's `os run-os-stateless` subcommand (which must be
+    built with the `os_input` and `transaction_serde` features).
+    """
+
+    enabled: bool = True
+    layout: str = "all_cairo"
+    cli_timeout_secs: int = 600
+    chain_id: str = "SN_MAIN"
+    strk_fee_token_address: str = (
+        "0x4718f5a0fc34cc1af16a1cdee98ffb20c31f5cd61d6ab07201858f4287c938d"
+    )
+    # Bounded backlog of OS-run tasks; tasks beyond it are dropped (logged) so the
+    # Flask handler never blocks blob storage.
+    queue_size: int = 30
+    # Worker threads draining the queue, one OS CLI subprocess each. At ~5 runs/min
+    # per worker vs ~26 mainnet blocks/min, ≥6 are needed; 8 also uses the pod's
+    # cpu-limit headroom under burst while staying within the memory limit.
+    parallelism: int = 8
+    # Width of the blob's `recent_state_commitment_infos` vector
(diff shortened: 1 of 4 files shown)


gkaempfer:
Independent Opus review confirmed this fix (casting `block_number` to `int`) fully closes the path-traversal it targets, and the regression tests are solid (verified to fail on pre-fix code, pass after).

One related-but-separate finding surfaced during that review, out of scope for this PR: `echonet/class_fetcher.py`'s `_read_cache`/`_write_cache` build cache file paths from `class_hash`/`compiled_class_hash` keys taken verbatim from the blob JSON, with no validation that they're canonical hex felts — the same traversal pattern (`cache_dir / f"{key}.json"`) as the bug fixed here, but for a field this PR doesn't touch (introduced by #14715). Flagging so it can be tracked as a follow-up rather than silently dropped.

---

PR #14855, opened 2026-07-21
apollo_dashboard: tolerate L1 pod spot eviction in L1 provider alerts

matanl-starkware:
## Problem

We receive sporadic `batcher_l1_events_provider_errors` pages caused by infrastructure churn, not a real fault. In the distributed topology the batcher reaches the L1 **events** provider — and consensus reaches the L1 **gas-price** provider — over HTTP to the same `L1` service pod (`crates/apollo_deployments/src/deployments/distributed.rs:54-55,99-100`). That pod runs on a **spot node pool** and is periodically evicted. While it's gone / rescheduling / catching up, the remote clients exhaust their retries and increment the error counters, easily crossing the old thresholds and paging on routine churn.

## Fix

Make the two L1-pod-dependent **paging** alerts fire only when errors persist *longer than the expected pod bootup window* — a genuine sustained outage — rather than on the normal eviction+reboot cycle. Tolerance is expressed via the Grafana `for`/pending-duration field (shared `L1_POD_BOOTUP_TOLERANCE = "3m"`) over a short `[1m]` increase window that clears within a minute of recovery so the `for` timer resets on a normal eviction.

| Alert | Before | After |
|---|---|---|
| `batcher_l1_events_provider_errors` | `increase[10m] > 10`, `for=30s`, p4 | `increase[1m] > 0`, `for=3m`, **p2** |
| `consensus_l1_gas_price_provider_failure` | `increase[1h] > 5`, `for=30s`, p4 | `increase[1m] > 0`, `for=3m`, p4 |

- `batcher_l1_events_provider_errors` escalated to **p2**, resolving the inline `TODO(Arni)` about its severity. A transient error here is harmless anyway (the propose path falls back to an empty L1-handler list and still builds the block).
- `consensus_l1_gas_price_provider_failure` **keeps p4**; it gains the same tolerance shape.
- `consensus_l1_gas_price_provider_failure_once` (p5, non-paging) is unchanged — it remains the immediate early signal.

Alert-file-only change (small blast radius): no new metrics, no batcher/consensus/`apollo_infra` code.

## Verification

- `cargo run --bin sequencer_dashboard_generator -q` — `dev_grafana_alerts.json` regenerated (only the fields above change).
- `SEED=0 cargo test -p apollo_dashboard` — 16/16 pass, incl. the `default_dev_grafana_dashboard` golden-file guard.
- `cargo clippy -p apollo_dashboard --all-targets` clean; `rust_fmt.sh` clean.

### DIFF
--- a/crates/apollo_dashboard/resources/dev_grafana_alerts.json
+++ b/crates/apollo_dashboard/resources/dev_grafana_alerts.json
@@ -70,12 +70,12 @@
       "name": "batcher_l1_events_provider_errors",
       "title": "Batcher L1 events provider errors",
       "ruleGroup": "evaluation_rate_default",
-      "expr": "(sum(increase(batcher_l1_events_provider_errors{cluster=~\"$cluster\", namespace=~\"$namespace\"}[10m])) or vector(0)) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)",
+      "expr": "(sum(increase(batcher_l1_events_provider_errors{cluster=~\"$cluster\", namespace=~\"$namespace\"}[1m])) or vector(0)) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)",
       "conditions": [
         {
           "evaluator": {
             "params": [
-              10.0
+              0.0
             ],
             "type": "gt"
           },
@@ -89,8 +89,8 @@
           "type": "query"
         }
       ],
-      "for": "30s",
-      "severity": "p4"
+      "for": "3m",
+      "severity": "p2"
     },
     {
       "name": "batcher_remote_server_number_of_connections",
@@ -564,12 +564,12 @@
       "name": "consensus_l1_gas_price_provider_failure",
       "title": "Consensus L1 gas price provider failure",
       "ruleGroup": "evaluation_rate_default",
-      "expr": "(sum(increase(consensus_l1_gas_price_provider_error{cluster=~\"$cluster\", namespace=~\"$namespace\"}[1h])) or vector(0)) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)",
+      "expr": "(sum(increase(consensus_l1_gas_price_provider_error{cluster=~\"$cluster\", namespace=~\"$namespace\"}[1m])) or vector(0)) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)",
       "conditions": [
         {
           "evaluator": {
             "params": [
-              5.0
+              0.0
             ],
             "type": "gt"
           },
@@ -583,7 +583,7 @@
           "type": "query"
         }
       ],
-      "for": "30s",
+      "for": "3m",
       "severity": "p4"
     },
     {
--- a/crates/apollo_dashboard/src/alert_definitions.rs
+++ b/crates/apollo_dashboard/src/alert_definitions.rs
@@ -131,6 +131,14 @@ pub fn get_dev_alerts_json_path() -> String {
 
 // TODO(shahak): Move the remaining alerts here into modules.
 
+/// The batcher and consensus reach their L1-hosted providers (`l1_events`, `l1_gas_price`) over
+/// HTTP; in distributed deployments the L1 service runs on a spot node pool and is periodically
+/// evicted. A normal eviction + reschedule + L1-provider catch-up recovers within a few minutes,
+/// during which these consumers log transient errors. The paging alerts below tolerate that by
+/// requiring the error condition to hold continuously for this duration, so routine spot churn
+/// stays silent and only a genuine sustained L1 outage pages.
+const L1_POD_BOOTUP_TOLERANCE: &str = "3m";
+
 fn get_consensus_decisions_reached_by_consensus_ratio() -> Alert {
     Alert::new(
         "consensus_decisions_reached_by_consensus_ratio",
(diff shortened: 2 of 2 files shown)


dan-starkware:
*[`crates/apollo_dashboard/src/alert_definitions.rs` line 257 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14855#-Oy7hB5BBfooJLR1r2gC:-Oy7hB5BBfooJLR1r2gD:brkkw34) ([raw file](https://github.com/starkware-libs/sequencer/blob/b5667b2caee11f8a8e385d7f9793bdbfa41776ae/crates/apollo_dashboard/src/alert_definitions.rs#L257)):*
> ```XML
>             "sum(increase({}[1m])) or vector(0)",
>             CONSENSUS_L1_GAS_PRICE_PROVIDER_ERROR.get_name_with_filter()
>         ),
> ```

Is it guaranteed that each client increment its error counter at least once per minute while the provider is down ? Can we have a  long backoff?



_Code quote:_
```XML
        // 1-minute window: reflects whether consensus is *currently* erroring and clears within a
        // minute of recovery, so the `for` timer resets on a normal eviction.
        format!(
            "sum(increase({}[1m])) or vector(0)",
            CONSENSUS_L1_GAS_PRICE_PROVIDER_ERROR.get_name_with_filter()
        ),
```

danbrownstein (commented):
> @dan-starkware made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14855)* status: 0 of 2 files reviewed, 3 unresolved discussions (waiting on matanl-starkware).

___
*[`crates/apollo_dashboard/src/alert_definitions.rs` line 139 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14855#-Oy7lChk39nlLaI5QBoX:-Oy7lChk39nlLaI5QBoY:brv6pjy) ([raw file](https://github.com/starkware-libs/sequencer/blob/b5667b2caee11f8a8e385d7f9793bdbfa41776ae/crates/apollo_dashboard/src/alert_definitions.rs#L139)):*
> ```XML
> /// during which these consumers log transient errors. The paging alerts below tolerate that by
> /// requiring the error condition to hold continuously for this duration, so routine spot churn
> /// stays silent and only a genuine sustained L1 outage pages.
> ```

Consider



_Suggestion:_
```XML
/// The L1 provider components can run on a spot pod that gets evicted, and clients see errors until
/// it comes back. This is how long we wait for that before paging, so a routine eviction stays quiet
/// and only a lasting outage pages.
```
___
*[`crates/apollo_dashboard/src/alert_definitions.rs` line 406 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14855#-Oy7m8rRDxfqeRyGNE1c:-Oy7m8rRDxfqeRyGNE1d:bifwzav) ([raw file](https://github.com/starkware-libs/sequencer/blob/b5667b2caee11f8a8e385d7f9793bdbfa41776ae/crates/apollo_dashboard/src/alert_definitions.rs#L406)):*
> ```XML
>             "sum(increase({}[1m])) or vector(0)",
>             BATCHER_L1_EVENTS_PROVIDER_ERRORS.get_name_with_filter()
>         ),
> ```

Consider defining the `1m` as const and remove the comment from here (same above)



_Code quote:_
```XML
        // 1-minute window: reflects whether the batcher is *currently* erroring and clears within
        // a minute of recovery, so the `for` timer resets on a normal eviction.
        format!(
            "sum(increase({}[1m])) or vector(0)",
            BATCHER_L1_EVENTS_PROVIDER_ERRORS.get_name_with_filter()
        ),
```

matanl-starkware (commented):
> @matanl-starkware made 1 comment and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14855)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on dan-starkware).

___
*[`crates/apollo_dashboard/src/alert_definitions.rs` line 257 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14855#-Oy7hB5BBfooJLR1r2gC:-Oy7tCys3lUBQfC7XlXo:b-7eldk) ([raw file](https://github.com/starkware-libs/sequencer/blob/b5667b2caee11f8a8e385d7f9793bdbfa41776ae/crates/apollo_dashboard/src/alert_definitions.rs#L257)):*
<details><summary><i>Previously, dan-starkware wrote…</i></summary><blockquote>

Is it guaranteed that each client increment its error counter at least once per minute while the provider is down ? Can we have a  long backoff?
</blockquote></details>

These clients run with retries=0, so there's no backoff loop today. It also stays robust if that changes: on an eviction the pod refuses connections, so every attempt fails in ≤500 ms (connection_timeout) — even at the default 15 retries that's ~14 s/call, well under the 1 m window. 
Consensus queries the gas price per height, so increase[1m] stays > 0 across the 3 m pending window and the alert fires. 
The only gap would be a provider that stays connected but hangs the full 30 s request_timeout on every attempt — a different failure from eviction, and one that stalls block production so it trips the block-stuck alerts anyway.

danbrownstein (approved):
:lgtm:

> @dan-starkware reviewed 2 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14855#-:-Oy81hY4DtFrBbSYOMEo:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

PR #14752, opened 2026-07-10
apollo_dashboard: fall back to vector(1) on no-data for consensus_block_number_stuck (#14705)

matanl-starkware:
Backport of #14705 to `main-v0.14.3`.

Clean `git cherry-pick -x` of `d6d91ab8b5`. No `starknet_version` change (`V0_14_3` preserved).

Generated `dev_grafana_alerts.json` regenerated and verified identical (no-op).

PR #14705, opened 2026-07-06
apollo_dashboard: fall back to vector(1) so no-data doesn't fire consensus_block_number_stuck

matanl-starkware:
## What
Change the `consensus_block_number_stuck` no-data fallback `or vector(0)` → `or vector(1)` (both the `stuck` and `stuck_long_time` variants) in `get_consensus_block_number_stuck`.

## Why
On 2026-07-06 this alert paged **p1** and flapped on `apollo-mainnet-0` while the node was fully healthy (`up`=1, ~24 blocks/min, 0 restarts, committing throughout). Grafana state history showed `A=0.000000` at each fire and `A≈47` when healthy — including a ~9-min continuous Alerting stretch on a healthy chain.

Root cause: the rule is `(sum(increase(consensus_block_number[120s])) or vector(0)) and on() (is_observer==0) < 1`. At Grafana's real-time evaluation, `increase([120s])` intermittently returned an **empty** result — it needs ≥2 fresh samples in the trailing 120s, and node-0 is the lone **us-west1** node (nodes 1–5 are us-east1), so its freshest samples are occasionally not yet queryable via the central Grafana/GMP query path (the *stored* series is complete, which is why the raw metric never "drops"). `or vector(0)` converts that empty into a hard `0`, so `0 < 1` → false page. The pattern is chronic and fleet-wide (testnet flaps identically).

## Fix
Use `or vector(1)`. PromQL `or` is a set operator on series *presence* (not boolean on values), so:

| Case | `or vector(0)` (before) | `or vector(1)` (after) | fires (`< 1`)? |
|---|---|---|---|
| Healthy chain | 48 | 48 | no → no |
| Real halt (present `0`) | 0 | 0 | **yes → yes** (unchanged, no added latency) |
| Absent series (query no-data) | 0 | 1 | **yes (false) → no** (the fix) |

A real halt keeps the gauge present, so `increase` yields a present `0` and still fires immediately. A genuinely dead pod remains covered by the `pod_state_not_ready` / `pod_state_crashloopbackoff` alerts.

## Verification
- `cargo build -p apollo_dashboard`; `cargo run --bin sequencer_dashboard_generator -q` (regenerated `dev_grafana_alerts.json`); `SEED=0 cargo test -p apollo_dashboard` (16 pass, incl. the `dashboard_definitions` fixture check); rustfmt clean.
[...]

PR #14876, opened 2026-07-23
apollo_dashboard: fall back to vector(1) so no-data doesn't fire add_tx idle alerts

matanl-starkware:
## Problem

At ~16:50 IDT on 2026-07-22 the p2 alert `apollo-sepolia-alpha-3-gateway_add_tx_idle_(p2p+rpc)` fired on Testnet, but the tx-rate panel showed no dip.

The three idle alerts built by `build_idle_alert` — `gateway_add_tx_idle_p2p_rpc` (p2), `mempool_add_tx_idle_p2p_rpc` (p1), `http_server_no_successful_transactions` (p5) — use:

```
sum(increase(<metric>[Ws])) or vector(0)   <   0.1   for 30s
```

A transient metric **scrape/ingestion gap** makes `increase(...)` return *empty*, and `or vector(0)` turns that into a hard **0**, which trips `< 0.1` and pages — even while tx ingestion is perfectly healthy.

## Evidence (apollo-sepolia-alpha-3, 2026-07-22 ~13:48–13:51 UTC)

- `count_over_time(gateway_transactions_received[1m])`: 36 → 18 → **absent 13:49–13:51** → 16 → 34 (a ~3-min scrape gap; counter caught up +404 at 13:51:30).
- `up{...gateway-pod-monitoring}` stayed `1`; the pod never restarted; RPC "Recorded transaction" logs were continuous (35 txs at 13:50).
- Backfilled data + `increase` interpolation now read ~230 tx/2min straight through the gap → the Grafana panel looks fine, which is why the false page is invisible after the fact.

So the pipeline was healthy; only the metric plane blipped, and `or vector(0)` converted "can't compute" into "exactly zero".

## Fix

Flip the shared helper's fallback `or vector(0)` → `or vector(1)` (`crates/apollo_dashboard/src/alert_scenarios/tps.rs`). An unscrapeable series reads as `1` (≥ the `0.1` threshold) → no page; a **real** idle keeps the counter present so `increase` returns a present `0` → still fires. Zero added latency.

This is the same fix already shipped for `consensus_block_number_stuck` (#14705, backport #14752) in `block_production_halt.rs`; it was never propagated to the `tps.rs` idle alerts.

## Verification

- Regenerated `dev_grafana_alerts.json` via `sequencer_dashboard_generator` — diff is exactly the three idle-alert expressions flipped `vector(0)` → `vector(1)`, nothing else.
- `SEED=0 cargo test -p apollo_dashboard` — 16 passed, 0 failed (incl. the `default_dev_grafana_dashboard` fixture round-trip).

### DIFF
--- a/crates/apollo_dashboard/resources/dev_grafana_alerts.json
+++ b/crates/apollo_dashboard/resources/dev_grafana_alerts.json
@@ -980,7 +980,7 @@
       "name": "gateway_add_tx_idle_p2p_rpc",
       "title": "Gateway add_tx idle (p2p+rpc)",
       "ruleGroup": "evaluation_rate_default",
-      "expr": "(sum(increase(gateway_transactions_received{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$$$gateway_add_tx_idle_p2p_rpc-sampling_window_secs-expression$$$s])) or vector(0)) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)",
+      "expr": "(sum(increase(gateway_transactions_received{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$$$gateway_add_tx_idle_p2p_rpc-sampling_window_secs-expression$$$s])) or vector(1)) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)",
       "conditions": [
         {
           "evaluator": {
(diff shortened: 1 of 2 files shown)


ron-starkware (approved):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14876)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

PR #14899, opened 2026-07-30
starknet_committer,starknet_committer_and_os_cli: rust-to-python serde test variants for commitment infos

itamar-starkware:
## What

1. **Ungates the `StateCommitmentInfos` compression codec** — `compress`/`decompress` (bincode + zstd) are no longer behind `os_input`, so any build can produce and read the compressed form.
2. **Adds CLI `python-test` serde variants**: `state_commitment_infos_serialize_test` emits one deterministic `StateCommitmentInfos` in *both* wire forms (plain JSON and the base64-zstd-bincode payload shipped in the cende blob); `compressed_state_commitment_infos_decompress_test` takes that payload and returns the plain JSON form. Plus `state_maps_serialize_test` on the OS CLI for the block's initial reads.

## Why

The starkware repo now decodes the compressed commitment infos in python (the recorder stores them; the OS input will consume them). Its test needs rust-produced bytes that follow the pinned sequencer commit — the CLI binary is already pinned by `sequencer_commit_main()`, so these variants give python a cross-language oracle with no checked-in fixtures and no image work.

Ungating the codec is what lets that test run in ordinary per-PR CI: the pinned CLI is built **without** `os_input`, and it still needs to compress/decompress.

## Verification

Built the CLI from this branch and ran both variants; a python decoder of the compressed form reproduces rust's own JSON form exactly — all 4 commitment infos, including a zero felt and a 185-bit felt (the fixture deliberately covers the variable-length felt encoding), 229 B bincode → 107 B compressed. The decompress variant round-trips the same payload back to the identical JSON. `cargo test -p starknet_committer`: 124 passed. `rust_fmt --check` clean.

The consuming python test is in the starkware repo (`state_commitments_test.py`), gated on this landing plus a pin bump.

### DIFF
--- a/crates/starknet_committer/Cargo.toml
+++ b/crates/starknet_committer/Cargo.toml
@@ -7,14 +7,14 @@ license.workspace = true
 description = "Computes and manages Starknet state."
 
 [features]
-os_input = ["dep:bincode", "dep:blake2", "dep:digest", "dep:zstd"]
+os_input = ["dep:blake2", "dep:digest"]
 testing = ["starknet_patricia/testing"]
 
 [dependencies]
 apollo_config.workspace = true
 async-trait.workspace = true
 base64.workspace = true
-bincode = { workspace = true, optional = true }
+bincode.workspace = true
 blake2 = { workspace = true, optional = true }
 blockifier.workspace = true
 derive_more = { workspace = true, features = ["as_ref", "from", "into"] }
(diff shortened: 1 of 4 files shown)


itamar-starkware:
Superseded by the rebuilt os_input stack: the CLI serde python-test variants now sit on top of the flag-removal commit, which removes the `os_input` gates outright — so the separate ungating commit in this PR is unnecessary. Will re-open the variants as part of that stack.

Verified along the way: the CLI emits both wire forms of the same `StateCommitmentInfos`, and a python decoder of the compressed form reproduces rust's JSON form exactly (4 commitment infos, including a zero felt and a 185-bit felt).

PR #14912, opened 2026-08-03
blockifier: avoid double-cloning initial reads in get_os_initial_reads

gkaempfer:
## Summary

Follow-up performance optimization on #14911, which introduced `OsInitialReadsCollection` to make OS-initial-reads collection opt-out for reexecution.

`CachedState::get_os_initial_reads` runs once per block during block finalization in production (`os_input` feature, used by `apollo_batcher`). It previously called `self.get_initial_reads()` — a full deep clone of the `StateMaps` struct (five `HashMap`s: `nonces`, `class_hashes`, `storage`, `compiled_class_hashes`, `declared_contracts`) — **twice**: once purely to derive two small key collections (accessed contract addresses and accessed class hashes) needed to drive the force-read loop, and again at the end for the actual return value.

The `storage` map in particular can hold thousands of entries per block, so the first clone was pure wasted work — it never even touched the storage values, only needed derived key sets.

This change computes the two derived collections (`HashSet<ContractAddress>`, `Vec<ClassHash>`) from a borrowed reference to the cache instead, dropping the borrow before the force-read loop runs (which needs `&mut self`), and removes the first full clone entirely. The second clone (needed for the final, updated result) is unchanged.

## Test plan
- [x] `cargo build -p blockifier --features os_input`
- [x] `SEED=0 cargo test -p blockifier --features os_input --lib state::cached_state` (43 passed)
- [x] `SEED=0 cargo test -p blockifier --features os_input --lib transaction_executor` (47 passed)
- [x] Reviewed by an independent agent for correctness (no double-borrow / `RefCell` panic risk, behavior parity verified against pre-change code) and style; no blocking or suggestion-level issues found

---
_Generated by [Claude Code](https://claude.ai/code/session_012Sq6TBm8eS6vgrhnop6uVT)_

### DIFF
--- a/crates/blockifier/src/state/cached_state.rs
+++ b/crates/blockifier/src/state/cached_state.rs
@@ -303,15 +303,24 @@ impl<S: StateReader> CachedState<S> {
         // Back-fill write-only storage cells so their pre-block values are part of the initial
         // reads.
         self.update_initial_values_of_write_only_access()?;
-        let raw_initial_reads = self.get_initial_reads()?;
+        // Collect only the keys needed below, instead of cloning the entire (potentially large)
+        // initial-reads map twice per block.
+        let (contract_addresses, declared_class_hashes) = {
+            let cache = self.cache.borrow();
+            let initial_reads = &cache.initial_reads;
+            (
+                initial_reads.get_contract_addresses(),
+                initial_reads.declared_contracts.keys().copied().collect::<Vec<_>>(),
+            )
+        };
         // Force-read the contract-trie and class-trie leaves so their pre-block values are cached
         // in the initial reads.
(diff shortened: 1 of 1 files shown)


yoavGrs (commented):
> @yoavGrs made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14912)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on gkaempfer).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/14912#-Oz5X2Qn1TnuFFq6R6Pz:-Oz5X2Qn1TnuFFq6R6Q-:b5tkzrj) (no related file):*
Already in our baclog:
https://reviewable.io/reviews/starkware-libs/sequencer/14624

PR #14913, opened 2026-08-03
apollo_storage,apollo_batcher,apollo_batcher_types,apollo_consensus_orchestrator: avoid full blob read for state commitment infos existence check

gkaempfer:
## Summary

Follow-up performance optimization to #14895 (apollo_consensus_orchestrator: verify retrospective state commitment infos before proposing).

That PR added `verify_retrospective_state_commitment_infos`, which runs once per block on the proposal-building hot path. It called `batcher_client.get_state_commitment_infos(block_number)` and only checked `Option::is_some()` on the result — but that RPC reads the *entire* compressed state-commitment-infos blob from `apollo_storage` (a table lookup plus a full blob read from the append-only mmap'd file), and in a remote (non-colocated) batcher deployment, serializes and transfers that whole blob over HTTP/JSON, just to immediately discard it.

This PR adds a lightweight `has_state_commitment_infos` existence check instead, plumbed through the stack:

- `apollo_storage`: new `has_state_commitment_infos` on `StateCommitmentInfosStorageReader`, doing only the table lookup (no blob read). Extracted a shared `state_commitment_infos_location` helper so `get_state_commitment_infos` and `has_state_commitment_infos` share the same lookup instead of duplicating it.
- `apollo_batcher`: new `has_state_commitment_infos` on `BatcherStorageReader` and on `Batcher`.
- `apollo_batcher_types`: new `BatcherClient::has_state_commitment_infos` method, request/response enum variants, and client wiring.
- `apollo_consensus_orchestrator`: `verify_retrospective_state_commitment_infos` now calls `has_state_commitment_infos` instead of `get_state_commitment_infos(...).is_some()`.

The existing `get_state_commitment_infos` client method/RPC is unchanged and still used where the actual payload is needed (`collect_recent_state_commitment_infos`, which builds the cende blob).

## Review

Reviewed by an independent pass focused on correctness and completeness of the RPC wiring. Findings addressed:
- Extracted the shared `state_commitment_infos_location` lookup so the two methods can't drift apart.
- Added revert-path test coverage for `has_state_commitment_infos` (a reverted height's blob stays in the append-only file, but the table entry — the only thing this check reads — is deleted).
- Documented that `false`/`None` carry the same meaning across the `has_`/`get_` pair, and that the existence check intentionally does not detect an unreadable stored blob (e.g. on-disk corruption), since detecting that would require reading the blob and defeat the purpose of the cheaper check.

## Test plan

- `cargo build -p apollo_storage -p apollo_batcher_types -p apollo_batcher -p apollo_consensus_orchestrator --features os_input` — clean.
- `SEED=0 cargo test -p apollo_storage -p apollo_batcher_types -p apollo_batcher -p apollo_consensus_orchestrator --features os_input` — all passing (118 + 193 + 318 + 9 doctests, etc.), 0 failures.
- `scripts/rust_fmt.sh` — no diff.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

---
_Generated by [Claude Code](https://claude.ai/code/session_018YEJMy3vFjgpRJ1tyWotFf)_

### DIFF
--- a/crates/apollo_batcher/src/batcher.rs
+++ b/crates/apollo_batcher/src/batcher.rs
@@ -1553,6 +1553,14 @@ impl Batcher {
         })
     }
 
+    #[cfg(feature = "os_input")]
+    pub fn has_state_commitment_infos(&self, block_number: BlockNumber) -> BatcherResult<bool> {
+        self.storage_reader.has_state_commitment_infos(block_number).map_err(|err| {
+            error!("Failed to check state commitment infos existence in storage: {err}");
+            BatcherError::InternalError
+        })
+    }
+
     fn get_commitment_results_and_write_to_storage(&mut self) -> BatcherResult<()> {
         self.commitment_manager
             .get_commitment_results_and_write_to_storage(
@@ -1758,6 +1766,11 @@ pub trait BatcherStorageReader: Send + Sync {
         height: BlockNumber,
     ) -> StorageResult<Option<CompressedStateCommitmentInfos>>;
 
+    /// Returns whether the state commitment infos for the given height are stored, without
+    /// reading the stored blob.
+    #[cfg(feature = "os_input")]
+    fn has_state_commitment_infos(&self, height: BlockNumber) -> StorageResult<bool>;
+
     fn get_parent_hash_and_partial_block_hash_components(
         &self,
         height: BlockNumber,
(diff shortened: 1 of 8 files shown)


yoavGrs:
Implemented here:
https://reviewable.io/reviews/starkware-libs/sequencer/14915

PR #14915, opened 2026-08-03
apollo_batcher,apollo_batcher_types,apollo_consensus_orchestrator: add has_state_commitment_infos

yoavGrs:

PR #14921, opened 2026-08-06
apollo_batcher,apollo_consensus_orchestrator: avoid full blob transfer for retrospective commitment infos check

gkaempfer:
## Summary

Follow-up performance optimization on top of #14914, which added `has_state_commitment_infos` (an existence-only check that avoids reading the stored blob) at the storage layer but didn't wire it through to its intended caller.

`verify_retrospective_state_commitment_infos` in `apollo_consensus_orchestrator` runs once per produced block — a per-block hot path in consensus block production — and only needed to know whether the batcher has stored the retrospective height's commitment infos. It was calling `batcher_client.get_state_commitment_infos(...).await?.is_some()`, which fetches and transfers the full compressed commitment-infos blob across the batcher's RPC boundary (a local mpsc channel or, for a remote batcher, HTTP + JSON via hyper) just to discard it and check `is_some()`.

This PR wires the new existence-only check end-to-end:
- `apollo_batcher_types`: adds `has_state_commitment_infos` to the `BatcherClient` trait plus a `HasStateCommitmentInfos` request/response RPC variant pair, mirroring the existing `get_state_commitment_infos` plumbing.
- `apollo_batcher`: adds `Batcher::has_state_commitment_infos` and the corresponding `BatcherStorageReader` trait method (backed by the storage layer's existing `has_state_commitment_infos`), and dispatches the new RPC request.
- `apollo_consensus_orchestrator`: `verify_retrospective_state_commitment_infos` now calls `has_state_commitment_infos` directly instead of fetching and discarding the full blob.
- Regenerates `dev_grafana.json` for the new request variant's infra-metrics panels, and updates/renames affected test fixtures.

## Expected impact

Removes a full compressed-witness blob transfer (serialize + deserialize, and for remote batchers a network round trip) from a call that runs on every produced block and only ever needed a boolean.

## Test plan
- [x] `cargo build` — clean
- [x] `cargo test -p apollo_storage -p apollo_batcher -p apollo_batcher_types -p apollo_consensus_orchestrator -p apollo_dashboard` — all pass (193/193 in apollo_consensus_orchestrator)
- [x] `cargo clippy` on touched crates — no warnings
- [x] Independent Opus review pass; the one blocking finding (stale generated dashboard JSON) was fixed and reverified

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01LawqajQe8bHks9fXSKWufW

---
_Generated by [Claude Code](https://claude.ai/code/session_01LawqajQe8bHks9fXSKWufW)_

### DIFF
--- a/crates/apollo_batcher/src/batcher.rs
+++ b/crates/apollo_batcher/src/batcher.rs
@@ -1535,6 +1535,13 @@ impl Batcher {
         })
     }
 
+    pub fn has_state_commitment_infos(&self, block_number: BlockNumber) -> BatcherResult<bool> {
+        self.storage_reader.has_state_commitment_infos(block_number).map_err(|err| {
+            error!("Failed to check state commitment infos existence in storage: {err}");
+            BatcherError::InternalError
+        })
+    }
+
     fn get_commitment_results_and_write_to_storage(&mut self) -> BatcherResult<()> {
         self.commitment_manager
             .get_commitment_results_and_write_to_storage(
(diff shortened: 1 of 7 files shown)


gkaempfer:
Closing: this duplicates #14915 (already open, same approach, same files) by @yoavGrs, which was itself the follow-up to #14913 after the storage-layer piece landed as #14914. No need for two PRs implementing the same wiring — please review/merge #14915 instead.

---

PR #14923, opened 2026-08-08
apollo_config_manager: validate only the dynamic config on refresh

matanl-starkware:
## Problem

The p2 alert `config_manager_update_error_increase` fired on `apollo-sepolia-alpha-4` for a **transient kube-DNS blip**, not a config problem.

`ConfigManagerRunner::update_config` ran every 20s and called `load_and_validate_config`, which validates the entire `SequencerNodeConfig`. That walks `components.*` and calls `validate_url`, which does `to_socket_addrs()` — a live blocking DNS lookup per component. One failed lookup rejected an otherwise valid config and incremented `config_manager_update_errors`.

The lookups were wasted work: `NodeDynamicConfig` holds only the `*_dynamic_config` subsets and **no `components.*` fields**, so the refresh resolved every peer Service name to validate static fields it then discarded. Static config cannot take effect without a restart and is already validated at boot.

Measured from the live ConfigMaps: 26 DNS lookups per tick per namespace, ~112k/day.

Observed on `apollo-sepolia-alpha-4` — 3 rejections in 14 days, a different component each time (which is what ruled out bad config content), each self-healing on the next 20s tick:

```
Failed to resolve url: sequencer-core-service, error: Custom { kind: Uncategorized,
  error: "failed to lookup address information: Name or service not known" }
```

## Change

- `config_utils.rs` — new `load_config()` that loads without validating. `load_and_validate_config` now delegates to it, so **boot behaviour is unchanged** and an unresolvable component url is still caught at startup.
- `config_manager_runner.rs` — `update_config` loads without static validation, then validates `NodeDynamicConfig` alone via the existing generic `config_validate`. No new validation code: `NodeDynamicConfig` already derives `Validate` with `#[validate(nested)]` on every member.
- The new failure path logs `failed to validate dynamic config`, distinct from the pre-existing `failed to update config` (load) and `Failed to update dynamic config` (dispatch), so each mode stays greppable.

## Deliberately not changed

The alert threshold stays at `> 0`. The same counter caught a real fault on `apollo-mainnet-4` (2026-07-28): a ConfigMap carrying `batcher_config.dynamic_config.results_polling_interval_millis` that the running binary's schema rejected — config/binary version skew. Raising the threshold would mask that class. Removing the DNS class is what makes `> 0` correct again.

## Test

`update_config_ignores_unresolvable_component_url` asserts `update_config` succeeds with an unresolvable `components.*.url`, while boot-time loading still rejects it. It pins the rejection reason to `"Failed to resolve url"` so it stays a test about url resolution rather than passing for an unrelated reason. Uses a `.invalid` hostname (RFC 2606), so it needs no network.

`cargo test -p apollo_config_manager -p apollo_node_config` — 36 passed. Clippy clean on both crates.

### DIFF
--- a/crates/apollo_config_manager/src/config_manager_runner.rs
+++ b/crates/apollo_config_manager/src/config_manager_runner.rs
@@ -3,13 +3,13 @@ use std::future::pending;
 use std::path::{Path, PathBuf};
 
 use apollo_config::presentation::get_config_presentation;
-use apollo_config::validators::validate_path_exists;
+use apollo_config::validators::{config_validate, validate_path_exists};
 use apollo_config::{CONFIG_FILE_ARG, CONFIG_FILE_SHORT_ARG_NAME};
 use apollo_config_manager_config::config::ConfigManagerConfig;
 use apollo_config_manager_types::communication::SharedConfigManagerClient;
 use apollo_infra::component_definitions::{default_component_start_fn, ComponentStarter};
 use apollo_infra::component_server::WrapperServer;
-use apollo_node_config::config_utils::load_and_validate_config;
+use apollo_node_config::config_utils::load_config;
 use apollo_node_config::node_config::NodeDynamicConfig;
 use async_trait::async_trait;
 use notify::{Config as NotifyConfig, EventKind, RecommendedWatcher, RecursiveMode, Watcher};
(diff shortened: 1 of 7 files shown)


matanl-starkware:
Bugbot's finding was correct, and the fix was broader than the two checks it named.

Validating only `NodeDynamicConfig` reached validators hosted on the dynamic structs, but silently dropped every validator hosted on a **parent** struct. There were three, not two:

| Validator | Host | Constrains |
|---|---|---|
| `validate_batcher_config` | `BatcherConfig` | `input_stream_content_buffer_size` >= `n_concurrent_txs` |
| `max_size_validations` | `HttpServerConfig` | `max_request_body_size` > `max_sierra_program_size` |
| `cross_member_validations` | `SequencerNodeConfig` | `proposer_idle_detection_delay_millis` < batcher deadline |

All three constrain dynamic fields that appear in `KEYS_TO_BE_REPLACED`, so they guard exactly the values operators change at runtime — this was a real regression, not a theoretical one.

**Rather than patch the three, I re-cut along the actual seam.** `to_socket_addrs` appears in exactly one place in the whole crate tree, so url resolution is the *only* network I/O in config validation. It is now out of the `Validate` derive and in an explicit `ComponentConfig::validate_urls()`:

- `validate_node_config()` — unchanged behaviour, now calls `validate_urls()` explicitly. All existing callers keep DNS validation.
- `validate_node_config_without_urls()` — everything except url resolution; used by the refresh.

So the refresh now runs the **full** validation and loses nothing, while still doing zero DNS on the 20s timer.

Two side benefits:
- The url error now names the component and the url (`Failed to resolve url X of components.batcher.`) instead of the old empty-value message.
- Existing config tests no longer depend on live DNS to pass.

Added `update_config_rejects_dynamic_value_invalid_against_static_one`, which fails on the previous commit and passes now.

matanl-starkware:
Superseded by #14924 — same change, rebuilt as a single commit on current `main` with the review feedback folded in rather than layered as fixups. Closing this to keep the history clean.

PR #14924, opened 2026-08-08
apollo_config_manager: stop resolving component urls on every config refresh

matanl-starkware:
Replaces #14923 — same change, rebuilt as a single commit on current `main`.

## Problem

`ConfigManagerRunner::update_config` re-reads and validates the node config every 20s. Validation walked `components.*` and called `validate_url`, which does `to_socket_addrs()` — a **live blocking DNS lookup per component**. A transient resolver failure therefore rejected an *unchanged, valid* config and incremented `config_manager_update_errors`, paging a p2 on a DNS blip.

Measured from the live ConfigMaps: **26 lookups per tick per namespace, ~112k/day**.

Observed on `apollo-sepolia-alpha-4` as 3 rejections in 14 days — a **different component each time**, which is what ruled out bad config content — each self-healing on the next 20s tick:

```
Failed to resolve url: sequencer-core-service, error: Custom { kind: Uncategorized,
  error: "failed to lookup address information: Name or service not known" }
```

## Change

Url resolution is the only network I/O in config validation (`to_socket_addrs` appears in exactly one place), so it moves out of the `Validate` derive into an explicit `ComponentConfig::validate_urls()`:

- `validate_node_config()` — unchanged behaviour; now calls `validate_urls()` explicitly, so every existing caller keeps DNS validation at boot.
- `validate_node_config_without_urls()` — everything else; used by the refresh.

Component urls are static, so boot-time resolution is sufficient. The refresh runs the **full** validation minus that one step, so no validator is lost — importantly including validators hosted on *parent* structs that constrain hot-updatable values (`input_stream_content_buffer_size` vs `n_concurrent_txs`, `max_request_body_size` vs `max_sierra_program_size`, `proposer_idle_detection_delay_millis` vs the batcher deadline).

The principle encoded here: **network-dependent validation runs on a change, never on a poll.** For a static url the change edge is boot. If a url ever becomes dynamic it must be resolved where the config manager detects the change, not on the timer — noted in the doc comment on `validate_node_config_without_urls`.

## Guards against regression

- `validate_urls` destructures `ComponentConfig` **exhaustively** (no `..`), so adding a component fails to compile until it is classified reactive or active.
- `component_config_validation_does_not_resolve_urls` asserts the derive accepts unresolvable urls while `validate_urls` rejects them — reintroducing network I/O into validation fails the suite. Verified it fails when the resolve call is put back.
- Refresh-path tests for both a parent-struct cross-check and a cross-config one.

## Also

[...]

PR #14926, opened 2026-08-08
apollo_config_manager: retry a dynamic config dispatch the component rejected

### LINKED ISSUE #14925: ConfigManagerRunner: a failed dynamic config dispatch is never retried and never alerts
Found while reviewing #14924. **Pre-existing on `main`, not introduced by that PR** — filing separately.

## The bug

`ConfigManagerRunner::update_config` (`crates/apollo_config_manager/src/config_manager_runner.rs`) advances `latest_node_dynamic_config` **before** awaiting the dispatch to the config manager client:

```rust
self.latest_node_dynamic_config = node_dynamic_config.clone();   // <-- advanced first
match self.config_manager_client.set_node_dynamic_config(node_dynamic_config.clone()).await {
    Ok(()) => { info!("Successfully updated dynamic config"); Ok(node_dynamic_config) }
    Err(e) => Err(format!("Failed to update dynamic config: {:?}", e).into()),   // <-- no revert, no metric
}
```

If `set_node_dynamic_config` fails, two things go wrong together:

1. **No retry, ever.** `latest_node_dynamic_config` already holds the new value, so every subsequent tick takes the `latest == new` early-return and does nothing. The node runs on the **old** config indefinitely while the runner believes the new one was applied.
2. **No alert.** Unlike the load and validate paths, this error path does not increment `CONFIG_MANAGER_UPDATE_ERRORS`, so `config_manager_update_error_increase` never fires. The failure is silent.

## Why it matters

This is the one path where a config update can be silently dropped. `set_node_dynamic_config` can legitimately fail — the receiving component validates the dynamic config (`ConfigManager` runs `validate()` on it), and it is a channel-based call that can error under load. An operator edits a ConfigMap, sees no error, and the value never takes effect.

## Impact

Not currently known to have fired in production — this is from code reading, not an incident. Severity is "silent staleness with no signal" rather than an outage.

## Suggested fix

Advance `latest_node_dynamic_config` only after a successful dispatch, and increment `CONFIG_MANAGER_UPDATE_ERRORS` on the error path so it is covered by the existing alert:

```rust
match self.config_manager_client.set_node_dynamic_config(node_dynamic_config.clone()).await {
    Ok(()) => {
        self.latest_node_dynamic_config = node_dynamic_config.clone();
        info!("Successfully updated dynamic config");
        Ok(node_dynamic_config)
    }
    Err(e) => {
        CONFIG_MANAGER_UPDATE_ERRORS.increment(1);
        Err(format!("Failed to update dynamic config: {:?}", e).into())
    }
}
```

Leaving `latest_` untouched makes the next 20s tick retry naturally, and a persistent failure then alerts instead of going quiet.

Worth a regression test: a mock client that fails `set_node_dynamic_config` should leave the runner retrying on the following `update_config` call, rather than reporting no change.

matanl-starkware:
Fixes #14925. Independent of #14924 — branched from `main`, touches only the dispatch branch of `update_config`.

## Problem

`update_config` advanced `latest_node_dynamic_config` **before** awaiting `set_node_dynamic_config`:

```rust
self.latest_node_dynamic_config = node_dynamic_config.clone();   // <-- advanced first
match self.config_manager_client.set_node_dynamic_config(node_dynamic_config.clone()).await {
    Ok(()) => { ... }
    Err(e) => Err(format!("Failed to update dynamic config: {:?}", e).into()),   // no revert, no metric
}
```

When the dispatch failed, two things went wrong together:

1. **Never retried.** `latest_` already held the new value, so every later tick took the `latest == new` early return and did nothing. The node kept running the **old** config indefinitely, while the runner behaved as though the new one had been applied.
2. **Never alerted.** Unlike the load and validate paths, this one did not increment `CONFIG_MANAGER_UPDATE_ERRORS`, so `config_manager_update_error_increase` could not fire. The dropped update was completely silent.

This is the one path where a config change can be lost with no signal. `set_node_dynamic_config` can legitimately fail — the receiving component validates the dynamic config, and it is a channel-based call that can error under load.

## Change

Record the config only once the component has accepted it, and count the failure:

- `latest_node_dynamic_config` is assigned inside the `Ok` arm. Leaving it untouched on failure means the next 20s tick sees a difference and retries by itself.
- The `Err` arm increments `CONFIG_MANAGER_UPDATE_ERRORS` and logs, so a *persistent* failure surfaces through the existing alert instead of going quiet. A single transient rejection self-heals on the next tick.

## Tests

- `failed_dispatch_is_retried_on_the_next_update` — a client that always rejects must be called **twice** across two `update_config` calls. Verified this fails on the old ordering (mockall reports the second dispatch never happens).
- `accepted_dispatch_is_not_repeated_while_the_config_is_unchanged` — the counterpart, pinning that a successful dispatch is still recorded, so an unchanged file is not re-dispatched every tick.

`cargo test -p apollo_config_manager` — 17 passed. `scripts/rust_fmt.sh --check` and clippy clean.

## Scope

Found while reviewing #14924 but present on `main` well before it; kept separate so an unrelated pre-existing bug does not ride along on that PR.

### DIFF
--- a/crates/apollo_config_manager/src/config_manager_runner.rs
+++ b/crates/apollo_config_manager/src/config_manager_runner.rs
@@ -143,18 +143,23 @@ impl ConfigManagerRunner {
         } else {
             // Log the diff between the latest and the new node dynamic config.
             self.log_config_diff(&self.latest_node_dynamic_config, &node_dynamic_config);
-            // Update the latest node dynamic config.
-            self.latest_node_dynamic_config = node_dynamic_config.clone();
             match self
                 .config_manager_client
                 .set_node_dynamic_config(node_dynamic_config.clone())
                 .await
             {
                 Ok(()) => {
+                    // Recorded only once the component has accepted it, so that a failed dispatch
+                    // is retried on the next update instead of being seen as no change.
+                    self.latest_node_dynamic_config = node_dynamic_config.clone();
                     info!("Successfully updated dynamic config");
                     Ok(node_dynamic_config)
                 }
-                Err(e) => Err(format!("Failed to update dynamic config: {:?}", e).into()),
+                Err(e) => {
+                    CONFIG_MANAGER_UPDATE_ERRORS.increment(1);
+                    error!("ConfigManagerRunner: failed to update dynamic config: {e}");
+                    Err(format!("Failed to update dynamic config: {:?}", e).into())
+                }
             }
         }
     }
--- a/crates/apollo_config_manager/src/config_manager_runner_tests.rs
+++ b/crates/apollo_config_manager/src/config_manager_runner_tests.rs
@@ -5,9 +5,11 @@ use std::time::Duration;
 use apollo_config::CONFIG_FILE_ARG;
 use apollo_config_manager_config::config::ConfigManagerConfig;
 use apollo_config_manager_types::communication::{
+    ConfigManagerClientError,
     MockConfigManagerClient,
     SharedConfigManagerClient,
 };
+use apollo_config_manager_types::errors::ConfigManagerError;
 use apollo_consensus_config::config::ConsensusDynamicConfig;
 use apollo_node_config::component_execution_config::ReactiveComponentExecutionConfig;
 use apollo_node_config::config_utils::{load_and_validate_config, DeploymentBaseAppConfig};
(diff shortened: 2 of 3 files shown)


matanl-starkware:
Reviewed with an independent adversarial pass. Verdict: no high-severity findings; the retry path was mutation-tested (reverting only `config_manager_runner.rs` to `main` makes `failed_dispatch_is_retried_on_the_next_update` fail at its own `expect_err`, not via a mock-internal panic).

Applied two of the findings in this push:
- Metric description was stale — now `(load/validate/dispatch)`, since the counter covers dispatch failures too.
- The new `error!` used `{e:?}` while the load-error path in the same function uses `{e}`. Aligned to Display, so both error paths grep the same way during an incident.

Two notes for reviewers, neither applied:

**Log volume on a permanently-rejected config.** Because `latest_node_dynamic_config` no longer advances on failure, a config the component keeps rejecting is retried every 20s, and `log_config_diff` re-emits one `info!` per changed key each time — a 5-key bad push left overnight is ~10k duplicate lines. There is partial precedent (the load-error path already re-logs per tick), but the multi-line diff re-log is new. Options if we care: log the diff only when it changes, or only on the first failure. Happy to add it here or follow up.

**`Err`-after-apply divergence (unreachable today).** If the client ever returns `Err` after the component already applied the config, the runner keeps the old value; a subsequent revert to that old value would then look like "no change" and never dispatch, leaving the component ahead. Unreachable in current deployments — `config_manager` is always `local_with_remote_disabled`, and `LocalComponentClient::send` panics rather than returning `Err`, so the only reachable `Err` is the component's own `InvalidConfig`, where it definitively did not apply. Noting it in case remote mode is ever enabled.

On the design trade-off itself: retrying forever with a firing p2 is clearly better than the old silent drop. The component rejects before writing, so there is no state thrash, and the alert now stays up exactly as long as the problem does.

itamar-starkware (approved):
:lgtm:

> @itamar-starkware reviewed 3 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14926#-:-OzhaaCv1sh3Tf_PY974:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

matanl-starkware (commented):
> @matanl-starkware reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14926)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

gkaempfer:

---

PR #14931, opened 2026-08-09
apollo_committer: collapse the four per-block commit logs into one

matanl-starkware:
## What

Committing one block emitted four log lines that largely restate each other:

| line | level | content |
|---|---|---|
| `"Received request to commit block number {height} with state diff {commitment:?}"` | info | height + commitment |
| `"Committing block number {height} with state diff {commitment:?}"` | debug | height + commitment (identical) |
| `"For block number {height}, writing filled forest to storage with metadata: {metadata:?}, delete N nodes"` | info | height + commitment + global root, serialized |
| `"Committed block number {height} with state diff {commitment:?}"` | debug | height + commitment (identical) |

This keeps **one** summary line in `commit_block`, promoted to `info!` and extended with the resulting global root, and slims the forest-write line down to the node count.

## Why

Part of a round of log-cost reduction. Measured over 24h of Mainnet logs via Log Analytics, these four lines are **40.9% of all `sequencer-committer` log bytes** between them (11.44% / 9.90% / 9.79% / 9.75%), at one occurrence each per block per node.

## What was kept

- The failure branch (`error!`) is untouched.
- The global root is now reported where it wasn't before, so the summary line is strictly more informative than any of the four it replaces.
- `metadata` was dropped from the forest-write line because `commit_tip_metadata_bundle` builds it out of exactly the height, global root and state diff commitment — all three now in the summary line. The deleted-node count is not available elsewhere, so that line survives as `debug!`.

## Test plan

- `cargo build -p apollo_committer`
- `SEED=0 cargo test -p apollo_committer` — 17 passed, 0 failed
- `scripts/rust_fmt.sh`

### DIFF
--- a/crates/apollo_committer/src/committer.rs
+++ b/crates/apollo_committer/src/committer.rs
@@ -181,9 +181,14 @@ where
         let result = self
             .commit_block_inner(CommitBlockRequest { state_diff, state_diff_commitment, height })
             .await;
+        // The single per-block line for the commit flow. The steps in between used to log the
+        // same height and state diff commitment three more times over.
         match &result {
-            Ok(_) => {
-                debug!("Committed block number {height} with state diff {state_diff_commitment:?}");
+            Ok(CommitBlockResponse { global_root }) => {
+                info!(
+                    "Committed block number {height} with state diff {state_diff_commitment:?}, \
+                     global root {global_root:?}"
+                );
             }
             Err(err) => {
                 error!("Failed to commit block number {height}: {err:?}");
@@ -198,29 +203,23 @@ where
         &mut self,
         CommitBlockRequest { state_diff, state_diff_commitment, height }: CommitBlockRequest,
     ) -> CommitterResult<CommitBlockResponse> {
-        info!(
-            "Received request to commit block number {height} with state diff \
-             {state_diff_commitment:?}"
-        );
-
         match self.commit_or_load(&state_diff, state_diff_commitment, height).await? {
             CommitBlockHeightPlan::Historical { global_root } => {
                 Ok(CommitBlockResponse { global_root })
             }
             CommitBlockHeightPlan::CommitTip { state_diff_commitment } => {
                 // Happy flow. Commits the state diff and returns the computed global root.
-                debug!(
-                    "Committing block number {height} with state diff {state_diff_commitment:?}"
-                );
                 let mut block_measurements = SingleBlockMeasurements::default();
                 block_measurements.start_measurement(Action::EndToEnd);
                 let CommitStateDiffOutput { filled_forest, global_root, deleted_nodes } =
                     self.commit_state_diff(state_diff, &mut block_measurements).await?;
                 let (metadata, next_offset) =
                     commit_tip_metadata_bundle(height, global_root, state_diff_commitment);
-                info!(
-                    "For block number {height}, writing filled forest to storage with metadata: \
-                     {metadata:?}, delete {} nodes",
+                // `metadata` is just the height, global root and state diff commitment, all of
+                // which the commit_block summary line already reports, so only the node count
+                // is worth logging here.
+                debug!(
+                    "Writing filled forest for block number {height} to storage, deleting {} nodes",
                     deleted_nodes.len()
                 );
                 block_measurements.start_measurement(Action::Write);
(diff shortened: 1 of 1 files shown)


yoavGrs (changes requested):
> @yoavGrs made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14931)* status: 0 of 1 files reviewed, 3 unresolved discussions (waiting on matanl-starkware).

___
*[`crates/apollo_committer/src/committer.rs` line 202 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14931#-OzeSauCEpQhnphkvimL:-OzeSauCEpQhnphkvimM:bwp8yrc) ([raw file](https://github.com/starkware-libs/sequencer/blob/baf1020146645e325dbbf8313707ef8f5f3c9a20/crates/apollo_committer/src/committer.rs#L202)):*
> ```XML
>     ) -> CommitterResult<CommitBlockResponse> {
>         info!(
>             "Received request to commit block number {height} with state diff \
> ```

Note that you are removing the marker for the start of the commit.
___
*[`crates/apollo_committer/src/committer.rs` line 185 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14931#-OzeTEYY1eaBxjCqqhNJ:-OzeTEYY1eaBxjCqqhNK:b-vskn0y) ([raw file](https://github.com/starkware-libs/sequencer/blob/2db40c7a7e344fedadca52d294126dcdd8016714/crates/apollo_committer/src/committer.rs#L185)):*
> ```XML
>             .await;
>         // The single per-block line for the commit flow. The steps in between used to log the
>         // same height and state diff commitment three more times over.
> ```

The three more times will no longer exist.



_Suggestion:_
```XML
        // The single per-block line for the commit flow.
```
___
*[`crates/apollo_committer/src/committer.rs` line 220 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14931#-OzeNdzT4ebArm5-I7qY:-OzeNdzT4ebArm5-I7qZ:b-ucl638) ([raw file](https://github.com/starkware-libs/sequencer/blob/2db40c7a7e344fedadca52d294126dcdd8016714/crates/apollo_committer/src/committer.rs#L220)):*
> ```XML
>                 // `metadata` is just the height, global root and state diff commitment, all of
>                 // which the commit_block summary line already reports, so only the node count
>                 // is worth logging here.
> ```

No need to explain why you don't log some fields.



_Code quote:_
```XML
                // `metadata` is just the height, global root and state diff commitment, all of
                // which the commit_block summary line already reports, so only the node count
                // is worth logging here.
```

matanl-starkware:
Thanks — all three applied in `acc1928`.

**1. Start-of-commit marker** — you're right, and it's the one that mattered. If a commit hangs, the completion line never fires and there'd be nothing in the committer log to show block N was even attempted. `COMMITTER_BLOCK_COMMIT_LATENCY` and the duration warn only cover commits that *finish*, so removing the marker was a real regression. Restored.

I did trim it to just the height:

```rust
info!("Received request to commit block number {height}.");
```

The `state_diff_commitment` is still reported by `commit_block` once the outcome is known, so carrying it on both lines was the duplication I was actually after — the marker itself only needs to identify the block. Say the word if you'd rather have the original line back verbatim.

**2. Stale comment** — fixed. Since the start line is back, "the single per-block line" would have been wrong too, so it now reads `// Outcome of the commit; commit_block_inner logs its start.`

**3. Metadata rationale** — removed.

**Net effect** is now 4 lines → 3, with the two duplicates gone (`"Committing block number ..."`, which restated the start line verbatim, and the `{metadata:?}` dump, which serialises to the height, global root and commitment that the completion line already carries). The completion line also gained the global root, which wasn't logged anywhere before.

Saving drops from ~$46/month to ~$30/month — the right trade for keeping the marker.

`SEED=0 cargo test -p apollo_committer` — 17 passed. clippy and fmt clean.

yoavGrs (approved):
> @yoavGrs reviewed 1 file and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14931)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

gkaempfer:

---

PR #14932, opened 2026-08-09
apollo_config,apollo_config_manager: stop re-logging the config load on every poll

matanl-starkware:
## What

`ConfigManagerRunner` reloads the config on a timer. Each tick re-emitted three `info!` lines in every pod:

- `config_manager_runner.rs` — `"ConfigManagerRunner: periodic check triggered, updating config"`
- `loading.rs` — `"Ignoring default values by overriding with an empty map."`
- `loading.rs` — `"Loading custom config file: ..."` (once per file)

Now:
- the two `loading.rs` lines report at `info!` for the **first** load of the process and `trace!` afterwards, via a `CONFIG_LOADED_BEFORE` atomic;
- the tick heartbeat drops to `trace!`.

## Why

Part of a round of log-cost reduction. Measured over 24h of Mainnet logs via Log Analytics, these three lines are **~9.8% of `sequencer-gateway`'s log bytes** and ~3% of `sequencer-mempool`'s — roughly $60/month fleet-wide. They fire in every pod, forever, for a config that almost never changes.

`update_config` already compares old vs new and no-ops when nothing changed, but `load_and_validate_config` underneath it logged unconditionally.

## Why not just `debug!`

Production runs `RUST_LOG=debug,cairo_vm=warn`, so demoting to `debug!` would save nothing. `trace!` is below the deployed threshold.

## What's preserved

- The **startup** config load still logs in full at `info!` — that is the occurrence anyone actually reads.
- Real config changes are still reported by `log_config_diff` (`"ConfigManagerRunner: {key} changed from {old} to {new}"`) and `"Successfully updated dynamic config"`, both untouched.
- The filesystem-watcher path (`"file change detected"`) is untouched — it only fires on an actual change.
- The flag is only set once a load runs to completion, so a first load that fails early still logs at `info!` on the next attempt.

## Test plan

- `cargo build -p apollo_config -p apollo_config_manager`
- `SEED=0 cargo test -p apollo_config` — 34 passed, 0 failed
- `SEED=0 cargo test -p apollo_config_manager` — 15 passed, 0 failed
- `scripts/rust_fmt.sh`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1254,6 +1254,7 @@ dependencies = [
  "apollo_consensus_orchestrator_config",
  "apollo_http_server_config",
  "apollo_infra",
+ "apollo_infra_utils",
  "apollo_mempool_config",
  "apollo_metrics",
  "apollo_node_config",
--- a/crates/apollo_config/src/loading.rs
+++ b/crates/apollo_config/src/loading.rs
@@ -7,6 +7,7 @@ use std::collections::{BTreeMap, HashSet};
 use std::fs::File;
 use std::ops::IndexMut;
 use std::path::PathBuf;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use clap::parser::Values;
 use clap::Command;
@@ -14,7 +15,7 @@ use command::{get_command_matches, update_config_map_by_command_args};
 use itertools::any;
 use serde::Deserialize;
 use serde_json::{json, Map, Value};
-use tracing::{error, info, instrument};
+use tracing::{error, info, instrument, trace};
 
 use crate::validators::validate_path_exists;
 use crate::{
@@ -47,6 +48,16 @@ pub fn load<T: for<'a> Deserialize<'a>>(
     Ok(serde_json::from_value(nested_map)?)
 }
 
+static CONFIG_LOADED_BEFORE: AtomicBool = AtomicBool::new(false);
+
+// Defined above its call sites because a `macro_rules!` is only in scope after its definition, and
+// a fn would report this location as the event's file and line instead of the caller's.
+macro_rules! log_config_load {
+    ($($arg:tt)*) => {
+        if CONFIG_LOADED_BEFORE.load(Ordering::Relaxed) { trace!($($arg)*) } else { info!($($arg)*) }
+    };
+}
+
 /// Deserializes a json config file, updates the values by the given arguments for the command, and
 /// set values for the pointers.
 pub fn load_and_process_config<T: for<'a> Deserialize<'a>>(
@@ -65,7 +76,7 @@ pub fn load_and_process_config<T: for<'a> Deserialize<'a>>(
     // Retaining values from the default config map for backward compatibility.
     let (mut values_map, types_map) = split_values_and_types(config_map);
     if ignore_default_values {
-        info!("Ignoring default values by overriding with an empty map.");
+        log_config_load!("Ignoring default values by overriding with an empty map.");
         values_map = BTreeMap::new();
     }
     // If the config_file arg is given, updates the values map according to this files.
@@ -133,6 +144,9 @@ pub fn load_and_process_config<T: for<'a> Deserialize<'a>>(
             );
         }
     }
+    // Set only once a load has run to completion, so a first load that fails early still reports
+    // in full on the next attempt.
+    CONFIG_LOADED_BEFORE.store(true, Ordering::Relaxed);
     // Return the loaded config result.
     load_result
 }
@@ -187,7 +201,7 @@ pub(crate) fn update_config_map_by_custom_configs(
     custom_config_paths: Values<PathBuf>,
 ) -> Result<(), ConfigError> {
     for config_path in custom_config_paths {
-        info!("Loading custom config file: {:?}", config_path);
+        log_config_load!("Loading custom config file: {config_path:?}");
         validate_path_exists(&config_path)?;
         let file = std::fs::File::open(config_path)?;
         let custom_config: Map<String, Value> = serde_json::from_reader(file)?;
--- a/crates/apollo_config_manager/Cargo.toml
+++ b/crates/apollo_config_manager/Cargo.toml
@@ -22,6 +22,7 @@ apollo_consensus_config.workspace = true
 apollo_consensus_orchestrator_config.workspace = true
 apollo_http_server_config.workspace = true
 apollo_infra.workspace = true
+apollo_infra_utils.workspace = true
 apollo_mempool_config.workspace = true
 apollo_metrics.workspace = true
 apollo_node_config.workspace = true
--- a/crates/apollo_config_manager/src/config_manager_runner.rs
+++ b/crates/apollo_config_manager/src/config_manager_runner.rs
@@ -9,6 +9,7 @@ use apollo_config_manager_config::config::ConfigManagerConfig;
 use apollo_config_manager_types::communication::SharedConfigManagerClient;
 use apollo_infra::component_definitions::{default_component_start_fn, ComponentStarter};
 use apollo_infra::component_server::WrapperServer;
+use apollo_infra_utils::info_every_n;
 use apollo_node_config::config_utils::load_config;
 use apollo_node_config::node_config::NodeDynamicConfig;
 use async_trait::async_trait;
@@ -21,6 +22,7 @@ use tracing::{error, info};
 use crate::metrics::{register_metrics, CONFIG_MANAGER_UPDATE_ERRORS};
 
 const FS_EVENT_CHANNEL_CAPACITY: usize = 16;
+const HEARTBEAT_LOG_EVERY_N_TICKS: usize = 10;
 
 #[cfg(test)]
 #[path = "config_manager_runner_tests.rs"]
@@ -111,9 +113,13 @@ impl ConfigManagerRunner {
                         }
                     }
                 }
-                // Periodic tick
+                // Periodic tick. Sampled rather than silenced, so a wedged poll loop is still
+                // visible as the heartbeat stopping.
                 _ = update_interval.tick() => {
-                    info!("ConfigManagerRunner: periodic check triggered, updating config");
+                    info_every_n!(
+                        HEARTBEAT_LOG_EVERY_N_TICKS,
+                        "ConfigManagerRunner: periodic check triggered, updating config"
+                    );
                     let _ = self.update_config().await;
                 }
             }


asaf-sw (changes requested):
> @asaf-sw reviewed 2 files and all commit messages, and made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14932)* status: all files reviewed, 2 unresolved discussions (waiting on matanl-starkware).

___
*[`crates/apollo_config/src/loading.rs` line 52 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKUbL1qph4IidHbfP:-OzuKUbO4K8DwsD7I_UX:b-utl3he) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config/src/loading.rs#L52)):*
> ```XML
> }
> 
> /// Whether a config has already been loaded in this process. See [`log_config_load`].
> ```

nit: both the static and the helper are module private, so these should be `//` rather than `///`. The style guide keeps `///` for the API a user sees and `//` for explaining the implementation, and the `#![warn(missing_docs)]` in `lib.rs` only covers public items, so `///` buys nothing here.

Same reason, the `[log_config_load]` intra-doc link never renders as a link: rustdoc does not emit pages for private items by default. Plain text reads identically.
___
*[`crates/apollo_config/src/loading.rs` line 53 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKX6_2MmyUYu1iIoy:-OzuKX6cCxSPJZO4QZ9B:byhesae) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config/src/loading.rs#L53)):*
> ```XML
> 
> /// Whether a config has already been loaded in this process. See [`log_config_load`].
> static CONFIG_LOADED_BEFORE: AtomicBool = AtomicBool::new(false);
> ```

nit: the flag is process global with no way to reset it, and `config_test.rs` calls `load_and_process_config` six times across tests that share one process and run in parallel. Whichever test gets there first logs at `info!` and the rest log at `trace!`.

Nothing breaks today, since `apollo_config` has no `tracing-test` dependency and no test asserts on these lines. Flagging it only because a log assertion added later would be flaky by construction, and a `#[cfg(test)]` reset helper is the easy escape hatch if that ever comes up.
___
*[`crates/apollo_config/src/loading.rs` line 60 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKPHEBVFQe52JacwP:-OzuKPHO9lDE--RAZvpJ:b-9divtm) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config/src/loading.rs#L60)):*
> ```XML
> /// forever in every pod for a config that almost never changes. The startup load is the one worth
> /// reading, so subsequent loads are demoted to `trace`.
> fn log_config_load(message: Arguments<'_>) {
> ```

`tracing` captures `file` and `line` at the macro call site, so moving `info!`/`trace!` into a function makes every event report this helper's location instead of the real one. Both messages now come out as `loading.rs:62` or `:64` instead of `:86` and `:211`.

That matters here because production logs drop the target and keep file and line as the only source location (`apollo_infra/src/trace_util.rs:88-91`, `.with_target(false)` plus `.with_file(true)` and `.with_line_number(true)`). So the two messages become indistinguishable by location, and any Log Analytics query or saved filter keyed on line number silently stops matching. That is the same line-level accounting used to measure the savings in the first place.

A local `macro_rules!` keeps the metadata and is less code than the current version, since it drops `use std::fmt::Arguments` and both `format_args!` wrappers:

```rust
macro_rules! log_config_load {
    ($($arg:tt)*) => {
        if CONFIG_LOADED_BEFORE.load(Ordering::Relaxed) { trace!($($arg)*) } else { info!($($arg)*) }
    };
}
```

Call sites become `log_config_load!("Loading custom config file: {config_path:?}");`. No `#[macro_export]` needed, since both uses are in the same module below the definition.
___
*[`crates/apollo_config/src/loading.rs` line 66 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKVsf1c6D_aJe1srY:-OzuKVsk80wlWrH4Dqj2:b-gi5cb3) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config/src/loading.rs#L66)):*
> ```XML
>         info!("{message}");
>     }
> }
> ```

nit: the static and this helper landed between `load` and `load_and_process_config`, so a log level helper is now the third thing a reader meets in the file. Convention here is public entry points first and private helpers after them, so both would sit better below `load_and_process_config`.
___
*[`crates/apollo_config_manager/src/config_manager_runner.rs` line 117 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKQuG7IPUaN_KlV38:-OzuKQuQEuP7fipVY8wK:bgovhbi) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config_manager/src/config_manager_runner.rs#L117)):*
> ```XML
>                 // an actual config change is still reported by log_config_diff below.
>                 _ = update_interval.tick() => {
>                     trace!("ConfigManagerRunner: periodic check triggered, updating config");
> ```

This heartbeat was the only external signal that the reload loop is still alive, so demoting it is worth a deliberate decision.

`update_config()` is awaited inline in this arm, and it calls `set_node_dynamic_config(...).await`, which can hang. If that happens, `run_watcher_loop` stops forever. After this change that is invisible in both logs and metrics: the tick is below the deployed `RUST_LOG=debug`, and `metrics.rs` defines only `CONFIG_MANAGER_UPDATE_ERRORS`, which counts failures. There is no counter for checks performed and no last-check timestamp. So a dead poll loop surfaces nowhere until someone pushes a dynamic config change and nothing happens.

Two cheap options: add a `CONFIG_MANAGER_UPDATE_CHECKS` counter incremented on each tick, or use `apollo_infra_utils::tracing::info_every_n_ms!` here instead of going all the way down to `trace!`. Either keeps the savings.

asaf-sw:
*[`crates/apollo_config/src/loading.rs` line 52 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKUbL1qph4IidHbfP:-Ozv8wUE4TjKSFCh3jjN:b2nhzyn) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config/src/loading.rs#L52)):*
<details><summary><i>Previously, asaf-sw wrote…</i></summary><blockquote>

nit: both the static and the helper are module private, so these should be `//` rather than `///`. The style guide keeps `///` for the API a user sees and `//` for explaining the implementation, and the `#![warn(missing_docs)]` in `lib.rs` only covers public items, so `///` buys nothing here.

Same reason, the `[log_config_load]` intra-doc link never renders as a link: rustdoc does not emit pages for private items by default. Plain text reads identically.
</blockquote></details>

bug

asaf-sw:
*[`crates/apollo_config/src/loading.rs` line 66 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKVsf1c6D_aJe1srY:-Ozv8yrO9PX9XU4b3XMW:b2nhzyn) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config/src/loading.rs#L66)):*
<details><summary><i>Previously, asaf-sw wrote…</i></summary><blockquote>

nit: the static and this helper landed between `load` and `load_and_process_config`, so a log level helper is now the third thing a reader meets in the file. Convention here is public entry points first and private helpers after them, so both would sit better below `load_and_process_config`.
</blockquote></details>

bug

asaf-sw:
*[`crates/apollo_config/src/loading.rs` line 53 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKX6_2MmyUYu1iIoy:-Ozv9-ZE8ep_rU9wU9Nf:b-5wyxkh) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config/src/loading.rs#L53)):*
<details><summary><i>Previously, asaf-sw wrote…</i></summary><blockquote>

nit: the flag is process global with no way to reset it, and `config_test.rs` calls `load_and_process_config` six times across tests that share one process and run in parallel. Whichever test gets there first logs at `info!` and the rest log at `trace!`.

Nothing breaks today, since `apollo_config` has no `tracing-test` dependency and no test asserts on these lines. Flagging it only because a log assertion added later would be flaky by construction, and a `#[cfg(test)]` reset helper is the easy escape hatch if that ever comes up.
</blockquote></details>

bug  maybe (not sure this is a real probable issue)

matanl-starkware (commented):
> @matanl-starkware made 5 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14932)* status: 0 of 3 files reviewed, 5 unresolved discussions (waiting on asaf-sw).

___
*[`crates/apollo_config/src/loading.rs` line 52 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKUbL1qph4IidHbfP:-P-ELFEY81hFBN4EziUH:b-i32s7s) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config/src/loading.rs#L52)):*
<details><summary><i>Previously, asaf-sw wrote…</i></summary><blockquote>

bug
</blockquote></details>

Removed. Pretty self-explanatory.
Thanks
___
*[`crates/apollo_config/src/loading.rs` line 53 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKX6_2MmyUYu1iIoy:-P-ELXvlEOjv9ixoPa5O:b-dnja5u) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config/src/loading.rs#L53)):*
<details><summary><i>Previously, asaf-sw wrote…</i></summary><blockquote>

bug  maybe (not sure this is a real probable issue)
</blockquote></details>

IMO it's not a problem at the moment.
___
*[`crates/apollo_config/src/loading.rs` line 60 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKPHEBVFQe52JacwP:-P-ELdZ9EL3MtBgY8xGZ:biifwie) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config/src/loading.rs#L60)):*
<details><summary><i>Previously, asaf-sw wrote…</i></summary><blockquote>

`tracing` captures `file` and `line` at the macro call site, so moving `info!`/`trace!` into a function makes every event report this helper's location instead of the real one. Both messages now come out as `loading.rs:62` or `:64` instead of `:86` and `:211`.

That matters here because production logs drop the target and keep file and line as the only source location (`apollo_infra/src/trace_util.rs:88-91`, `.with_target(false)` plus `.with_file(true)` and `.with_line_number(true)`). So the two messages become indistinguishable by location, and any Log Analytics query or saved filter keyed on line number silently stops matching. That is the same line-level accounting used to measure the savings in the first place.

A local `macro_rules!` keeps the metadata and is less code than the current version, since it drops `use std::fmt::Arguments` and both `format_args!` wrappers:

```rust
macro_rules! log_config_load {
    ($($arg:tt)*) => {
        if CONFIG_LOADED_BEFORE.load(Ordering::Relaxed) { trace!($($arg)*) } else { info!($($arg)*) }
    };
}
```

Call sites become `log_config_load!("Loading custom config file: {config_path:?}");`. No `#[macro_export]` needed, since both uses are in the same module below the definition.
</blockquote></details>

Good catch
Thanks
___
*[`crates/apollo_config/src/loading.rs` line 66 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKVsf1c6D_aJe1srY:-P-ELhWh76h2OjJPf-gI:b-896fix) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config/src/loading.rs#L66)):*
<details><summary><i>Previously, asaf-sw wrote…</i></summary><blockquote>

bug
</blockquote></details>

Done.
___
*[`crates/apollo_config_manager/src/config_manager_runner.rs` line 117 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14932#-OzuKQuG7IPUaN_KlV38:-P-ELor0-DW9e3srOz0V:b-4n2n04) ([raw file](https://github.com/starkware-libs/sequencer/blob/7f0051f3e08af48af98b4f10ae388a361df20987/crates/apollo_config_manager/src/config_manager_runner.rs#L117)):*
<details><summary><i>Previously, asaf-sw wrote…</i></summary><blockquote>

This heartbeat was the only external signal that the reload loop is still alive, so demoting it is worth a deliberate decision.

`update_config()` is awaited inline in this arm, and it calls `set_node_dynamic_config(...).await`, which can hang. If that happens, `run_watcher_loop` stops forever. After this change that is invisible in both logs and metrics: the tick is below the deployed `RUST_LOG=debug`, and `metrics.rs` defines only `CONFIG_MANAGER_UPDATE_ERRORS`, which counts failures. There is no counter for checks performed and no last-check timestamp. So a dead poll loop surfaces nowhere until someone pushes a dynamic config change and nothing happens.

Two cheap options: add a `CONFIG_MANAGER_UPDATE_CHECKS` counter incremented on each tick, or use `apollo_infra_utils::tracing::info_every_n_ms!` here instead of going all the way down to `trace!`. Either keeps the savings.
</blockquote></details>

Used info_every_n!(10)

asaf-sw (commented):
> @asaf-sw reviewed 4 files and all commit messages, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14932)* status: all files reviewed, 1 unresolved discussion (waiting on matanl-starkware).

asaf-sw (approved):
> @asaf-sw resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14932)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

matanl-starkware (commented):
> @matanl-starkware reviewed 4 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14932)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

gkaempfer:

---

PR #14624, opened 2026-06-24
blockifier: avoid full StateMaps clone in get_os_initial_reads

gkaempfer:
## Summary

Follow-up performance optimization to #14621.

`get_os_initial_reads` called `get_initial_reads()` twice — the first call cloned the full `StateMaps` (all five `HashMap`s including the potentially-large `storage` map) only to iterate over contract addresses and declared-class keys. That snapshot is immediately thrown away after the force-read loops.

This PR replaces the first full clone with a targeted borrow that extracts only the two small collections actually needed for iteration:

- `HashSet<ContractAddress>` via `get_contract_addresses()` — already computed in both old and new paths
- `Vec<ClassHash>` of declared-contract keys — a handful of entries

**Impact:** For a busy block the `storage` map in `initial_reads` can hold tens of thousands of entries (one per unique storage read across all transactions). Skipping its clone avoids an O(storage-reads-per-block) allocation on the critical post-execution path, reducing memory pressure and improving cache locality.

No semantic change — the second `get_initial_reads()` clone (the actual return value) is unchanged.

## Test plan

- [x] `cargo test -p blockifier --features "reexecution,testing" cached_state` — all 43 tests pass
- [ ] CI green

---

_Generated by_ [_Claude Code_](https://claude.ai/code/session_01UuYJnStAwUa9Ri8Vag4kwp)

PR #14943, opened 2026-08-11
blockifier: skip cloning declared_contracts in get_os_initial_reads

gkaempfer:
## Summary

Follow-up performance optimization to #14624.

`get_os_initial_reads` builds its return value by calling `get_initial_reads()`, which clones the *entire* `initial_reads` `StateMaps` — including `declared_contracts: HashMap<ClassHash, bool>` — and then immediately calls `.declared_contracts.clear()` on the result, since the OS does not consume that field.

This PR replaces that clone-then-clear with a field-by-field construction of the returned `StateMaps`: `nonces`, `class_hashes`, `storage`, and `compiled_class_hashes` are cloned as before (byte-identical to the previous behavior), while `declared_contracts` is set directly to an empty `HashMap::new()` instead of being cloned and discarded.

**Impact:** for a block that declares many classes, this removes one wasted `HashMap` allocation + copy on the block-finalization hot path (`get_os_initial_reads` runs once per block when OS-initial-reads collection is enabled). No semantic change — `declared_contracts` was always cleared before being returned, so the returned value is unchanged; `StateMaps::default()` (used in the `Skip` arm) already returns an empty map in this field, so this is consistent with existing behavior.

## Test plan

- [x] `cargo build -p blockifier`
- [x] `SEED=0 cargo test -p blockifier --lib cached_state` — all 43 `state::cached_state` tests pass
- [x] `scripts/rust_fmt.sh --check` — clean
- [x] Reviewed by an Opus subagent (correctness of `HashMap::new()` vs. cloned-then-cleared map, `RefCell` borrow scoping, and style) — no blocking findings; two comment-wording nits applied
- [ ] CI green

---

_Generated by_ [_Claude Code_](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01RaNkWM3kEJfFCjEBRfPvB5)_

### DIFF
--- a/crates/blockifier/src/state/cached_state.rs
+++ b/crates/blockifier/src/state/cached_state.rs
@@ -298,7 +298,8 @@ impl<S: StateReader> CachedState<S> {
     /// Returns the pre-block values the OS needs to replay the block: the values read during
     /// execution, extended with the class hash and nonce of every accessed contract and the
     /// compiled class hash of every accessed class (the OS reads these trie leaves even when
-    /// execution itself did not). `declared_contracts` is cleared, as the OS does not consume it.
+    /// execution itself did not). `declared_contracts` is left empty, as the OS does not consume
+    /// it.
     pub fn get_os_initial_reads(&mut self) -> StateResult<StateMaps> {
         // Back-fill write-only storage cells so their pre-block values are part of the initial
         // reads.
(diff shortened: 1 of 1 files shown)


gkaempfer:
This PR is still ready to merge — CI is fully green (all 29 checks passing) and it went through Opus review with no blocking findings. It just hasn't had a human review yet. @gkaempfer — could you take a look when you have a chance?

---

PR #14945, opened 2026-08-11
apollo_batcher,blockifier: bound the resources a call_contract view call may consume

asaf-sw:
Independent of the L1 oracle stack, but a **prerequisite** for it: [#14942](https://github.com/starkware-libs/sequencer/pull/14942) adds the first production consumer of `call_contract`, reading a Chainlink proxy whose implementation class an external owner key can `upgrade`. That makes the called contract adversarial, and today the view path is not bounded for it.

## The problem

`Batcher::call_contract` runs `call_view_entry_point` with `initial_gas = HIGH_GAS_AMOUNT` (10^10). The batcher runs on `LocalComponentServer`, whose `process_requests` loop awaits one `process_request` at a time, and `impl PrioritizedRequest for BatcherRequest {}` leaves everything at Normal priority. So `ProposeBlock`, `ValidateBlock`, and `DecisionReached` all queue behind an in-flight `CallContract`, and a hostile contract that burns its budget stalls block production.

One correction to the framing, since it came up during review: `limit_steps = false` does **not** disable the step cap. It makes `max_steps` fall back to the block limit `invoke_tx_max_n_steps` (10^7) rather than deriving it from the tx's resource bounds. The genuinely unbounded path was `cairo_native`, which never consults `vm_run_resources` and was therefore metered only by the 10^10 gas.

## What this does

`ViewCallResourceBounds` on `call_view_entry_point`, with a single named constructible value `BOUNDED_VIEW_CALL` (`VIEW_CALL_MAX_SIERRA_GAS = 10^8`, `VIEW_CALL_MAX_N_STEPS = 10^6`), plus `EntryPointExecutionContext::cap_remaining_steps`.

**Both** resources are bounded because they bind on different contract kinds, and neither alone is sufficient:
- Sierra gas is the only bound on natively executed contracts.
- Step-tracked execution (Cairo 0, Cairo 1 below `min_sierra_version_for_sierra_gas`, and any nested call under a step-tracked frame) consumes no Sierra gas, so the step cap is its only bound.

The fields are private and there is exactly one constructible value, so "no caller can raise the bound" is structural rather than a convention.

## Consensus safety

`call_view_entry_point` has exactly four callers: `Batcher::call_contract` (the only production one), two blockifier tests, and one `apollo_staking` test. None is on a block-execution, validation, or state-commitment path, and the signature change is compile-forced so none can be silently missed. Execution runs against a throwaway `CachedState` that is discarded on both the Ok and Err paths.

## Bound sizing

Measured: **4.015 Cairo steps/level** and **973 Sierra gas/level** on the recursion contract. Exhausting the budget takes ~2.3s (gas) and ~1.4s (steps) in a **debug** build; release is substantially faster. The legitimate oracle reads (`decimals()`, `latest_round_data()`) need a few thousand steps, well under 1% of the budget.

**Headroom note for `apollo_staking`.** `get_stakers` has no pagination, and at ~3,000 Sierra gas per returned felt the return value alone costs ~12,000 gas per staker before storage reads. `CairoStakingContract` has no production construction site today (`create_committee_provider` builds `MockStakingContract`), so nothing is affected now, but whoever wires it up should confirm the staker set fits, or take a larger bound. The parameter makes that a one-line change.

## Testing

Both bound tests are parameterised over a step-tracked (Cairo 0) and a gas-tracked (Cairo 1) contract, at depths chosen so **each proves its own bound and not the pre-existing one**: 400,000 (~1.6M steps, above 10^6 and below `invoke_tx_max_n_steps`) and 300,000 (~292M gas, above 10^8 and below `HIGH_GAS_AMOUNT`).

Verified by mutation: deleting `cap_remaining_steps` fails the Cairo 0 case; reverting the bounds to the old budget fails the gas case.

A versioned-constants test pins the documented invariants, which caught a false claim in the original: 10^8 equals `validate_max_sierra_gas` only from 0.13.4 on, not "every supported version" (0.13.0-0.13.3 have 10^10).

## Known gap, stated rather than hidden

`MAX_VIEW_CALL_RETDATA_LENGTH = 100_000` is enforced but **deleting the check leaves the suite green**. Measurement showed why: at ~3,040 gas per emitted felt, the 10^8 gas bound caps a gas-tracked call at ~30,000 felts, so the limit is unreachable on that path and only a step-tracked callee could reach it. Covering it end to end needs a new Cairo 0 entry point, which shifts the selector dispatch table and breaks the many blockifier tests that pin exact Cairo 0 step counts. That blast radius did not seem worth it; the isolated tests were strengthened instead. Happy to do it if reviewers disagree.

### DIFF
--- a/crates/apollo_batcher/src/batcher.rs
+++ b/crates/apollo_batcher/src/batcher.rs
@@ -770,6 +770,8 @@ impl Batcher {
         })?
         .map_err(|err| BatcherError::ContractCallFailed { reason: err.to_string() })?;
 
+        validate_retdata_length(retdata.len())?;
+
         Ok(CallContractOutput { retdata })
     }
 
@@ -1535,6 +1537,27 @@ impl Batcher {
             .expect("The commitment offset unexpectedly doesn't match the given block height.");
         Ok(())
     }
+}
+
+/// Maximal number of felts a view entry point call may return, about 3.2 MB at 32 bytes per felt.
+/// Far above the largest legitimate reader (the staking contract's staker list, a few felts per
+/// staker), and small enough to stay cheap to serialize over the batcher's remote server boundary.
+///
+/// Bounds the top-level return value only; the view call's resource bounds are what limit memory
+/// during execution.
+pub(crate) const MAX_VIEW_CALL_RETDATA_LENGTH: usize = 100_000;
+
+/// Rejects a view call whose return value is too large to hand back to the caller.
+pub(crate) fn validate_retdata_length(retdata_length: usize) -> BatcherResult<()> {
+    if retdata_length > MAX_VIEW_CALL_RETDATA_LENGTH {
+        return Err(BatcherError::ContractCallFailed {
+            reason: format!(
+                "Returned {retdata_length} felts, exceeding the limit of \
+                 {MAX_VIEW_CALL_RETDATA_LENGTH}."
+            ),
+        });
+    }
+    Ok(())
 }
 
 /// Logs the result of the transactions execution in the proposal.
--- a/crates/apollo_batcher/src/batcher_test.rs
+++ b/crates/apollo_batcher/src/batcher_test.rs
@@ -82,6 +82,7 @@ use validator::Validate;
 
 use crate::batcher::{
     finished_proposal_info_from_artifacts,
+    validate_retdata_length,
     Batcher,
     BatcherStorageReader,
     BatcherStorageWriter,
@@ -90,6 +91,7 @@ use crate::batcher::{
     StorageCommitmentBlockHash,
     StorageViewStateReaderFactory,
     ViewStateReaderFactory,
+    MAX_VIEW_CALL_RETDATA_LENGTH,
 };
 use crate::block_builder::{
     AbortSignalSender,
@@ -141,6 +143,21 @@ const STAKING_CONTRACT: FeatureContract =
     FeatureContract::MockStakingContract(RunnableCairo1::Casm);
 const ACCOUNT_CONTRACT: FeatureContract =
     FeatureContract::AccountWithoutValidations(CairoVersion::Cairo1(RunnableCairo1::Casm));
+/// Both versions of the test contract expose a `recurse(depth)` entry point whose cost grows with
+/// the requested depth. The Cairo 1 contract is tracked by Sierra gas and the Cairo 0 one by Cairo
+/// steps, so between them they exercise both of the view call resource bounds.
+const SIERRA_GAS_TRACKED_RECURSIVE_CONTRACT: FeatureContract =
+    FeatureContract::TestContract(CairoVersion::Cairo1(RunnableCairo1::Casm));
+const CAIRO_STEPS_TRACKED_RECURSIVE_CONTRACT: FeatureContract =
+    FeatureContract::TestContract(CairoVersion::Cairo0);
+/// Recursion depth whose cost is a small fraction of either view call resource bound.
+const RECURSION_DEPTH_WITHIN_RESOURCE_BOUNDS: u64 = 1_000;
+/// Recursion depths that exhaust a view call resource bound: above `VIEW_CALL_MAX_SIERRA_GAS` for
+/// the Cairo 1 contract, and between `VIEW_CALL_MAX_N_STEPS` and the block's
+/// `invoke_tx_max_n_steps` (10^7) for the Cairo 0 one, so the step case fails on the view bound and
+/// not the block limit. One level of `recurse` costs about 973 Sierra gas and about 4 Cairo steps.
+const SIERRA_GAS_RECURSION_DEPTH_EXCEEDING_RESOURCE_BOUNDS: u64 = 300_000;
+const CAIRO_STEPS_RECURSION_DEPTH_EXCEEDING_RESOURCE_BOUNDS: u64 = 400_000;
 
 struct TestViewStateReaderFactory {
     state: Arc<Mutex<CachedState<DictStateReader>>>,
@@ -2078,4 +2095,92 @@ async fn call_contract_success() {
         .unwrap();
 
     assert_eq!(result.retdata, vec![Felt::ONE, Felt::TWO, Felt::THREE]);
+}
+
+async fn create_batcher_with_recursive_contract(contract: FeatureContract) -> Batcher {
+    let state = test_state(&ChainInfo::create_for_testing(), BALANCE, &[(contract, 1)]);
+    create_batcher(MockDependencies {
+        view_state_reader_factory: Box::new(TestViewStateReaderFactory {
+            state: Arc::new(Mutex::new(state)),
+            expected_block_number: INITIAL_HEIGHT,
+        }),
+        ..Default::default()
+    })
+    .await
+}
+
+fn recurse_call_contract_input(contract: FeatureContract, depth: u64) -> CallContractInput {
+    CallContractInput {
+        contract_address: contract.get_instance_address(0),
+        entry_point: "recurse".to_string(),
+        calldata: vec![Felt::from(depth)],
+    }
+}
+
+#[rstest]
+#[case::sierra_gas_tracked(SIERRA_GAS_TRACKED_RECURSIVE_CONTRACT)]
+#[case::cairo_steps_tracked(CAIRO_STEPS_TRACKED_RECURSIVE_CONTRACT)]
+#[tokio::test]
+async fn call_contract_within_resource_bounds_succeeds(#[case] contract: FeatureContract) {
+    let batcher = create_batcher_with_recursive_contract(contract).await;
+
+    let result = batcher
+        .call_contract(recurse_call_contract_input(
+            contract,
+            RECURSION_DEPTH_WITHIN_RESOURCE_BOUNDS,
+        ))
+        .await
+        .unwrap();
+
+    assert_eq!(result.retdata, vec![]);
+}
+
+#[rstest]
+#[case::sierra_gas_bound(
+    SIERRA_GAS_TRACKED_RECURSIVE_CONTRACT,
+    SIERRA_GAS_RECURSION_DEPTH_EXCEEDING_RESOURCE_BOUNDS,
+    "Out of gas"
+)]
+#[case::cairo_steps_bound(
+    CAIRO_STEPS_TRACKED_RECURSIVE_CONTRACT,
+    CAIRO_STEPS_RECURSION_DEPTH_EXCEEDING_RESOURCE_BOUNDS,
+    "RunResources has no remaining steps."
+)]
+#[tokio::test]
+async fn call_contract_exceeding_resource_bounds_fails(
+    #[case] contract: FeatureContract,
+    #[case] depth: u64,
+    #[case] expected_reason: &str,
+) {
+    let batcher = create_batcher_with_recursive_contract(contract).await;
+
+    let result = batcher.call_contract(recurse_call_contract_input(contract, depth)).await;
+
+    assert_matches!(
+        result,
+        Err(BatcherError::ContractCallFailed { reason }) if reason.contains(expected_reason),
+        "Expected the call to fail with {expected_reason:?}."
+    );
+}
+
+#[rstest]
+#[case::empty_retdata(0)]
+#[case::retdata_below_the_limit(MAX_VIEW_CALL_RETDATA_LENGTH - 1)]
+#[case::retdata_at_the_limit(MAX_VIEW_CALL_RETDATA_LENGTH)]
+fn validate_retdata_length_accepts_lengths_up_to_the_limit(#[case] retdata_length: usize) {
+    assert_eq!(validate_retdata_length(retdata_length), Ok(()));
+}
+
+/// The reason string is all the caller sees, so it must name both the returned length and the
+/// limit.
+#[rstest]
+#[case::retdata_just_above_the_limit(MAX_VIEW_CALL_RETDATA_LENGTH + 1)]
+#[case::retdata_far_above_the_limit(MAX_VIEW_CALL_RETDATA_LENGTH * 10)]
+fn validate_retdata_length_rejects_length_above_the_limit(#[case] retdata_length: usize) {
+    assert_matches!(
+        validate_retdata_length(retdata_length),
+        Err(BatcherError::ContractCallFailed { reason })
+            if reason.contains(&retdata_length.to_string())
+                && reason.contains(&MAX_VIEW_CALL_RETDATA_LENGTH.to_string())
+    );
 }
--- a/crates/blockifier/src/execution/entry_point.rs
+++ b/crates/blockifier/src/execution/entry_point.rs
@@ -18,7 +18,6 @@ use starknet_api::transaction::fields::{
     Calldata,
     ResourceBounds,
     ValidResourceBounds,
-    HIGH_GAS_AMOUNT,
 };
 use starknet_api::transaction::TransactionVersion;
 use starknet_types_core::felt::Felt;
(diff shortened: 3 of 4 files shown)


asaf-sw:
* **#14945** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14945?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14945?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main-v0.14.3`


<h2></h2>

yoavGrs (changes requested):
> @yoavGrs reviewed 3 files and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14945)* status: 3 of 5 files reviewed, 4 unresolved discussions (waiting on asaf-sw and ron-starkware).

___
*[`crates/blockifier/src/execution/entry_point.rs` line 494 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-Ozq1EyM-rNlDSkb4JZu:-Ozq1EyM-rNlDSkb4JZv:brvj4t0) ([raw file](https://github.com/starkware-libs/sequencer/blob/b8350462e6b276a1e72a9d4a33b506da2ec1067f/crates/blockifier/src/execution/entry_point.rs#L494)):*
> ```XML
>     /// Returns the remaining number of steps.
>     pub fn cap_remaining_steps(&mut self, max_n_steps: usize) -> usize {
>         self.subtract_steps(self.n_remaining_steps().saturating_sub(max_n_steps))
> ```

Please define `var = self.n_remaining_steps().saturating_sub(max_n_steps)`
I'm trying to follow the meaning of `x - max(y - z, 0)`
___
*[`crates/apollo_batcher/src/batcher.rs` line 772 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-Ozpmxt6AnokhsMNgxJq:-Ozpmxt7B0ub3byljKxl:bqy8myp) ([raw file](https://github.com/starkware-libs/sequencer/blob/b8350462e6b276a1e72a9d4a33b506da2ec1067f/crates/apollo_batcher/src/batcher.rs#L772)):*
> ```XML
>         );
> 
>         // The batcher serves one request at a time, so block production waits behind this call.
> ```

Please shorten the comment.
Consider moving it down, above `ViewCallResourceBounds::BOUNDED_VIEW_CALL`.
___
*[`crates/apollo_batcher/src/batcher.rs` line 1599 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-Ozpq9Vg1poY1NYcjouC:-Ozpq9Vg1poY1NYcjouD:bbktu46) ([raw file](https://github.com/starkware-libs/sequencer/blob/b8350462e6b276a1e72a9d4a33b506da2ec1067f/crates/apollo_batcher/src/batcher.rs#L1599)):*
> ```XML
> /// what bound that transient memory; this constant bounds what leaves the batcher.
> ///
> /// Which calls this can actually reject: a Sierra-gas tracked call pays about 3,000 Sierra gas per
> ```

So you can return up to 100k felts (32 bytes), which is ~3.2 MB?
I'm a bit lost in this comment.
___
*[`crates/apollo_batcher/src/batcher.rs` line 1607 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-OzpoA3S9SB9-ba7ElXL:-OzpoA3S9SB9-ba7ElXM:b3spepb) ([raw file](https://github.com/starkware-libs/sequencer/blob/b8350462e6b276a1e72a9d4a33b506da2ec1067f/crates/apollo_batcher/src/batcher.rs#L1607)):*
> ```XML
> pub(crate) const MAX_VIEW_CALL_RETDATA_LENGTH: usize = 100_000;
> 
> /// Rejects a view call whose return value is too large to hand back to the caller. This runs after
> ```

Can you end the comment at this point?



_Code quote:_
```XML
to the caller.
```

asaf-sw:
Thanks, all four applied in \`64678d0\`.

1. \`cap_remaining_steps\` now names the intermediate: \`let steps_above_cap = self.n_remaining_steps().saturating_sub(max_n_steps);\`. That is the quantity \`subtract_steps\` wants, so the \`x - max(y - z, 0)\` shape is gone.
2. Shortened to one line and moved down next to \`BOUNDED_VIEW_CALL\`. Dropped the "only value of its type" sentence entirely, since it is already stated on the type itself in \`entry_point.rs\`.
3. Answered your question in the first line: 100k felts at 32 bytes is 3.2 MB. Cut 15 lines to 6, dropping the per-felt gas measurements and the call-tree discussion, which are now in the PR description.
4. Ends at "to the caller."

Your comments prompted a wider pass, since you reviewed 3 of 5 files and the same problem was in the other two. Also removed: wall-clock measurements with a date that nothing pins and that rot invisibly, a duplicated phrase appearing twice 12 lines apart, a threat-model sentence already in the commit message, mutation-test narration on a self-naming test, and a stale reference to \`HIGH_GAS_AMOUNT\` as a live constraint when this PR deletes it. Kept the 973 Sierra gas / 4 Cairo steps per level figures, since those are what someone recalibrating the test depths needs.

Comment-only, verified zero code lines changed. 1104 tests pass.

asaf-sw (commented):
> @asaf-sw+AGNT made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14945)* status: 1 of 5 files reviewed, 4 unresolved discussions (waiting on ron-starkware and yoavGrs).

___
*[`crates/apollo_batcher/src/batcher.rs` line 772 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-Ozpmxt6AnokhsMNgxJq:-OzrSHBm2YYIHxEGYhwP:b-xdprpr) ([raw file](https://github.com/starkware-libs/sequencer/blob/b8350462e6b276a1e72a9d4a33b506da2ec1067f/crates/apollo_batcher/src/batcher.rs#L772)):*
<details><summary><i>Previously, yoavGrs wrote…</i></summary><blockquote>

Please shorten the comment.
Consider moving it down, above `ViewCallResourceBounds::BOUNDED_VIEW_CALL`.
</blockquote></details>

Done in `64678d0`, both halves. Cut to one line and moved next to the argument:

```rust
                // The batcher serves one request at a time, so block production waits behind this call.
                ViewCallResourceBounds::BOUNDED_VIEW_CALL,
```

Dropped the "only value of its type" sentence entirely rather than shortening it: it is already stated on the type in `entry_point.rs`, and `This` had no antecedent when the comment sat above the argument.
___
*[`crates/apollo_batcher/src/batcher.rs` line 1599 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-Ozpq9Vg1poY1NYcjouC:-OzrSHTf4JIIQ2uPi5yN:b57i3ph) ([raw file](https://github.com/starkware-libs/sequencer/blob/b8350462e6b276a1e72a9d4a33b506da2ec1067f/crates/apollo_batcher/src/batcher.rs#L1599)):*
<details><summary><i>Previously, yoavGrs wrote…</i></summary><blockquote>

So you can return up to 100k felts (32 bytes), which is ~3.2 MB?
I'm a bit lost in this comment.
</blockquote></details>

Yes, 100_000 x 32 B = 3.2 MB. That is the first line now, and the comment is 15 lines down to 6:

```rust
/// Maximal number of felts a view entry point call may return, about 3.2 MB at 32 bytes per felt.
/// Far above the largest legitimate reader (the staking contract's staker list, a few felts per
/// staker), and small enough to stay cheap to serialize over the batcher's remote server boundary.
///
/// Bounds the top-level return value only; the view call's resource bounds are what limit memory
/// during execution.
```

The per-felt gas measurements and the call-tree discussion moved to the PR description.

Your comments prompted a wider pass, since you reviewed 3 of 5 files and the same problem was in the other two. Also removed there: dated wall-clock measurements nothing pins, a phrase duplicated verbatim 12 lines apart, mutation-test narration on a self-naming test, and a stale reference to `HIGH_GAS_AMOUNT` as a live constraint when this PR deletes it. Kept the 973 Sierra gas / 4 Cairo steps per level figures, which are what someone recalibrating the test depths needs.
___
*[`crates/apollo_batcher/src/batcher.rs` line 1607 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-OzpoA3S9SB9-ba7ElXL:-OzrSHka6UeXu84mGCZu:b1bhha2) ([raw file](https://github.com/starkware-libs/sequencer/blob/b8350462e6b276a1e72a9d4a33b506da2ec1067f/crates/apollo_batcher/src/batcher.rs#L1607)):*
<details><summary><i>Previously, yoavGrs wrote…</i></summary><blockquote>

Can you end the comment at this point?
</blockquote></details>

Done in `64678d0`, ends exactly there:

```rust
/// Rejects a view call whose return value is too large to hand back to the caller.
```

The dropped clause noted the check runs after execution. That fact now sits on `MAX_VIEW_CALL_RETDATA_LENGTH` two lines up, which says the resource bounds are what limit memory during execution.
___
*[`crates/blockifier/src/execution/entry_point.rs` line 494 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-Ozq1EyM-rNlDSkb4JZu:-OzrSGuX8uKIfivZXjAK:b-suq3nq) ([raw file](https://github.com/starkware-libs/sequencer/blob/b8350462e6b276a1e72a9d4a33b506da2ec1067f/crates/blockifier/src/execution/entry_point.rs#L494)):*
<details><summary><i>Previously, yoavGrs wrote…</i></summary><blockquote>

Please define `var = self.n_remaining_steps().saturating_sub(max_n_steps)`
I'm trying to follow the meaning of `x - max(y - z, 0)`
</blockquote></details>

Done in `64678d0`:

```rust
pub fn cap_remaining_steps(&mut self, max_n_steps: usize) -> usize {
    let steps_above_cap = self.n_remaining_steps().saturating_sub(max_n_steps);
    self.subtract_steps(steps_above_cap)
}
```

`subtract_steps` takes the amount to remove, so the intermediate is the excess over the cap.

yoavGrs (changes requested):
> @yoavGrs reviewed 4 files and all commit messages, made 3 comments, and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14945)* status: all files reviewed, 3 unresolved discussions (waiting on asaf-sw and ron-starkware).

___
*[`crates/blockifier/src/execution/entry_point.rs` line 494 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-OztmXq71PvdTXN7C7fX:-OztmXq71PvdTXN7C7fY:b-3q7put) ([raw file](https://github.com/starkware-libs/sequencer/blob/64678d082e1fc4ebaf3a98efbde5cf5826d509f8/crates/blockifier/src/execution/entry_point.rs#L494)):*
> ```XML
>     /// Returns the remaining number of steps.
>     pub fn cap_remaining_steps(&mut self, max_n_steps: usize) -> usize {
>         let steps_above_cap = self.n_remaining_steps().saturating_sub(max_n_steps);
> ```

non-blocking.
Consider implementing the min-logic by `min` method:

```
let remaining_steps = self.n_remaining_steps();
let new_remaining_steps = min(remaining_steps, max_n_steps);
self.vm_run_resources = RunResources::new(new_remaining_steps);
self.n_remaining_steps()
```
And maybe you'll also find a clean way to share code with `subtract_steps`.
___
*[`crates/blockifier/src/execution/entry_point.rs` line 608 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-OzuYSv0AzGQ4Aocv1ew:-OzuYSv0AzGQ4Aocv1ex:b8c55tb) ([raw file](https://github.com/starkware-libs/sequencer/blob/64678d082e1fc4ebaf3a98efbde5cf5826d509f8/crates/blockifier/src/execution/entry_point.rs#L608)):*
> ```XML
> /// [`VIEW_CALL_MAX_N_STEPS`] Cairo steps, so a call gets the same budget whichever resource it is
> /// tracked by. Both claims are pinned by `view_call_resource_bounds_match_versioned_constants`.
> pub const VIEW_CALL_MAX_SIERRA_GAS: GasAmount = GasAmount(100_000_000);
> ```

Move definitions to the top of the file.
___
*[`crates/apollo_batcher/src/batcher.rs` line 781 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-OzuYel28iiZ-COZt32q:-OzuYel28iiZ-COZt32r:b669mbr) ([raw file](https://github.com/starkware-libs/sequencer/blob/64678d082e1fc4ebaf3a98efbde5cf5826d509f8/crates/apollo_batcher/src/batcher.rs#L781)):*
> ```XML
>                 // The batcher serves one request at a time, so block production waits behind this
>                 // call.
>                 ViewCallResourceBounds::BOUNDED_VIEW_CALL,
> ```

What is the advantage of moving const into a function?
Consider using the const inside the function.

I'm also not convinced that a struct is needed, and it's not better to settle for two consts.

asaf-sw (commented):
> @asaf-sw+AGNT made 3 comments and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14945)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on ron-starkware and yoavGrs).

___
*[`crates/apollo_batcher/src/batcher.rs` line 781 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-OzuYel28iiZ-COZt32q:-Ozuw1MB2QZinH88S92_:b-q645ai) ([raw file](https://github.com/starkware-libs/sequencer/blob/64678d082e1fc4ebaf3a98efbde5cf5826d509f8/crates/apollo_batcher/src/batcher.rs#L781)):*
<details><summary><i>Previously, yoavGrs wrote…</i></summary><blockquote>

What is the advantage of moving const into a function?
Consider using the const inside the function.

I'm also not convinced that a struct is needed, and it's not better to settle for two consts.
</blockquote></details>

You're right, the struct is gone in `84f11ee`.

I had it as a parameter to make the bounds unwidenable by a caller, but that reasoning was backwards: applying the two consts *inside* `call_view_entry_point` and dropping the parameter is both simpler and strictly stronger, since there is no longer a parameter to pass a wider value through. So it is now:

```rust
pub fn call_view_entry_point(
    state_reader: impl StateReader,
    block_context: Arc<BlockContext>,
    storage_address: ContractAddress,
    entry_point_name: &str,
    calldata: Calldata,
) -> EntryPointExecutionResult<CallInfo> {
    let mut remaining_gas = VIEW_CALL_MAX_SIERRA_GAS;
    ...
    context.cap_remaining_steps(VIEW_CALL_MAX_N_STEPS);
```

I checked it costs no coverage before doing it: all three call sites passed `BOUNDED_VIEW_CALL` and nothing varied it, and the cap is tested directly against `cap_remaining_steps` rather than through this parameter. Dropping it also reverted `cairo_staking_contract_test.rs` to untouched, so the PR is 4 files instead of 5.

The per-field docs moved onto the consts themselves, and the "why bounded at all" reason onto the function. `crates/blockifier`, `apollo_batcher` and `apollo_staking` all compile, and `test_call_view_entry_point`, `test_call_view_entry_point_non_existing_function`, `cap_remaining_steps_lowers_the_limit`, `cap_remaining_steps_never_raises_the_limit`, `view_call_resource_bounds_match_versioned_constants` and the 98 `apollo_staking` tests pass.
___
*[`crates/blockifier/src/execution/entry_point.rs` line 494 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-OztmXq71PvdTXN7C7fX:-Ozuvug-4XGB-NSlanmt:b-ug3c2f) ([raw file](https://github.com/starkware-libs/sequencer/blob/64678d082e1fc4ebaf3a98efbde5cf5826d509f8/crates/blockifier/src/execution/entry_point.rs#L494)):*
<details><summary><i>Previously, yoavGrs wrote…</i></summary><blockquote>

non-blocking.
Consider implementing the min-logic by `min` method:

```
let remaining_steps = self.n_remaining_steps();
let new_remaining_steps = min(remaining_steps, max_n_steps);
self.vm_run_resources = RunResources::new(new_remaining_steps);
self.n_remaining_steps()
```
And maybe you'll also find a clean way to share code with `subtract_steps`.
</blockquote></details>

Done in `84f11ee`. `min` reads better than "subtract the excess", so I took it, and shared the write with `subtract_steps` through a private `set_remaining_steps`:

```rust
fn set_remaining_steps(&mut self, n_steps: usize) -> usize {
    self.vm_run_resources = RunResources::new(n_steps);
    self.n_remaining_steps()
}

pub fn subtract_steps(&mut self, steps_to_subtract: usize) -> usize {
    // Saturating, since subtracting more steps than remain would underflow.
    let new_remaining_steps = self.n_remaining_steps().saturating_sub(steps_to_subtract);
    self.set_remaining_steps(new_remaining_steps)
}

pub fn cap_remaining_steps(&mut self, max_n_steps: usize) -> usize {
    let new_remaining_steps = min(self.n_remaining_steps(), max_n_steps);
    self.set_remaining_steps(new_remaining_steps)
}
```

Each now states its own rule and neither reaches into `vm_run_resources` directly.
___
*[`crates/blockifier/src/execution/entry_point.rs` line 608 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-OzuYSv0AzGQ4Aocv1ew:-OzuvxvxBJj7QgGmZkmI:bbeciar) ([raw file](https://github.com/starkware-libs/sequencer/blob/64678d082e1fc4ebaf3a98efbde5cf5826d509f8/crates/blockifier/src/execution/entry_point.rs#L608)):*
<details><summary><i>Previously, yoavGrs wrote…</i></summary><blockquote>

Move definitions to the top of the file.
</blockquote></details>

Done in `84f11ee`. Both consts now sit directly under `FAULTY_CLASS_HASH` at the top of the file, which is also where the file already keeps its consts.

yoavGrs (approved):
:lgtm:

> @yoavGrs reviewed 5 files and all commit messages, made 1 comment, and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14945#-:-Ozuyl7nBSMZAaHwVhal:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ron-starkware).

gkaempfer:

---

gkaempfer:

---

PR #14823, opened 2026-07-15
apollo_consensus_orchestrator: add CendeContext::get_accessed_keys_input

yoavGrs:
New os_input-gated trait method plus CendeAmbassador implementation that
fetches a committed block's accessed keys from the recorder, returning None
when the recorder has none stored. Reuses the shared send_recorder_get helper
(extracted from fetch_block_number) for the request/status/parse error
handling, and adds the get_accessed_keys endpoint path/URL. No callers yet.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

PR #14949, opened 2026-08-12
apollo_consensus_orchestrator: cap recorder response size in send_recorder_get

gkaempfer:
## Summary

Automated security scan of #14823, which merged `CendeContext::get_accessed_keys_input`.

That endpoint fetches per-block data from the Cende recorder and deserializes the response body directly via `response.json::<T>()` in the shared `send_recorder_get` helper, with no limit on response size. Previously, every caller of this helper only ever deserialized a tiny fixed-shape `BlockNumberResponse { block_number: Option<u64> }`. `get_accessed_keys_input` is the first caller to deserialize into `BlockAccessedKeysData { transactions: Vec<CentralTransactionWritten>, execution_infos: Vec<CentralTransactionExecutionInfo> }` — collections whose length and nested field sizes (calldata, signatures, paymaster data, etc.) are entirely controlled by the response body. The same PR also removed the `#[cfg(feature = "testing")]` gate on `Deserialize` for these central transaction types, so they are now deserializable in production builds, not just tests.

**Vulnerability (CWE-400, Uncontrolled Resource Consumption):** a malicious or compromised recorder (or a response with an absent/understated `Content-Length`) can return an arbitrarily large body. Because `response.json` buffers the entire body before parsing, and the request timeout only bounds wall-clock time (not bytes-per-second), a fast, oversized response can drive unbounded memory allocation in the consensus orchestrator — a liveness-critical process per this repo's own architecture (it orchestrates block production). This matches the project's own documented guidance in `.claude/rules/code-style.md`: *"Any value deserialized from an HTTP request, query parameter, or other external input must be assumed hostile... Cap allocations derived from user input with hard limits."*

## Fix

`send_recorder_get` now streams the response body via `bytes_stream()` and rejects it with `RecorderRequestFailed` as soon as accumulated bytes exceed a fixed `MAX_RECORDER_RESPONSE_BYTES` cap (64 MiB), before ever calling into `serde_json`. The cap is applied uniformly to all recorder GET endpoints through the shared helper. The size limit is parameterized (`send_recorder_get_with_limit`) so it can be exercised in tests without buffering tens of megabytes.

## Test plan

- [x] Added `send_recorder_get_rejects_response_over_the_size_limit`: a response over a configured byte cap is rejected with `RecorderRequestFailed` instead of being parsed.
- [x] Added `send_recorder_get_accepts_response_within_the_size_limit`: a response within the cap still parses correctly (no regression).
- [x] Existing `get_accessed_keys_input` test cases (parses `Some`, `None`, error status) still pass unmodified.
- [x] `cargo test -p apollo_consensus_orchestrator cende::` — 39 passed, 0 failed.
- [x] `scripts/rust_fmt.sh` — clean.
- [x] `cargo clippy -p apollo_consensus_orchestrator --all-targets --features testing -- -D warnings` — clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01ANTjUoyz2ak7VdR9wsv2uw)_

### DIFF
--- a/crates/apollo_consensus_orchestrator/Cargo.toml
+++ b/crates/apollo_consensus_orchestrator/Cargo.toml
@@ -35,7 +35,7 @@ ethnum.workspace = true
 futures.workspace = true
 indexmap.workspace = true
 num-rational.workspace = true
-reqwest = { workspace = true, features = ["json"] }
+reqwest = { workspace = true, features = ["json", "stream"] }
 reqwest-middleware = { workspace = true, features = ["json"] }
 reqwest-retry.workspace = true
 serde.workspace = true
--- a/crates/apollo_consensus_orchestrator/src/cende/cende_test.rs
+++ b/crates/apollo_consensus_orchestrator/src/cende/cende_test.rs
@@ -20,7 +20,13 @@ use super::{
     RECORDER_GET_LATEST_RECEIVED_BLOCK_PATH,
     RECORDER_WRITE_BLOB_PATH,
 };
-use crate::cende::{BlobParameters, CendeAmbassadorError, CendeAmbassadorResult, CendeConfig, CendeContext};
+use crate::cende::{
+    BlobParameters,
+    CendeAmbassadorError,
+    CendeAmbassadorResult,
+    CendeConfig,
+    CendeContext,
+};
 use crate::metrics::{
     register_metrics,
     CendeWriteFailureReason,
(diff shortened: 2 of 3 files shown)


gkaempfer:
An independent review pass (Opus) found that the non-success-status branch of `send_recorder_get_with_limit` still read the error body via `response.text()`, unbounded — so a recorder answering with any non-2xx status and an oversized body would bypass the size cap this PR adds for the success path. Pushed a follow-up commit that extracts the streaming-with-cap logic into `read_response_body_prefix` and applies a (smaller, diagnostics-sized) cap to the error branch too, plus tightened the size-limit tests and added a regression test for the oversized-error-body case. All tests, `rustfmt`, and `clippy -D warnings` pass on the current head.

---

PR #14951, opened 2026-08-12
apollo_deployments: assert every service composes a loadable node config

asaf-sw:
Adds a fast test that catches a config parameter missing from the deployment app configs. Today that crash-loops every node and is only caught by `system_test_hybrid`, about 20 minutes later. The new test fails in 0.28s naming the missing key.

---

## Problem

`SequencerNodeConfig::load_and_process` loads with `ignore_default_values = true`, so schema defaults are discarded at runtime and every value must come from `apollo_deployments/resources/app_configs/`. A param added to the schema but missing there crash-loops every node with `MissingParam`.

Addresses the existing `TODO(Tsabary)` in `deployment_definitions_test.rs`.

## Change

`services_load_a_valid_node_config` composes, per node type and service, the same file set a deployment passes to `--config_file`, then asserts the load succeeds and `validate_node_config_without_urls` passes.

Two details make the real files loadable outside a cluster:

- The per-service file exists on disk only in replacer form, so `NodeType::service_component_configs` exposes the pruned JSON before `insert_replacer_annotations`. The dump path is unchanged.
- `KEYS_TO_BE_REPLACED` values are injected by the deployment infra, so they are stripped and replaced with the schema's own typed defaults. Coverage is not lost: `replacer_config_entries_are_in_config` already asserts every replaced key is present in the app configs.

## Verification

6/6 pass. Removing one required key from `app_configs/consensus_manager_config.json` fails in 0.28s with ``missing field `base` ``, naming the service.

### DIFF
--- a/.claude/CLAUDE.md
+++ b/.claude/CLAUDE.md
@@ -63,6 +63,7 @@ ci: update rust toolchain
 - **Cairo native feature**: When modifying `cairo_native`-gated code, verify with `cargo check -p blockifier --features cairo_native`
 - **CI flaky tests**: `blockifier_reexecution` can fail from transient GCloud network issues (not code-related). `merge-gatekeeper` fails when other checks fail (downstream).
 - **Feature flag removal**: `define_infra_metrics!` and `define_metrics!` macros expand code with `cfg(feature = "testing")`. Crates using these macros MUST keep `testing = []` in Cargo.toml even if it looks unused — `check-cfg` will error otherwise. Grep for macro invocations, not just direct `cfg` checks.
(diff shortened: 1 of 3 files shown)


asaf-sw:
Closing: superseded by #15096.

The jsonnet config migration removed everything this PR builds on: `KEYS_TO_BE_REPLACED`, the replacer dump path, `resources/app_configs/`, and the defaults-discarding `load_and_process` path that caused the crash loop. The node now loads a jsonnet-built ConfigMap JSON plus a secrets file, with plain serde structs.

`assert_build_deserializes` in `crates/apollo_deployments/src/jsonnet_tests.rs` covers the same gap on the new pipeline: per layout, it evaluates `build.libsonnet` and, for every service, deserializes the output into `SequencerNodeConfig` and runs `validate_node_config`. A struct field the build does not emit fails there instead of at node boot.

The CLAUDE.md line in this PR refers to the removed directory and test, so nothing here is worth carrying forward.

PR #14995, opened 2026-08-18
apollo_l1_gas_price_config,apollo_deployments,apollo_node: add ExchangeRateOracleSource

asaf-sw:
Adds `ExchangeRateOracleSource` (`Http` default, `Chainlink`) and one `*_oracle_source` field per feed, so a feed moves off the HTTP oracle by editing a config key rather than by deploying code. The two feeds are selected independently so they can be migrated one at a time. Also rewrites the micro-unit bound descriptions, which were written for ETH/USD but served under `strk_usd` too, where the worked example disagreed with the shipped default by five orders of magnitude.

B3 of the split of #14944. 2 new parameters.

- Deferred: both feeds default to `Http` and nothing reads either field. B4 is the first PR whose behavior depends on them, so merging this changes nothing at runtime.
- Scheduled for deletion: the enum, both fields, both `ser_param`s and the three tests all go in C1 with the HTTP oracle. The enum carries a `TODO(asaf-sw)` naming that removal, which is why the switch is deliberately thin. The description rewrite is the part that survives.
- Worth a look: `unrecognized_oracle_source_fails_to_load` pins that the enum takes no `serde(rename_all)`, so a lowercase `"chainlink"` cannot silently fall back to `Http` and produce a node that looks migrated.

---

## Detailed Summary for AI Bots

Stacked on #14994. Part of the [L1 price oracle replacement](https://starkwareindustries.slack.com/archives/D0ACHD9K27N/p1786280310659939) split of #14942.

## What

`ExchangeRateOracleSource` (`Http` default, `Chainlink`) and the two `*_oracle_source` fields that carry it, one per feed, so a feed moves off the HTTP oracle by editing a config key rather than by deploying code. 2 new parameters, no existing parameter's value or privacy touched.

Both feeds default to `Http`, and nothing reads either field yet: B4 builds each feed's client from its source and is the first PR whose behavior depends on these values. Merging this changes nothing at runtime.

The two feeds are selected independently so they can be migrated one at a time, and so the second flip can wait on a week of evidence from the first.

## Every line of the switch has a deletion date

The enum, both fields, both `ser_param`s and the three tests below all go away in C1, together with the HTTP oracle they exist to switch away from. The enum carries a `TODO(asaf-sw)` naming that removal. Nothing here is built to last, so it is deliberately thin: an enum, two `pub` fields, a `Default`, and no accessor, no helper, no validation beyond what serde already rejects.

What survives C1 is the description rewrite in the second half of this PR, which is about micro units rather than about sources.

[...]

PR #14983, opened 2026-08-18
apollo_l1_gas_price_config: Chainlink oracle config

asaf-sw:
Adds `ChainlinkOracleConfig`, the mechanics half of the Chainlink feed source: the two mainnet proxy `feed_address`es, the `FreshnessWindow` a round's `updated_at` must fall in, a 900 second `sampling_interval_seconds` and a 60 second `failure_retry_interval_seconds`, with a validator keeping the forward freshness bound strictly below the backward one.

A6b of the split of #14942. The rate bounds that sit alongside these fields on the source branch were extracted into `RateBoundsConfig` in the base PR, so this carries only Chainlink-specific fields.

- `sampling_interval_seconds` is deliberately a new field: the source branch read the HTTP source's `lag_interval_seconds` instead, so the Chainlink path no longer reaches into the HTTP config and deleting that source later stays a pure deletion.
- Deferred: nothing reads this config. The client that reads the feeds arrives in A9 and the source is wired in after that. The config is still un-nested, so `config_schema.json` and the app configs are untouched until B2 (#14994).

---

## Detailed Summary for AI Bots

Adds `ChainlinkOracleConfig`, the mechanics half of the Chainlink price-feed source: the two mainnet proxy `feed_address`es, the `FreshnessWindow` a round's `updated_at` must fall in, and the two intervals the source is read on.

The rate bounds that live alongside these fields in the original branch were already extracted into `RateBoundsConfig` in the base PR, so this one carries only the Chainlink-specific fields.

### Fields

- `eth_usd_feed_address` / `strk_usd_feed_address`: Chainlink proxy addresses on Starknet mainnet. The proxies rather than the aggregators behind them, because aggregators are rotated without notice.
- `freshness`: `max_staleness_seconds` (the 24h feed heartbeat plus an hour of margin) and `max_future_updated_at_seconds` (300s of clock skew). A schema-level validator keeps the forward bound strictly below the backward one, since both are plain second counts and exchanging them would otherwise pass every per-field check.
- `sampling_interval_seconds`: 900 by default, `range(min = 1)`.
- `failure_retry_interval_seconds`: 60 by default, `range(min = 1)`.

### `sampling_interval_seconds` is deliberately new

In the source branch, the Chainlink factory read the sampling interval off the HTTP source's `ExchangeRateOracleConfig::lag_interval_seconds`. This PR replaces that read with a field of its own, for two reasons:

- The Chainlink path never has to reach into the HTTP source's config to learn how often to sample.
[...]

PR #14994, opened 2026-08-18
apollo_l1_gas_price_config,apollo_deployments,apollo_node: nest the Chainlink config

asaf-sw:
Nests `RateBoundsConfig` and `ChainlinkOracleConfig` under `L1GasPriceProviderConfig`, both `#[validate(nested)]`, so they reach a node through `config_schema.json` and the deployment app configs instead of existing only as types nothing loads. 12 new parameters, no existing parameter's value, privacy or description touched.

B2 of the split of #14944, and the only PR besides B3 and C1 in either stack that touches the schema and the app configs.

- This is what makes the validation reachable. `validator` only walks into a nested struct from a field marked `#[validate(nested)]`, so until now every check added in the config PRs ran from its own unit tests alone, including the `range(min = 1)` on both intervals.
- Deferred: nothing reads the keys. B3 adds the per-feed source switch and B4 builds the client that reads these configs, and every feed stays on HTTP through both, so this is a pure schema and app-config addition with no behavior change.
- Worth a look: the hand-added app-config keys. A key in the schema but missing from the app configs fails no unit test and crash-loops every pod at startup, surfacing only as a pod readiness timeout.

---

## Detailed Summary for AI Bots

Stacked on #14993. Part of the [L1 price oracle replacement](https://starkwareindustries.slack.com/archives/D0ACHD9K27N/p1786280310659939) split of #14942.

## What

Nests `RateBoundsConfig` and `ChainlinkOracleConfig` under `L1GasPriceProviderConfig`, both `#[validate(nested)]`, so the two configs reach a node from `config_schema.json` and the deployment app configs instead of existing only as types nothing loads.

12 new parameters, no existing parameter's value, privacy, or description touched: 6 under `rate_bounds_config` (a minimum and a maximum for each of `eth_usd`, `strk_usd`, `eth_to_fri`) and 6 under `chainlink_oracle_config` (the two feed addresses, the two intervals, and the two `freshness` bounds).

One `chainlink_oracle_config` serves both feeds, unlike the per-feed `ExchangeRateOracleConfig`s: the ETH/STRK rate is derived from the same two on-chain feeds the STRK/USD rate is read from, so there is nothing per-feed to configure. `rate_bounds_config` sits at provider level for a different reason: the bounds describe the rate rather than the source, so they hold whichever source reports it and they survive the HTTP source's removal in place.

## This is what makes the validation reachable

[...]

PR #14992, opened 2026-08-18
apollo_l1_gas_price,apollo_l1_gas_price_types: hold Chainlink query failures for the retry interval

asaf-sw:
Holds a failed Chainlink query's error in `OracleState::last_error` and spawns the next query once `failure_retry_interval_seconds` elapses rather than on every call. Without it, every `fetch_rate` against a broken feed spawns a fresh query, two view calls per feed per proposal, for as long as the feed stays broken.

A9b of the split of #14942, on top of the A9a lifecycle.

- Precedence: a held failure is served only when no valid read is held, and a success clears it, so the client never reports a failure older than its newest read. `no_call_is_denied_a_rate_the_client_holds` is the regression test for the availability bug found in review of #14942, where the one call that observed a failing query finish was handed that failure without consulting the rate the client already held.
- Deferred: bounding how far a held read may travel from the block timestamp it is served to is the next slice, #14993. The client is still constructed only by tests until B4.

---

## Detailed Summary for AI Bots

Stacked on #14991. Part of the [L1 price oracle replacement](https://starkwareindustries.slack.com/archives/D0ACHD9K27N/p1786280310659939) split of #14942.

## What

Negative caching for the Chainlink client: a failed query's error is held in `OracleState::last_error` and served to callers with nothing better, and the next query is spawned once `failure_retry_interval_seconds` elapses instead of on every call.

## Why hold a failure

Without it, every `fetch_rate` against a failing feed spawns a fresh query: two view calls per feed per attempt, on every proposal, for as long as the feed stays broken. A hostile or broken feed would otherwise cost blocking VM executions per block. Holding the failure bounds the retry cadence to one query per retry interval, which is also why the interval is its own config key rather than the sampling interval: a transient failure costs one retry interval, not the rest of the sampling interval.

## The precedence rule

A held failure is served only when no valid read is held. Bugbot's first review of #14942 caught an availability regression in exactly this spot: the one call that observed a failing query finish was returned that failure directly, without consulting the rate the client already held. `no_call_is_denied_a_rate_the_client_holds` is the regression test, asserting every call through the failure's harvest is served the held rate; `a_held_failure_does_not_mask_the_last_valid_rate` pins the steady state after the failure is held.

A success clears `last_error`, so the client never reports a failure older than its newest read.

## Landing state

- The client is constructed only by tests until B4 wires it into the factory; `metrics().register()` runs under #14977's `Once` guard.
- The module doc names source selection, which arrives in B3/B4.
[...]

PR #14996, opened 2026-08-19
apollo_l1_gas_price,apollo_l1_gas_price_config,apollo_node: build oracle clients per feed source

asaf-sw:
Builds each feed's oracle client from the source #14995 made configurable, so enabling Chainlink for a feed becomes a config rollout rather than a code deploy.

Both feeds default to `Http`, so **merging changes nothing at runtime** — the same two `ExchangeRateOracleClient`s are built from the same two configs as before.

Also closes a transposition the previous tests could not see: crossing a feed's config with another feed's metrics bundle would have published every ETH/STRK reading under `snip35_strk_usd_*`, pointing every dashboard and alert at the wrong series. The factory now has no positional argument left to cross, and every assertion is a number rather than a `Debug` string.

Stacked on #14995. Part of the [L1 price oracle replacement](https://starkwareindustries.slack.com/archives/D0ACHD9K27N/p1786280310659939) split of #14942. Everything the source switch adds is deleted with the HTTP source.

---

### What

`new_with_oracle` builds each feed's client from the source #14995 made configurable. Both feeds default to `Http`, so merging changes nothing at runtime.

The factory is generic over the feed's rate kind, and everything about the feed comes from that kind: its selected source, the HTTP config that source reads, the metrics bundle it publishes on, the Chainlink feeds and bounds it reads, and the config key named when the selection cannot be honoured. `new_with_oracle` passes no per-feed argument at all, only the whole config and the optional batcher client.

### The transposition that tests could not see

`new_with_oracle` on the base branch pairs each feed's HTTP config with that feed's metrics constant positionally, and no test constructs it at all, so crossing either pair is invisible: every ETH/STRK reading would publish under `snip35_strk_usd_*` and vice versa, pointing every dashboard and alert at the wrong series.

The factory in #14944 kept that shape one level down. It took the feed's source, HTTP config and source config key as positional arguments independent of the rate kind, so any two of them could still be crossed. Its HTTP cases read the rates and the series back per feed, but its Chainlink cases asserted `format!("{client:?}")` substrings, `starts_with("ChainlinkOracleClient")` and `pair: EthStrk`, over byte-identical default configs, and never read a rate through a Chainlink client.

This PR closes it from both sides. Structurally, a private `SourceSelectableFeed` impl per rate kind holds that feed's source, HTTP config and config key, so the factory has no argument left to transpose, and `ChainlinkRate::metrics()` remains the only place a pair maps to a metrics bundle. Behaviourally, every assertion is now a number rather than a `Debug` string:

- Two `mockito` servers and a mocked batcher are wired in every case of `each_feed_is_served_by_the_source_selected_for_it`, and the four rates the two sources serve are distinct. A feed built from the other feed's source, config or rate kind resolves to a number named for that other feed.
- `each_feed_publishes_on_its_own_metrics_series` runs on each source and reads both rates back off a local Prometheus recorder, so a client holding the other feed's bundle fails on the series whichever source built that client.

Each fix was verified by mutation: transposing the per-feed source, the per-feed HTTP config, the rate kind at a call site, the source config key, and the metrics bundle each fails a test named for it, as do hardcoding one pair's bundle where `ChainlinkOracleClient::new` reads `Kind::metrics()` and a `MissingBatcherClientError` message that stops naming the offending keys. Nothing in this PR asserts on a `Debug` string.

### One deleted test, and why

#14944 had `chainlink_clients_sample_on_their_own_feeds_interval`, pinning that a Chainlink client samples on its feed's `ExchangeRateOracleConfig::lag_interval_seconds`. #14983 moved the Chainlink cadence into `ChainlinkOracleConfig::sampling_interval_seconds`, which both feeds share, so the per-source cadence distinction that test protected is gone by design and the factory threads no interval at all. That also removes the `NonZeroU64` conversion #14944 needed, and with it the `range(min = 1)` that PR had to add to `lag_interval_seconds` to keep an operator zero from surfacing as a node panic.

### Misconfiguration behaviour

Selecting `Chainlink` where no batcher client exists fails while the components are built, rather than falling back to `Http` or failing per proposal. `MissingBatcherClientError` resolves both feeds before reporting either, so one startup names every offending key: an operator who set both feeds wrong learns of both at once instead of paying two restart cycles for one mistake. Every shipped topology (consolidated, hybrid, distributed) gives the L1 service a batcher client, so this is reachable only by misconfiguration.

`components.rs` panics on the error, at the composition root where a config mistake should stop the process.

### Everything here has a deletion date

The HTTP source removal deletes `MissingBatcherClientError`, the factory's `Http` arm, and both `*_ORACLE_SOURCE_CONFIG_KEY` consts, together with the `SourceSelectableFeed` selection that exists only to choose between the two sources. All are marked. The error type is worth its lines even so: during the window an operator is actively editing these keys, and without it a misconfigured node dies on an `expect` inside the factory, with a message that names neither the feed that was misconfigured nor the key that misconfigured it.

The two `ChainlinkRate` impls get their production callers here: `ChainlinkOracleClient` was constructed only by tests until now, so `Kind::metrics()` and the guard counters were reachable only under `cfg(test)`.

### Testing

10 new cases, 99 in the crate. The four `each_feed_is_served_by_the_source_selected_for_it` cases cover both feeds on each source and each feed alone on Chainlink; the two `each_feed_publishes_on_its_own_metrics_series` cases cover the metrics wiring under each source; `http_sources_do_not_need_a_batcher_client` and the three `chainlink_source_without_a_batcher_client_is_rejected` cases cover the batcher requirement, the rejection cases each asserting that the rendered error names every offending config key verbatim.

The Chainlink cases read through the batcher fixtures the `chainlink_oracle` tests already use, whose two feed answers derive to one rate for ETH/STRK and read as another for STRK/USD. Only the entry point composing them is new; the `test_utils` module and the three fixture consts the provider test names widened to `pub(crate)`.

### DIFF
--- a/crates/apollo_l1_gas_price/src/chainlink_oracle/mod.rs
+++ b/crates/apollo_l1_gas_price/src/chainlink_oracle/mod.rs
@@ -39,7 +39,7 @@ mod feed_read;
 #[cfg(test)]
 mod test;
 #[cfg(test)]
-mod test_utils;
+pub(crate) mod test_utils;
 
 // How many sampling intervals may separate the last valid read's block timestamp from the block
 // timestamp being served, while that read is still served. Three read intervals, 45 minutes at the
@@ -108,8 +108,6 @@ pub trait ChainlinkRate: RateKind {
 /// different rounds for the same block timestamp. Chainlink's deviation threshold is far inside the
 /// `l1_gas_price_margin_percent` validators compare within, so this is not expected to reject
 /// proposals.
-// [Temporary comment] Constructed only by tests; B4 builds it for a feed whose configured oracle
-// source is Chainlink.
 pub struct ChainlinkOracleClient<Kind: ChainlinkRate> {
     // `Arc` so a spawned query captures a refcount bump rather than a copy of both configs.
     config: Arc<ChainlinkOracleConfig>,
--- a/crates/apollo_l1_gas_price/src/chainlink_oracle/test_utils.rs
+++ b/crates/apollo_l1_gas_price/src/chainlink_oracle/test_utils.rs
@@ -23,7 +23,7 @@ pub(super) const FEED_DECIMALS: u32 = 8;
 
 /// The block timestamp every fixture below is dated against, and every freshness bound measured
 /// against.
-pub(super) const TIMESTAMP: u64 = 1_700_000_000;
+pub(crate) const TIMESTAMP: u64 = 1_700_000_000;
 
 /// $3000 per ETH at `FEED_DECIMALS`.
 pub(super) const ETH_USD_ANSWER: u128 = 300_000_000_000;
@@ -31,9 +31,9 @@ pub(super) const ETH_USD_ANSWER: u128 = 300_000_000_000;
 pub(super) const STRK_USD_ANSWER: u128 = 3_000_000;
 
 /// $0.03 per STRK at `EXCHANGE_RATE_DECIMALS`, which `STRK_USD_ANSWER` rescales to.
-pub(super) const STRK_TO_USD_RATE: ExchangeRate = 30_000_000_000_000_000;
+pub(crate) const STRK_TO_USD_RATE: ExchangeRate = 30_000_000_000_000_000;
 /// 100,000 STRK per ETH at `EXCHANGE_RATE_DECIMALS`, which the two feed answers derive to.
-pub(super) const ETH_TO_FRI_RATE: ExchangeRate = 100_000 * EXCHANGE_RATE_SCALE;
+pub(crate) const ETH_TO_FRI_RATE: ExchangeRate = 100_000 * EXCHANGE_RATE_SCALE;
 
 /// The mocked reply per feed address and entry point. A call with no entry here fails.
 pub(super) type FeedResponses = HashMap<(ContractAddress, String), Vec<Felt>>;
@@ -108,6 +108,15 @@ pub(super) fn eth_and_strk_responses(eth_usd: FeedFixture, strk_usd: FeedFixture
 
 pub(super) fn batcher_client_from_responses(responses: FeedResponses) -> SharedBatcherClient {
     counting_batcher_client(responses).0
+}
+
+/// Serves both feeds fresh at every call, so a `StrkToUsd` read resolves to `STRK_TO_USD_RATE` and
+/// an `EthToFri` read derives to `ETH_TO_FRI_RATE`.
+pub(crate) fn batcher_client_serving_fresh_feeds() -> SharedBatcherClient {
+    batcher_client_from_responses(eth_and_strk_responses(
+        FeedFixture::new(ETH_USD_ANSWER, fresh_updated_at()),
+        FeedFixture::new(STRK_USD_ANSWER, fresh_updated_at()),
+    ))
 }
 
 /// A batcher client alongside the number of calls made through it.
--- a/crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs
+++ b/crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs
@@ -2,29 +2,35 @@ use std::any::type_name;
 use std::collections::VecDeque;
 use std::sync::Arc;
 
+use apollo_batcher_types::communication::SharedBatcherClient;
 use apollo_infra::component_definitions::ComponentStarter;
 use apollo_infra_utils::info_every_n_ms;
-use apollo_l1_gas_price_config::config::L1GasPriceProviderConfig;
+use apollo_l1_gas_price_config::config::{
+    ExchangeRateOracleConfig,
+    ExchangeRateOracleSource,
+    L1GasPriceProviderConfig,
+};
 use apollo_l1_gas_price_types::errors::L1GasPriceProviderError;
 use apollo_l1_gas_price_types::{
+    EthToFri,
     ExchangeRate,
     ExchangeRateOracleClientTrait,
     GasPriceData,
     L1GasPriceProviderResult,
     PriceInfo,
+    StrkToUsd,
 };
 use async_trait::async_trait;
 use starknet_api::block::BlockTimestamp;
 use tracing::{info, trace, warn};
 
+use crate::chainlink_oracle::{ChainlinkOracleClient, ChainlinkRate};
 use crate::exchange_rate_oracle::ExchangeRateOracleClient;
 use crate::metrics::{
     register_provider_metrics,
-    ETH_TO_STRK_ORACLE_METRICS,
     L1_DATA_GAS_PRICE_LATEST_MEAN_VALUE,
     L1_GAS_PRICE_LATEST_MEAN_VALUE,
     L1_GAS_PRICE_PROVIDER_INSUFFICIENT_HISTORY,
-    STRK_TO_USD_ORACLE_METRICS,
 };
 
 #[cfg(test)]
@@ -82,16 +88,16 @@ impl L1GasPriceProvider {
         }
     }
 
-    pub fn new_with_oracle(config: L1GasPriceProviderConfig) -> Self {
-        let eth_to_strk_oracle_client = ExchangeRateOracleClient::new(
-            config.eth_to_strk_oracle_config.clone(),
-            ETH_TO_STRK_ORACLE_METRICS,
-        );
-        let strk_to_usd_oracle_client = ExchangeRateOracleClient::new(
-            config.strk_to_usd_oracle_config.clone(),
-            STRK_TO_USD_ORACLE_METRICS,
-        );
-        Self::new(config, Arc::new(eth_to_strk_oracle_client), Arc::new(strk_to_usd_oracle_client))
+    /// Builds each feed's oracle client from the source selected for it in `config`.
+    pub fn new_with_oracle(
+        config: L1GasPriceProviderConfig,
+        batcher_client: SharedBatcherClient,
+    ) -> Self {
+        Self::new(
+            config.clone(),
+            build_exchange_rate_oracle_client::<EthToFri>(&config, &batcher_client),
+            build_exchange_rate_oracle_client::<StrkToUsd>(&config, &batcher_client),
+        )
     }
 
     pub fn initialize(&mut self) -> L1GasPriceProviderResult<()> {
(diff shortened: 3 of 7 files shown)


asaf-sw:
* **#14996** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14996?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14996?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#14995** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14995?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14994** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14994?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14993** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14993?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#14997](https://github.com/starkware-libs/sequencer/pull/14997) <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14997?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#14992** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14992?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14991** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14991?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#14973](https://github.com/starkware-libs/sequencer/pull/14973) <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14973?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#14990** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14990?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14989** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14989?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14983** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14983?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14982** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14982?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#14981** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14981?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#14948](https://github.com/starkware-libs/sequencer/pull/14948) <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14948?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* **#14998** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/14998?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main-v0.14.3`


<h2></h2>

asaf-sw:
Migrated review comment from #14944, closed in favor of this split.

@matanl-starkware wrote on `crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs` line 118, on `ChainlinkRateKind::EthToFri` ([original thread](https://reviewable.io/reviews/starkware-libs/sequencer/14944#-Ozw-zRfCMYyQWGPPve7)):

> It's weird to call this enum "ChainLink" when the mode is "Http"; it actually refers to Pragma.

This PR owns the code the comment lands on. The enum is gone, replaced by the `RateKind` / `ChainlinkRate` traits and their `EthToFri` and `StrkUsd` types, and the factory here is generic over the rate kind for both sources, so the `ChainlinkRate` bound is what names an HTTP feed's kind.

asaf-sw:
Migrated review comment from #14944, closed in favor of this split.

@matanl-starkware wrote on `crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs` line 270, on `config: &L1GasPriceProviderConfig,` ([original thread](https://reviewable.io/reviews/starkware-libs/sequencer/14944#-Ozw0Mc15sG1HCUToOBS)):

> Pass the relevant struct from the call site (after splitting as suggested before)

This PR owns that signature. `build_exchange_rate_oracle_client` still takes `&L1GasPriceProviderConfig` plus the optional batcher client, with the per-feed fields selected inside by the rate kind through `SourceSelectableFeed` rather than passed positionally. The "splitting as suggested before" it refers to is #14983 and #14982, nested by #14994.

asaf-sw:
Migrated review comment from #14944, closed in favor of this split.

@matanl-starkware wrote on `crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs` line 279 ([original thread](https://reviewable.io/reviews/starkware-libs/sequencer/14944#-Ozw0eXs-olOU_EIVAL0)), on the `ChainlinkRateKind::EthToFri` arm that selected `config.eth_to_strk_oracle_source`, `&config.eth_to_strk_oracle_config` and `ETH_TO_STRK_ORACLE_SOURCE_CONFIG_KEY`:

> To calc the Eth/STRK ratio, the code must also read the STRK/USD from the ChainLink contract.
> IMO we should either take both from Http or both from ChainLink. The separation just adds more room for bugs.

This PR owns the per-feed source selection the comment lands on: `new_with_oracle` builds each feed's client from its own configured source, so the two feeds can sit on different sources. The derived ETH/STRK rate reading both legs off Chainlink, and its freshness across those legs, is #14993, where the same comment is copied.

matanl-starkware (changes requested):
> @matanl-starkware made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14996)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on asaf-sw).

___
*[`crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs` line 73 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14996#-P-hfCsB9Tu7tjM0UnWI:-P-hfCsB9Tu7tjM0UnWJ:b-89qs2y) ([raw file](https://github.com/starkware-libs/sequencer/blob/a0a3b49f369b38f82d3506ba2a6ce39c3dd4ccf9/crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs#L73)):*
> ```XML
> // [Temporary comment] Deleted in C1 together with the two config params they name.
> const ETH_TO_STRK_ORACLE_SOURCE_CONFIG_KEY: &str =
>     "l1_gas_price_provider_config.eth_to_strk_oracle_source";
> ```

Is it the same string that appears in the KEYS_TO_BE_REPLACED?
Can you make it a single source of truth?



_Code quote:_
```XML
"l1_gas_price_provider_config.eth_to_strk_oracle_source"
```
___
*[`crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs` line 91 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14996#-P-hgNuaDMGvEb8IL8IM:-P-hgNuaDMGvEb8IL8IN:b7gk73b) ([raw file](https://github.com/starkware-libs/sequencer/blob/a0a3b49f369b38f82d3506ba2a6ce39c3dd4ccf9/crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs#L91)):*
> ```XML
>     // Every feed that selects `Chainlink`, so that one startup reports the whole misconfiguration.
>     source_config_keys: Vec<&'static str>,
> }
> ```

Seems like an overkill to me. It's OK if the Operator gets 2 different errors.



_Code quote:_
```XML
/// Whether a batcher client is available is decided by the composition root, not by config, so this
/// is raised while the components are built rather than caught by config validation.
// [Temporary comment] Deleted in C1: once Chainlink is the only source, a provider without a
// batcher client cannot serve a rate at all, so there is no source selection left to reject.
#[derive(Debug, Eq, Error, PartialEq)]
#[error(
    "Chainlink is selected as the oracle source by {}, but this service has no batcher client. \
     Chainlink feeds are read through the batcher, so either run the batcher alongside this \
     service or select Http instead.",
    .source_config_keys.join(" and ")
)]
pub struct MissingBatcherClientError {
    // Every feed that selects `Chainlink`, so that one startup reports the whole misconfiguration.
    source_config_keys: Vec<&'static str>,
}
```
___
*[`crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs` line 123 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14996#-P-hgadf1WOnkkU68Sid:-P-hgadg-dycB5JrUzET:bvp333f) ([raw file](https://github.com/starkware-libs/sequencer/blob/a0a3b49f369b38f82d3506ba2a6ce39c3dd4ccf9/crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs#L123)):*
> ```XML
>         batcher_client: Option<SharedBatcherClient>,
>     ) -> Result<Self, MissingBatcherClientError> {
>         // Both feeds are resolved before either failure is reported, so an operator who
> ```

I don't really think it's necessary.
Why do we even allow the batcher_client to be None under the "new_with_oracle" function?
This flow has no meaning without a client.



_Code quote:_
```XML
Both feeds are resolved before either failure is reported
```

asaf-sw (commented):
> @asaf-sw+AGNT made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14996)* status: 0 of 7 files reviewed, 3 unresolved discussions (waiting on matanl-starkware).

___
*[`crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs` line 73 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14996#-P-hfCsB9Tu7tjM0UnWI:-P-hl7MG0f9ia32-KlVU:b-carbkw) ([raw file](https://github.com/starkware-libs/sequencer/blob/a0a3b49f369b38f82d3506ba2a6ce39c3dd4ccf9/crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs#L73)):*

Is it the same string that appears in the KEYS_TO_BE_REPLACED?
Can you make it a single source of truth?
</blockquote></details>

Done, and all three of these collapse into one deletion.

`new_with_oracle` now requires the batcher client and is infallible:

```rust
pub fn new_with_oracle(
    config: L1GasPriceProviderConfig,
    batcher_client: SharedBatcherClient,
) -> Self
```

`components.rs` does the `expect` at the composition root. `MissingBatcherClientError` is gone, and with it the resolve-both-feeds logic, so there is no second error for an operator to receive either.

The two `*_ORACLE_SOURCE_CONFIG_KEY` consts existed only to name the offending key in that error message, so they are deleted too. That removes the string duplicated against `KEYS_TO_BE_REPLACED` rather than trying to share it, which was not reachable anyway: `apollo_deployments` has no dependency on `apollo_l1_gas_price`, and its `phf_set!` holds plain literals for every component key.

Net -111 lines, 98 tests pass in the crate.
___
*[`crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs` line 91 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14996#-P-hgNuaDMGvEb8IL8IM:-P-hlB2k3EbU5uiCn0_E:b-qgn7o3) ([raw file](https://github.com/starkware-libs/sequencer/blob/a0a3b49f369b38f82d3506ba2a6ce39c3dd4ccf9/crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs#L91)):*

Seems like an overkill to me. It's OK if the Operator gets 2 different errors.
</blockquote></details>

Agreed, and it is gone entirely rather than reduced to two errors. `new_with_oracle` requires the batcher client now, so there is no failure to report: see the reply on the config-key thread for the shape.
___
*[`crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs` line 123 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/14996#-P-hgadf1WOnkkU68Sid:-P-hlBLUAOFyQyq1XtAi:b-9r6dm9) ([raw file](https://github.com/starkware-libs/sequencer/blob/a0a3b49f369b38f82d3506ba2a6ce39c3dd4ccf9/crates/apollo_l1_gas_price/src/l1_gas_price_provider.rs#L123)):*

I don't really think it's necessary.
Why do we even allow the batcher_client to be None under the "new_with_oracle" function?
This flow has no meaning without a client.
</blockquote></details>

You are right, there is no meaning to this flow without a client. `batcher_client` is now `SharedBatcherClient` rather than `Option`, and `components.rs` resolves the `Option` from `get_batcher_shared_client()` with an `expect` at the composition root.

Checked before removing it: `get_batcher_shared_client()` returns `None` only for `Client::Disabled`, and every shipped topology gives the L1 service a batcher client, including distributed, where `l1.json` has `components.batcher.execution_mode = Remote`.

This also deleted the two tests that only existed to cover the removed error.

matanl-starkware (approved):
> @matanl-starkware reviewed 7 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14996)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on asaf-sw).

asaf-sw (commented):
> @asaf-sw reviewed 7 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14996)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on asaf-sw).

matanl-starkware (approved):
:lgtm:

> @matanl-starkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/14996#-:-P-iiBkG8se1bMwnpaaO:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on asaf-sw).

gkaempfer:

---

PR #15000, opened 2026-08-19
starknet_patricia_storage: split RocksDB multi-key reads across blocking tasks

ron-starkware:
## Problem

`RocksDbStorage::mget` calls `db.multi_get`, which batches at the API level but issues the underlying file reads **one at a time**. A large batch therefore runs at queue depth ~1 and costs `n_keys * read_latency`, no matter how many IOPS the device can serve.

This dominates the committer's pre-commit Patricia witness fetch, which is `tokio::join!`ed with the compute phase — so on state-heavy blocks the whole commit is gated on it, the batcher reports "not ready", and consensus stalls for tens of seconds.

Measured on a committer replaying mainnet traffic (block with 101,639 witness entries):

```
Block stats: durations in ms (total/read/compute/write): 58641/2578/154/171,
  witness fetch ms (pre-commit/post-commit): 50883/167, witness entries: 101639
Component Committer took 58641 ms to process "CommitterRequest::ReadPathsAndCommitBlock"
```

101,639 reads in 50.9s = **2,008 reads/sec at 0.65ms each** — that is hyperdisk's single-read latency, i.e. serial round-trips. At the time this used **12% of the disk's provisioned IOPS and 2% of its bandwidth**, on a node at 25% CPU. Disk-level sampling confirmed **queue depth 1.30**.

## Fix

Split the batch across blocking tasks so several reads are in flight. All tasks are spawned before any is awaited, so they run concurrently; awaiting them in order keeps values aligned with the requested keys.

`max_read_tasks` is configurable and `1` restores the previous behaviour exactly, so it can be A/B'd or reverted without a rebuild.

## Results

Same blocks, replayed deterministically — witness-entry counts byte-identical before and after, so this is the same work done faster:

| block | entries | witness fetch before → after | speedup |
|---|---|---|---|
| 11752174 | 101,639 | 50,883 → 2,699 ms | **18.9x** |
| 11752207 | 135,879 | 29,364 → 2,101 ms | **14.0x** |
| 11752175 | 111,824 | 18,374 → 1,695 ms | **10.8x** |
| 11752177 | 82,655 | 9,031 → 1,211 ms | 7.5x |
| 11752176 | 68,349 | 7,738 → 1,059 ms | 7.3x |

Commit-duration distribution over ~320 blocks:

| | before | after |
|---|---|---|
| p50 | 87 ms | 86 ms |
| p99 | 5,792 ms | **1,920 ms** |
| max | **58,641 ms** | **3,696 ms** |
| commits > 10s | 4 | **0** |

p50 is unchanged, as expected — light blocks were never affected.

Mechanism confirmed by disk sampling: **queue depth 1.30 → 24.4**, peak read IOPS 3,612 → 45,140, **read latency unchanged** (0.650 → 0.632 ms). Same disk, same reads, just issued concurrently.

Note the run that produced these numbers also raised the disk from 30k to 60k provisioned IOPS. That cannot explain the gain on its own — the baseline used only 12% of the old ceiling — but peak demand did reach 45,140 IOPS, so the extra headroom is needed to realise the full benefit.

## Notes for review

- `MIN_KEYS_PER_READ_TASK` caps the split as well as flooring it: a batch only reaches `max_read_tasks` tasks once it holds `max_read_tasks * MIN_KEYS_PER_READ_TASK` keys. For a typical ~400-key level that yields 13 tasks regardless of whether `max_read_tasks` is 16 or 32, which is why measured queue depth peaked at 24.4 rather than 32. Happy to tune either constant.
- The split is **per `mget` call**, not global. `fetch_contract_storage_paths_concurrently` already fetches several tries at once, so peak blocking-task demand is (concurrent tries x tasks). It is bounded only by tokio's blocking pool, which queues rather than fails. A global semaphore would be the stricter alternative if reviewers prefer.
- I considered RocksDB's `multi_get_opt` with `ReadOptions::set_async_io`, which the crate exposes; it only parallelises when the build has io_uring, which I could not verify, so this approach does not depend on it.
- **Follow-up, not fixed here:** `MdbxStorage::mget` has the same defect — a `for` loop of `txn.get()`. Its single-RO-transaction/mmap model makes parallelising it a different design. `AerospikeStorage` is fine (real batch API).

## Testing

`cargo test -p starknet_patricia_storage --features rocksdb_storage,mdbx_storage` — 10/10 pass.

New test asserts values stay positionally aligned across chunk boundaries, at `max_read_tasks` of both 32 and 1. Only even keys are written, so a misaligned chunk surfaces immediately as a `None` in the wrong slot.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13145,6 +13145,7 @@ dependencies = [
  "serde_json",
  "starknet-types-core",
  "starknet_api",
+ "tempfile",
  "thiserror 1.0.69",
  "tokio",
  "validator",
--- a/crates/apollo_deployments/resources/app_configs/committer_config.json
+++ b/crates/apollo_deployments/resources/app_configs/committer_config.json
@@ -12,6 +12,7 @@
     "committer_config.storage_config.inner_storage_config.max_background_jobs": 8,
     "committer_config.storage_config.inner_storage_config.use_mmap_reads": false,
     "committer_config.storage_config.inner_storage_config.spawn_blocking_reads": true,
+    "committer_config.storage_config.inner_storage_config.max_read_tasks": 32,
     "committer_config.storage_config.inner_storage_config.max_subcompactions": 8,
     "committer_config.storage_config.inner_storage_config.max_write_buffers": 4,
     "committer_config.storage_config.inner_storage_config.num_threads": 8,
--- a/crates/apollo_deployments/resources/app_configs/replacer_committer_config.json
+++ b/crates/apollo_deployments/resources/app_configs/replacer_committer_config.json
@@ -14,6 +14,7 @@
   "committer_config.storage_config.inner_storage_config.max_write_buffers": 4,
   "committer_config.storage_config.inner_storage_config.num_threads": 8,
   "committer_config.storage_config.inner_storage_config.spawn_blocking_reads": true,
+  "committer_config.storage_config.inner_storage_config.max_read_tasks": 32,
   "committer_config.storage_config.inner_storage_config.use_mmap_reads": false,
   "committer_config.storage_config.inner_storage_config.write_buffer_size": 134217728,
   "committer_config.verify_state_diff_hash": "$$$_COMMITTER_CONFIG-VERIFY_STATE_DIFF_HASH_$$$"
--- a/crates/apollo_node/resources/config_schema.json
+++ b/crates/apollo_node/resources/config_schema.json
@@ -604,6 +604,11 @@
     "privacy": "Public",
     "value": 8
   },
+  "committer_config.storage_config.inner_storage_config.max_read_tasks": {
+    "description": "Max parallel disk reads per multi-key batch (disk queue depth). Higher trades provisioned IOPS for faster large batches; 1 disables splitting. Small batches use fewer tasks.",
+    "privacy": "Public",
+    "value": 32
+  },
   "committer_config.storage_config.inner_storage_config.max_subcompactions": {
     "description": "Maximum number of subcompactions for parallel compaction",
     "privacy": "Public",
--- a/crates/starknet_patricia_storage/Cargo.toml
+++ b/crates/starknet_patricia_storage/Cargo.toml
@@ -41,4 +41,5 @@ libmdbx.workspace = true
 page_size.workspace = true
 rstest.workspace = true
 rust-rocksdb.workspace = true
+tempfile.workspace = true
 tokio = { workspace = true, features = ["rt-multi-thread"] }
--- a/crates/starknet_patricia_storage/src/rocksdb_storage.rs
+++ b/crates/starknet_patricia_storage/src/rocksdb_storage.rs
@@ -1,5 +1,6 @@
 use core::fmt;
 use std::collections::BTreeMap;
+use std::num::NonZeroUsize;
 use std::path::Path;
 use std::sync::Arc;
 
@@ -58,6 +59,14 @@ const MAX_WRITE_BUFFERS: i32 = 4;
 const NUM_THREADS: i32 = 8;
 // Maximum number of background compactions (STT files merge and rewrite) and flushes.
 const MAX_BACKGROUND_JOBS: i32 = 8;
+// `multi_get` issues its reads one at a time, so an unsplit batch is bound by read latency rather
+// than by IOPS. Raising this only shifts the wait into the disk and blocking-pool queues; lowering
+// it leaves IOPS idle on large batches. Binds only above TARGET_KEYS_PER_READ_TASK keys per task.
+const MAX_READ_TASKS: usize = 32;
+// Keys per task. Each task reads its chunk serially, so wall time grows with this times the
+// per-read latency; lower buys parallelism, higher only amortizes the task spawn, which is
+// negligible beside a disk read. Chunks are balanced, so the smallest is about half this.
+const TARGET_KEYS_PER_READ_TASK: usize = 32;
 
 pub struct RocksDbOptions {
     pub db_options: Options,
@@ -156,6 +165,9 @@ pub struct RocksDbStorageConfig {
     pub use_mmap_reads: bool,
     /// Whether to spawn blocking tasks for read operations.
     pub spawn_blocking_reads: bool,
+    /// Max parallel disk reads per multi-key batch (disk queue depth). Higher trades provisioned
+    /// IOPS for faster large batches; 1 disables splitting. Small batches use fewer tasks.
+    pub max_read_tasks: NonZeroUsize,
 }
 
 impl Default for RocksDbStorageConfig {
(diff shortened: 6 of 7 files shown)


itamar-starkware (changes requested):
Looks good.

> @itamar-starkware reviewed 7 files and all commit messages, and made 4 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15000#-:-P-PYgxX3E6QCPoos6lh:b3kn4ww)* status: all files reviewed, 3 unresolved discussions (waiting on matanl-starkware and ron-starkware).

___
*[`crates/apollo_node/resources/config_schema.json` line 608 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/15000#-P-PS61fERz24CUumMoz:-P-PS61fERz24CUumMp-:b-vg33mg) ([raw file](https://github.com/starkware-libs/sequencer/blob/b62ff28752109426c16ba77c29a4539c710052cc/crates/apollo_node/resources/config_schema.json#L608)):*
> ```JSON
>   },
>   "committer_config.storage_config.inner_storage_config.max_read_tasks": {
>     "description": "Upper bound on the blocking tasks a multi-key read is split across; smaller batches use fewer",
> ```

Suggestion:
```
"Max parallel disk reads per multi-key batch (disk queue depth). Higher trades provisioned \ 
 IOPS for faster large batches; 1 disables splitting (previous behavior). Small batches use \
 fewer tasks.",                                                                              
```
___
*[`crates/starknet_patricia_storage/src/rocksdb_storage.rs` line 63 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/15000#-P-PS1dk56ixyC91yuMZ:-P-PS1dk56ixyC91yuM_:b-eepn5f) ([raw file](https://github.com/starkware-libs/sequencer/blob/b62ff28752109426c16ba77c29a4539c710052cc/crates/starknet_patricia_storage/src/rocksdb_storage.rs#L63)):*
> ```XML
> const MAX_BACKGROUND_JOBS: i32 = 8;
> // `multi_get` issues its reads one at a time, so an unsplit batch is bound by read latency.
> const MAX_READ_TASKS: usize = 32;
> ```

Suggestion:
```
// Cap on parallel reads in flight per mget. Sized so a single batch can approach the disk's 
// provisioned IOPS; higher only moves the wait into the disk and tokio-blocking-pool queues,
// lower leaves provisioned IOPS idle on large batches. Only binds for batches holding more  
// than TARGET_KEYS_PER_READ_TASK keys per task.                                             
```
___
*[`crates/starknet_patricia_storage/src/rocksdb_storage.rs` line 66 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/15000#-P-PRf0o-8_bakQEYdIe:-P-PRf0o-8_bakQEYdIf:b-nbbmk9) ([raw file](https://github.com/starkware-libs/sequencer/blob/b62ff28752109426c16ba77c29a4539c710052cc/crates/starknet_patricia_storage/src/rocksdb_storage.rs#L66)):*
> ```XML
> // Chunk size at which a task hop pays for itself. Chunks are balanced, so the smallest can be
> // about half this. Also caps the split: MAX_READ_TASKS tasks need that many times this.
> const TARGET_KEYS_PER_READ_TASK: usize = 32;
> ```

Suggestion:
```
// Keys per task; each task reads its chunk serially, so a batch's wall time grows with this  
// times the per-read latency. Lower = more parallelism for a typical trie level; higher only 
// amortizes the task-spawn overhead, which is negligible next to a disk read and matters just
// for cache-hot keys. Chunks are balanced, so the smallest can be about half this.           
```

ron-starkware (commented):
> @ron-starkware made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15000)* status: 5 of 7 files reviewed, 3 unresolved discussions (waiting on itamar-starkware and matanl-starkware).

___
*[`crates/apollo_node/resources/config_schema.json` line 608 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/15000#-P-PS61fERz24CUumMoz:-P-P_jdX9swFlFIEMGA0:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/b62ff28752109426c16ba77c29a4539c710052cc/crates/apollo_node/resources/config_schema.json#L608)):*
<details><summary><i>Previously, itamar-starkware wrote…</i></summary><blockquote>

Suggestion:
```
"Max parallel disk reads per multi-key batch (disk queue depth). Higher trades provisioned \ 
 IOPS for faster large batches; 1 disables splitting (previous behavior). Small batches use \
 fewer tasks.",                                                                              
```
</blockquote></details>

Done
___
*[`crates/starknet_patricia_storage/src/rocksdb_storage.rs` line 63 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/15000#-P-PS1dk56ixyC91yuMZ:-P-P_kGx1YNLNzJ6fdBu:b3ywf1k) ([raw file](https://github.com/starkware-libs/sequencer/blob/b62ff28752109426c16ba77c29a4539c710052cc/crates/starknet_patricia_storage/src/rocksdb_storage.rs#L63)):*
<details><summary><i>Previously, itamar-starkware wrote…</i></summary><blockquote>

Suggestion:
```
// Cap on parallel reads in flight per mget. Sized so a single batch can approach the disk's 
// provisioned IOPS; higher only moves the wait into the disk and tokio-blocking-pool queues,
// lower leaves provisioned IOPS idle on large batches. Only binds for batches holding more  
// than TARGET_KEYS_PER_READ_TASK keys per task.                                             
```
</blockquote></details>

Done (with rewrite)
___
*[`crates/starknet_patricia_storage/src/rocksdb_storage.rs` line 66 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/15000#-P-PRf0o-8_bakQEYdIe:-P-Pa87o8n7BxK9VX8xN:bb6un1j) ([raw file](https://github.com/starkware-libs/sequencer/blob/b62ff28752109426c16ba77c29a4539c710052cc/crates/starknet_patricia_storage/src/rocksdb_storage.rs#L66)):*
<details><summary><i>Previously, itamar-starkware wrote…</i></summary><blockquote>

Suggestion:
```
// Keys per task; each task reads its chunk serially, so a batch's wall time grows with this  
// times the per-read latency. Lower = more parallelism for a typical trie level; higher only 
// amortizes the task-spawn overhead, which is negligible next to a disk read and matters just
// for cache-hot keys. Chunks are balanced, so the smallest can be about half this.           
```
</blockquote></details>

Done

itamar-starkware (approved):
> @itamar-starkware reviewed 2 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15000)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

gkaempfer:

---

PR #15001, opened 2026-08-19
echonet: serve the recorder's witness height offset

ron-starkware:
## Problem

Main added `verify_retrospective_state_commitment_infos` (#14895), which calls `GET /cende_recorder/get_witness_height_offset` before every proposal. Echonet's mock recorder never implemented it, so the sequencer got a 404 and **every consensus round failed**:

```
PROPOSAL_FAILED: Proposal failed as proposer. Error: RetrospectiveStateCommitmentInfosError(
  CendeError(RecorderRequestFailed { path: "/cende_recorder/get_witness_height_offset",
  message: "returned error status 404 Not Found" }))
```

The sequencer already handles a recorder with nothing stored — it skips the check when the offset is null:

```rust
if cende_recorder_height_offset.is_none() {
    warn!("...the cende recorder hasn't stored any; skipping the retrospective ... validation.");
    return Ok(());
}
```

It just never got there: a 404 is a transport error and fails before that branch.

## Fix

Serve the offset from the existing commit store — one past the last contiguously-stored height, or `null` when nothing is stored. Wired to `SharedContext.get_last_stored_commitment_height`, so it stays meaningful as blobs accumulate rather than being hardcoded to `null`.

## Note

`/cende_recorder/get_last_stored_commitment_height` is left in place even though current main no longer calls it, because the previous sequencer image still does — removing it would break rollback to that image.

### DIFF
--- a/echonet/echo_center.py
+++ b/echonet/echo_center.py
@@ -1167,6 +1167,17 @@ class EchoCenterService:
         last = self.shared.get_last_stored_commitment_height()
         return self._json_response({"block_number": last}, requests.codes.ok)
 
+    def handle_get_witness_height_offset(self) -> flask.Response:
+        """
+        GET /cende_recorder/get_witness_height_offset
+
+        Returns one past the last contiguously-stored commitment-info height (or null);
+        the sequencer skips its retrospective check entirely on null.
+        """
+        last = self.shared.get_last_stored_commitment_height()
+        offset = None if last is None else last + 1
+        return self._json_response({"block_number": offset}, requests.codes.ok)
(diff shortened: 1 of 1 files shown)


matanl-starkware (approved):
> @matanl-starkware reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15001)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ron-starkware).

gkaempfer:

---

PR #15006, opened 2026-08-22
apollo_batcher,starknet_api: source view call block info from the committed block hash components

matanl-starkware:
`Batcher::call_contract` cannot serve any height above genesis. `get_block_info` reads the `headers` table, which nothing in the batcher writes, so every non-genesis call fails with a bare `InternalError`.

The two production header writers (`apollo_central_sync`, `apollo_p2p_sync`) write `/data/state_sync`, not `/data/batcher`; `BatcherStorageWriter` has no header method at all. The bug was invisible because `test_utils.rs` mocked `get_block_header` to succeed for any height, and the real-storage test seeded a header itself — a precondition the node never has.

This gates the Chainlink oracle flip, which reads a feed through `call_contract`, and it applies equally to the other consumer, `CairoStakingContract`.

## Fix

Source the block info from `PartialBlockHashComponents`, which every `commit_proposal` already writes and which carries every `BlockInfo` field. No new write, no storage-scope change, no fabricated block hash, and no backfill — only `height.prev()` is ever read, so it works from the first commit after the upgrade.

`use_kzg_da` is not stored directly, but it is recoverable: `l1_da_mode` is packed into the top bit of byte 24 of `concatenated_counts`, so `extract_l1_da_mode_from_concatenated_counts` (added beside the existing `extract_event_count_from_concatenated_counts`) recovers it. It is not defaulted, because the mode is committed to by the block hash — `concatenated_counts` is chained into `calculate_block_hash`, and the commit path runs the same value in reverse via `L1DataAvailabilityMode::from_use_kzg_da`. Assuming it would put wrong data in a `BlockInfo` that a future caller could feed back into a hash.

Two smaller corrections on the same path:

- A committed zero gas price is floored to `NonzeroGasPrice::MIN` rather than failing the call, matching the RPC view-call path (`apollo_rpc_execution/src/lib.rs:382-397`). Both test helpers previously set `GasPrice(1)` specifically to dodge this.
- A block predating block hash components now returns `ContractCallFailed` naming what is missing, instead of a bare `InternalError`.

`get_block_info` is now `pub(crate)` so the sibling test module can assert on it directly — the view call cannot observe the DA mode end to end, since a fee-free call never reaches the cost table. `validate_retdata_length` in the same file is already `pub(crate)` for the same reason.

## Tests

The regression guard is the removal of `append_header` from `deploy_contract_in_real_storage`. Dropping that one call — a precondition production never satisfies — makes both real-storage `call_contract` tests fail with `InternalError`, the exact production symptom, before the fix.

Added: a committed zero gas price case, a missing-record case asserting the error names the problem, a `get_block_info` case per DA mode, and a round-trip test for the extractor with `state_diff_length` maxed out to catch a bit-offset error.

Verified the DA mode test discriminates rather than merely passes: with `use_kzg_da: false` put back, the blob case fails `left: false, right: true` while calldata still passes.

## Verification

- `apollo_batcher` 136/136, `starknet_api` 98/98
- `cargo clippy -p starknet_api --all-targets` under `-D warnings` clean
- `scripts/run_tests.py --command clippy --changes_only` clean
- `cargo build -p apollo_node` succeeds

No integration test exercises `Batcher::call_contract`. The integration harness (`apollo_integration_tests/src/state_reader.rs:489,497`) writes both a header and the block hash components for block 0, so it is compatible either way.

### DIFF
--- a/crates/apollo_batcher/src/batcher.rs
+++ b/crates/apollo_batcher/src/batcher.rs
@@ -55,7 +55,6 @@ use apollo_storage::global_root_marker::{
     GlobalRootMarkerStorageReader,
     GlobalRootMarkerStorageWriter,
 };
-use apollo_storage::header::HeaderStorageReader;
 use apollo_storage::metrics::BATCHER_STORAGE_OPEN_READ_TRANSACTIONS;
 use apollo_storage::partial_block_hash::{
     PartialBlockHashComponentsStorageReader,
@@ -100,7 +99,6 @@ use indexmap::{IndexMap, IndexSet};
 use mockall::automock;
 use starknet_api::block::{
     BlockHash,
-    BlockHeaderWithoutHash,
     BlockInfo,
     BlockNumber,
     GasPrice,
@@ -110,6 +108,7 @@ use starknet_api::block::{
     UnixTimestamp,
 };
 use starknet_api::block_hash::block_hash_calculator::{
+    extract_l1_da_mode_from_concatenated_counts,
     PartialBlockHash,
     PartialBlockHashComponents,
 };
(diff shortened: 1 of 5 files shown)


dafnamatsry (approved):
> @dafnamatsry reviewed 5 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15006)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on asaf-sw).

gkaempfer:

---

PR #15010, opened 2026-08-24
apollo_gateway: validate replayed txs against their source block's gas prices

ron-starkware:
## Problem

`validate_tx_l2_gas_price_within_threshold` compares a tx's `l2 max_price_per_unit` against `min_gas_price_percentage` of the **committed tip's** L2 gas price. Under replay the tip trails the block the tx came from, and mainnet's L2 price moves up to ~2% per block, so a tx mainnet accepted gets rejected here for having less headroom than the price fell in between.

That is a **deadlock, not a delay**. A tx in block N clears the threshold only once the tip's price drops to its cap, which in a falling market first happens at `tip = N` — and block N is closed by then, so the tx can never enter its own block. It is deterministic, so the retry after a resync reproduces it exactly and Echonet ends up permanently skipping a mainnet block.

Block **11926844** hit it twice this weekend:

```
tx      0x1549342b9b52c7d2046e604d920fd788a52afb54240ef0f904667186dbb9b3f
l2 cap  32,000,000,000        (a round bot cap)
block   11926844, price 31,507,218,957   -> only +1.56% headroom
```

Mainnet's price was in a five-block slide into that block (34.85e9 → 31.51e9), and 57 `GAS_PRICE_TOO_LOW` rejections walked the threshold from 32,093,567,243 (block 11926843) up to 34,854,264,801 (11926839) without ever dropping under the cap. Echonet built 11926844 with **1 of its 4 txs**, produced the byte-identical wrong hash on the retry, and skipped the block. Block 11802770 four days earlier was the same mechanism.

## Change

In `BehaviorMode::Echonet`, threshold against the L2 gas price of the block the tx **actually came from**, resolved from the recorder.

A tx mainnet included in block N necessarily paid `price(N)`, so `price(N) <= cap` always holds and the tx clears on its **first** forward, regardless of where the tip is. The check keeps its teeth: anything that could not have paid its own block's price is still rejected. Outside Echonet mode nothing changes.

Why not the alternatives:

- **`validate_resource_bounds: false`** removes the check entirely.
- **Lowering `min_gas_price_percentage`** weakens it for everything, and no fixed percentage is safe — the required discount depends on how far the price falls.
- **Thresholding on `next_l2_gas_price`** (which is what the `TODO(Arni)` at that line wants, and is a genuine latent improvement) is not enough on its own: it moves the pass condition from `tip = N` to `tip = N-1`, which is only reached moments before block N is built, leaving a ~1 s race against the gateway's state-sync lag. Worth doing separately, but not as the fix for this.

## Notes

- Both recorder endpoints already exist and are already relied on by the mempool in this mode (`apollo_mempool/src/communication.rs`), so **no recorder-side change is needed** — `echonet/get_block_metadata` already returns `l2_gas_price_fri`.
- Any lookup failure — endpoint down, unparsable body, zero price — logs and falls back to the committed tip's price rather than failing the tx.
- Costs two recorder round trips per tx on the validation path, both in-cluster. At Echonet's ~3 tx/s sustained that is negligible; the 2 s timeout matches the mempool's.

## Testing

`crates/apollo_gateway/src/source_block_gas_price_test.rs`, 5 cases against a `mockito` recorder: resolves the real block-11926844 price from the exact payload the Python handler emits, and falls back on missing tx metadata, missing block metadata, an unparsable body, and a zero price.

`cargo test -p apollo_gateway` — 170 passed, 0 failed. `cargo clippy -p apollo_gateway --all-targets -- -D warnings` clean. `scripts/rust_fmt.sh` clean.


ron-starkware:
Reworked after a code review found the first attempt did not achieve its goal. Pushed as a second commit rather than a force-push so the review's line anchors survive; the net diff against `main` is what matters and no longer contains the old module.

**What was wrong:** overriding only the explicit `GAS_PRICE_TOO_LOW` threshold left blockifier's `check_fee_bounds` (via `run_validate_entry_point`) still comparing the bid against the committed block's prices from the same `get_block_info()`. Its bound is the stricter of the two — `max_price_per_unit >= actual_gas_price`, no percentage — so the deadlocked tx was still rejected, just as a more confusing `ValidateFailure`. And it covers all three resources, so an l2-only override was too narrow regardless: seven of the nine `Resource bounds were not satisfied` rejections echonet logged over four days were **l1_data_gas**.

**What changed:** the source block's whole STRK price vector is now resolved once per tx and applied in a single `block_info()` seam that both the explicit check and blockifier's `BlockContext` are built from. The threshold helper and the per-call client plumbing are gone.

Also addressed from the review: resolution moved ahead of the state-snapshot pin; retry with the mempool's policy; a last-block memo so the block lookup is ~1 per block; a u64 sanity bound on recorder-supplied prices; `Url::join` base normalization; the stale `previous_block_l2_gas_price` parameter name; and a validator-level test on the replay branch, which is now trivial because the validator holds a plain `Option<GasPriceVector>` instead of an HTTP client.

Two review items deliberately left as follow-ups, noted in the description: no metric on the recorder dependency (needs a `define_metrics!` entry plus a dashboard regen), and this being a third hand-rolled copy of the recorder client (a cross-crate extraction).

One partial disagreement on the fixture finding: the previous test suite *did* cover tx-metadata-OK plus block-metadata-absent. It's right that 500 and unparsable on the second leg were missing and that the rstest parameter was dead — both fixed, along with the hex/decimal duplication and the hardcoded block number.

matanl-starkware (changes requested):
> @matanl-starkware reviewed 9 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15010)* status: all files reviewed, 2 unresolved discussions (waiting on ron-starkware).

___
*[`crates/apollo_gateway/src/replay_gas_prices.rs` line 40 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/15010#-P2IQwtB2eqTv0hr3dKL:-P2IQwtB2eqTv0hr3dKM:b45h) ([raw file](https://github.com/starkware-libs/sequencer/blob/af23ca47e10f44bc86774a209f28a92956e106aa/crates/apollo_gateway/src/replay_gas_prices.rs#L40)):*
> ```XML
>     client: ClientWithMiddleware,
>     // Immutable per block, and consecutive replayed txs share one, so a single slot suffices.
>     last_resolved: Mutex<Option<(BlockNumber, GasPriceVector)>>,
> ```





_Suggestion:_
```XML
last_resolved_block_prices
```
___
*[`crates/apollo_gateway/src/replay_gas_prices.rs` line 131 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/15010#-P2IMZI0EdGFNCRS0BDn:-P2IMZI0EdGFNCRS0BDo:bbycz5f) ([raw file](https://github.com/starkware-libs/sequencer/blob/af23ca47e10f44bc86774a209f28a92956e106aa/crates/apollo_gateway/src/replay_gas_prices.rs#L131)):*
> ```XML
>             .await
>             .map_err(|parse_error| format!("invalid response: {parse_error}"))
>     }
> ```

Looks very similar to the code of the mempool.
Consider putting is some common place and reuse.



_Code quote:_
```XML
    async fn try_fetch_json<ResponseBody: DeserializeOwned>(
        &self,
        path_and_query: &str,
    ) -> Result<ResponseBody, String> {
        let url = self
            .recorder_url
            .join(path_and_query)
            .map_err(|join_error| format!("invalid recorder URL: {join_error}"))?;
        let response = self
            .client
            .get(url.as_str())
            .timeout(REQUEST_TIMEOUT)
            .send()
            .await
            .map_err(|request_error| format!("request failed: {request_error}"))?;

        if !response.status().is_success() {
            return Err(format!("HTTP {}", response.status()));
        }

        response
            .json::<ResponseBody>()
            .await
            .map_err(|parse_error| format!("invalid response: {parse_error}"))
    }
```

PR #14450, opened 2026-06-10
apollo_starknet_os_program: remove support for proof version 0

einat-starkware:

PR #14432, opened 2026-06-10
starknet_os,starknet_api,starknet_proof_verifier: remove support for proof version 0

einat-starkware:

PR #15012, opened 2026-08-24
starknet_api,starknet_os,starknet_proof_verifier: move to proof version V2

einat-starkware:
## What

Moves client-side proving from proof version **V1** to **V2**, and switches the proof verifier to the V2 circuit. Mirrors the V0 → V1 migration (#14432 + #14450), including its "one supported version at a time" policy: only V2 is verifiable, older markers are retained purely so historical blocks stay replayable.

> [!IMPORTANT]
> **Draft — do not merge.** The V2 `privacy-circuit-verify` / `privacy-prove` release does not exist yet, so the workspace does not resolve and several derived artifacts cannot be regenerated. See [Blocked on the V2 verifier](#blocked-on-the-v2-verifier).

> [!NOTE]
> Opened against `main` because there is no `main-v0.14.4` release branch yet (release branches currently stop at `main-v0.14.3`, which is cut and carries V1). **Once `main-v0.14.4` is cut, this PR should be retargeted to it.**

## Changes

**Cairo OS** — `PROOF_VERSION_V1` replaced by `PROOF_VERSION_V2 = 'PROOF2'` in `virtual_os_output.cairo`; `check_proof_facts` in `execution_constraints.cairo` now asserts `proof_version = PROOF_VERSION_V2`. This changes the `os` program hash only — `execution_constraints.cairo` is a virtual-OS-swapped file, so the `virtual_os` hash and `allowed_virtual_os_program_hashes` are untouched.

**`starknet_api`** — adds `PROOF_VERSION_V2` (`0x50524f4f4632`) and `ProofVersion::V2`. `V0` and `V1` are kept, documented as replay-only: they still appear in historical blocks (V0 under protocol 0.14.2, V1 under the released 0.14.3), and acceptance stays gated per protocol version by `allowed_proof_versions`.

**`starknet_os`** — `Const::ProofVersionV1` → `Const::ProofVersionV2`, with the matching Rust/Cairo constant equality test.

**`blockifier`** — `blockifier_versioned_constants_0_14_4.json` `allowed_proof_versions`: `[V1]` → `[V2]`, plus the regenerated `0.14.3_0.14.4.txt` diff regression. 0.14.4 is the in-flight version on `main`, so this is edited in place — exactly what the V0 → V1 PR did to the then-in-flight 0.14.3.

**`starknet_proof_verifier`** — workspace alias `privacy-circuit-verify-v1` → `privacy-circuit-verify-v2`; `verify_proof` dispatches V2 to the V2 circuit and rejects V0/V1 up front with `InvalidProofVersion`; `ProgramOutput::try_into_proof_facts` stamps V2.

**Fixtures** — proof-facts markers bumped to V2 in the cende central objects fixtures; `ProofFacts::custom_proof_facts_for_testing` and the batcher block-builder test now build V2 facts.

## Blocked on the V2 verifier

These are deliberately left stale and must be finished before this leaves draft:

- [ ] **`Cargo.toml`**: `privacy-circuit-verify-v2` and `privacy-prove` point at `tag = "v0.14.4"`, which does not exist on `proving-utils` yet. Until it is pushed, `cargo metadata` fails and nothing in the workspace builds.
- [ ] **`Cargo.lock`**: cannot be regenerated while the tag is missing; left untouched.
- [ ] **`crates/apollo_starknet_os_program/src/program_hash.json`** (`os` entry) and the OS bytecode length in **`virtual_os_test.rs`**: stale, because the Cairo change alters the OS program. Regenerate with:
      ```bash
      UPDATE_EXPECT=1 cargo test -p apollo_starknet_os_program test_program_hashes test_program_bytecode_lengths
      ```
      (requires the `sequencer_venv` / `cairo-compile` toolchain).
- [ ] **`crates/apollo_gateway/resources/proof_facts_big_storage_keys.json`**: derived from `snos_proof_facts_for_testing()`, so both the marker and the `big_storage_key` are stale. Regenerate with `UPDATE_EXPECT=1 cargo test -p apollo_gateway proof_facts_big_storage_keys_fixture` (needs `pkg-config` / OpenSSL headers, absent here).
- [ ] **Proof fixtures** are V1 proofs and are no longer verifiable by the V2 circuit. Regenerate `apollo_transaction_converter/resources/example_proof{.bin,_facts.json}` via the ignored `regenerate_proof_fixtures` test, add a new `starknet_proof_verifier/resources/regression_test/<version>/` directory, then swap `FIXTURE_PROOF_VERSION` and the `#[case]` in `proof_verifier_test.rs` and drop `0.14.3`.
- [ ] **Tests that fail until those fixtures land**: `starknet_proof_verifier` (`regression_verify_proof_from_old_prover` and the negative tests built on `VALID_PROOF_FIXTURE`), `apollo_transaction_converter` (`test_proof_verification_called_for_invoke_v3_with_proof_facts`, `test_consensus_tx_to_internal_with_proof_facts_verifies_and_sets_proof`), `starknet_transaction_prover` prover tests.
- [ ] Retarget to `main-v0.14.4` once that branch exists.

The V1 fixtures and their test cases are kept in place rather than deleted, so the outstanding work stays visible instead of silently dropping proof-verification regression coverage.


einat-starkware:
Split into a 3-PR stack for reviewability: #15013 → #15014 → #15015. Closing this in favor of those.

PR #15014, opened 2026-08-24
workspace,apollo_starknet_os_program,blockifier,starknet_proof_verifier: switch to proof version V2

einat-starkware:
Replaces proof version V1 with V2 for client-side proving. Follow-up to #15013, which added the V2 marker.

## Changes

- **Cairo OS** — `PROOF_VERSION_V1` → `PROOF_VERSION_V2 = 'PROOF2'`, asserted by `check_proof_facts`. Only the `os` program hash changes; `virtual_os` and `allowed_virtual_os_program_hashes` are unaffected.
- **`starknet_os`** — `Const::ProofVersionV1` → `Const::ProofVersionV2`.
- **`blockifier`** — `allowed_proof_versions` for 0.14.4 → `[V2]`, plus the regenerated diff regression.
- **`starknet_proof_verifier`** — `privacy-circuit-verify` and `privacy-prove` now come from `starkware-libs/proving`; V2 dispatches to the V2 circuit, V0 and V1 are rejected. `try_into_proof_facts` stamps V2, and the negative tests split the proof at `ProofHeader::SIZE`.
- **Fixtures** — regenerated program hash, proof fixtures, and the derived test fixtures.

`rand` moves 0.9.2 → 0.9.5 in `Cargo.lock`, since the new `stwo` requires `^0.9.4`.

PR #15013, opened 2026-08-24
starknet_api: add proof facts version V2

einat-starkware:
> [!NOTE]
> **Part of a 3-PR stack** moving client-side proving from proof version V1 to V2. Review and merge in order:
> 1. **#15013** — `starknet_api: add proof facts version V2` (additive, independently mergeable)
> 2. **#15014** — `apollo_starknet_os_program,starknet_os,blockifier: switch client-side proving to V2`
> 3. **#15015** — `workspace,starknet_proof_verifier: verify proofs with the V2 circuit`
>
> Targets the **`main-v0.14.4`** release branch, since 0.14.4 is the protocol version whose `allowed_proof_versions` this change flips.

## What

Adds the proof-facts version marker `PROOF_VERSION_V2` (`'PROOF2'`, `0x50524f4f4632`) and the `ProofVersion::V2` variant, so proof facts carrying it parse and round-trip through `ProofFactsVariant`.

This is the additive first step of the V1 → V2 migration, mirroring how V0 → V1 was done (#14432 + #14450) — including the "one supported version at a time" policy: older markers are retained purely so historical blocks stay replayable.

## Why this is safe on its own

Nothing accepts V2 yet:

- `allowed_proof_versions` is unchanged, so the blockifier still rejects V2 proof facts under every protocol version.
- The Cairo OS still asserts `PROOF_VERSION_V1`.
- `verify_proof` rejects V2 up front — the new match arm exists only to keep the match exhaustive, and is what #15015 replaces with the real V2 circuit dispatch.

Both placeholder spots carry a `TODO(Einat)` naming exactly what to change once the OS and the V2 circuit are wired up:

- the `ProofVersion` doc comment → restate V1 as replay-only and V2 as the only verifiable version (done in #15015);
- the `verify_proof` match arm → give V2 its own arm dispatching to the V2 circuit, and reject V1 there instead (done in #15015).

`V0` stays documented as replay-only: it appears in historical blocks under protocol 0.14.2, and acceptance is gated per protocol version in the blockifier.

## Test coverage

`ProofVersion` derives `EnumIter` under `testing`, so the existing `fields_test.rs` tests pick V2 up for free — `proof_facts_variant_accepts_supported_versions` now parses V2-marked facts, and `proof_version_str_encodes_to_felt` checks that `"PROOF2"` encodes to the new felt.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #15015, opened 2026-08-24
workspace,starknet_proof_verifier: verify proofs with the V2 circuit

einat-starkware:
> [!NOTE]
> **Part of a 3-PR stack** moving client-side proving from proof version V1 to V2. Review and merge in order:
> 1. **#15013** — `starknet_api: add proof facts version V2` (additive, independently mergeable)
> 2. **#15014** — `apollo_starknet_os_program,starknet_os,blockifier: switch client-side proving to V2`
> 3. **#15015** — `workspace,starknet_proof_verifier: verify proofs with the V2 circuit`
>
> Targets the **`main-v0.14.4`** release branch, since 0.14.4 is the protocol version whose `allowed_proof_versions` this change flips.

> [!IMPORTANT]
> **This PR does not build.** `privacy-circuit-verify-v2` and `privacy-prove` point at `tag = "v0.14.4"`, which does not exist on `proving-utils` yet. Until that tag is pushed, `cargo metadata` fails and nothing in the workspace resolves. PRs 1 and 2 of the stack are unaffected and keep their CI signal.

## What

Switches the proof verifier to the V2 circuit.

- Workspace alias `privacy-circuit-verify-v1` → `privacy-circuit-verify-v2`, in the root `Cargo.toml` and in `starknet_proof_verifier` / `starknet_transaction_prover`.
- `verify_proof` dispatches V2 proof facts to the V2 circuit and rejects V0 and V1 up front with `InvalidProofVersion` — their circuits were removed. Both remain tolerated by the blockifier (gated per protocol version) so historical blocks stay replayable: V0 appears under protocol 0.14.2, V1 under the released 0.14.3.
- `ProgramOutput::try_into_proof_facts` stamps V2, so newly proven transactions carry the marker the OS asserts after PR2.
- Negative tests gain a V1 rejection case alongside the existing V0 one; `verify_proof_rejects_non_v1_proof_version` is renamed accordingly.
- The V2 wire format prefixes the proof with a `ProofHeader` (version + the circuit's preprocessed root) instead of the bare version, so `split_and_decompress_proof` splits at `ProofHeader::SIZE` rather than `VERSION_BYTES`. Regenerated fixtures will be 32 bytes longer in the header.

## Validated against the draft prover

Checked against the large-SNIP-36 draft (`proving-dev` @ `1db917c04859`, review #184): `cargo check -p starknet_proof_verifier --tests` passes clean. `PrivacyProofOutput`, `verify_recursive_circuit`, and `privacy_recursive_prove` all keep their current shapes, so this PR's call sites are unaffected.

A single `verify_recursive_circuit` entry point covers both small and large proofs — it selects the circuit config from the header's `preprocessed_root` — which is why one `ProofVersion::V2` arm is sufficient and no separate large/small markers are needed.

Two lockfile-level consequences to expect when the real tag lands — both `Cargo.lock` only, no manifest changes needed:

- `stwo v2.4.0` declares `rand = "0.9.4"`, and our single locked `0.9.x` entry is `0.9.2` (required as `^0.9.2` by `starknet-types-core`). The ranges overlap, so this is just a pin to release: `cargo update -p rand@0.9.2 --precise 0.9.5`.
- `proving-dev` carries stwo in-tree, so the transitive stwo crates re-source from `starkware-libs/stwo` / `stwo-cairo` / `stwo-circuits` onto it and bump version (`stwo` 2.2.0 → 2.4.0, `stwo-cairo` 1.2.2 → 2.4.0) — about 24 lockfile entries. The sequencer declares no stwo dependency itself, so nothing in our manifests is affected.

## Blocked on the V2 release

- [ ] **`Cargo.toml`**: replace the placeholder `v0.14.4` tag with the real V2 `proving-utils` release.
- [ ] **`Cargo.lock`**: cannot be regenerated while the tag is missing; deliberately left unchanged.
- [ ] **Proof fixtures** are V1 proofs, no longer verifiable by the V2 circuit. Regenerate `apollo_transaction_converter/resources/example_proof{.bin,_facts.json}` via the ignored `regenerate_proof_fixtures` test, add a new `starknet_proof_verifier/resources/regression_test/<version>/` directory, then point `FIXTURE_PROOF_VERSION` and the `#[case]` at it and drop `0.14.3`.
- [ ] **Tests failing until those fixtures land**: `starknet_proof_verifier` (`regression_verify_proof_from_old_prover` plus the negative tests built on `VALID_PROOF_FIXTURE`), `apollo_transaction_converter` (`test_proof_verification_called_for_invoke_v3_with_proof_facts`, `test_consensus_tx_to_internal_with_proof_facts_verifies_and_sets_proof`), and the `starknet_transaction_prover` prover tests.

The V1 fixtures and their test cases are kept in place with TODOs rather than deleted, so the outstanding work stays visible instead of silently dropping proof-verification regression coverage.


einat-starkware:
Folded into #15014. The protocol switch and the verifier switch can't pass CI independently: `allowed_proof_versions` accepts only V2, but `ProgramOutput::try_into_proof_facts` (which stamps V2) lives in this PR — so the protocol half alone leaves V1 fixtures the gateway rejects, and moving the stamp down alone leaves V2 fixtures the V1 verifier rejects. Closing in favour of the combined PR.

PR #15050, opened 2026-08-26
apollo_consensus_orchestrator: clamp the fee proposal band to the L2 gas price cap

asaf-sw:
Ports #14946 onto `main-v0.14.4`. The fee proposal band is now clamped to the L2 gas price ceiling, and this adds the **only increment site in the codebase** for `SNIP35_FEE_TARGET_ABOVE_MAXIMUM`.

That second part closes a live gap. The metric is already defined and registered on trunk (from #14978), and both an alert and a dashboard panel read it, but nothing increments it. The alert is deployed on all six mainnet nodes today and is permanently silent until this lands.

Deferred from 0.14.3 at Matan's request, so the metric would be rehomed into the PR that actually enforces the cap rather than sitting unused in #14978.

---

# Detailed Summary for AI Bots

## What changed

- `compute_fee_proposal` and `fee_proposal_bounds` take a `max_gas_price` and cap **both** band edges at it.
- The oracle-failure path (`fee_target == None`) now goes through the same clamp instead of returning `fee_actual` raw.
- `ProposalInitValidation` gains `max_l2_gas_price`. The validator derives it from its **own** `min_l2_gas_price_per_height` via `l2_gas_price_cap(get_min_gas_price_for_height(..))` at both validate call sites, never from the proposal. A malicious proposer therefore cannot widen the band it is checked against.
- `SNIP35_FEE_TARGET_ABOVE_MAXIMUM` is incremented in `compute_proposer_fee_proposal`, guarded by `override_l2_gas_price_fri.is_none()` so that targets originating from the operator pin are excluded, per the metric's own description. Reachable from `ConsensusContext::build_proposal`.

## Port notes

The source branch also added the metric definition and its `register_metrics()` line. Those were **deliberately not ported**, because `main-v0.14.4` already carries both from #14978. Verified there is no duplicate definition or registration.

The predecessor commit on the source branch (`l2_gas_price_cap`, `MAX_GAS_PRICE_MULTIPLIER`, `NextL2GasPrice`, `CONSENSUS_L2_GAS_PRICE_CLAMPED`) is already on `main-v0.14.4`, so only the second commit needed porting.

## Verification

`rust_fmt`, `cargo clippy -p apollo_consensus_orchestrator --all-features --all-targets -- -D warnings`, and `SEED=0 cargo nextest run -p apollo_consensus_orchestrator --all-features` (238 tests) all pass. `apollo_dashboard` is untouched, so no dashboard regeneration was needed.

[...]

PR #15051, opened 2026-08-26
apollo_versioned_constants,apollo_consensus_orchestrator: version the L2 gas price multiplier

asaf-sw:
Moves the L2 gas price ceiling multiplier from a compile-time constant into `VersionedConstants`, so changing the ceiling no longer needs a release. Resolves the `TODO(asaf-sw)` in `fee_market/mod.rs`.

Every version's resource states the value explicitly as 10, which is what `MAX_GAS_PRICE_MULTIPLIER` was as a compile-time constant applied to all versions, so no version's pricing changes.

Stacked on #15050.

---

## What changed

- `VersionedConstants` (`apollo_versioned_constants`) gains `max_gas_price_multiplier: u128`.
- `"max_gas_price_multiplier": 10` added to all five `orchestrator_versioned_constants_0_14_{0..4}.json`. All five carry an identical key set, matching the convention across `blockifier`'s 17 resources: every version's JSON states every field, and neither VC struct uses a serde default.
- `l2_gas_price_cap` reads the value from `VersionedConstants::latest_constants()` and keeps `saturating_mul`.
- The deleted `const _: () = assert!(MAX_GAS_PRICE_MULTIPLIER > 1)` is replaced by a test, `max_gas_price_multiplier_is_above_one_in_every_version` in `apollo_versioned_constants`, which asserts the invariant for every version the crate ships. There is no runtime check: the value is compiled in via `include_str!` behind a `LazyLock`, so it cannot vary at runtime, and a per-block assert on the fee path would only turn a CI failure into a node panic.

## Required step for the forward merge to main

**Mandatory.** `main` carries `orchestrator_versioned_constants_0_14_5.json`, which this branch cannot touch because that file does not exist on the `main-v0.14.4` lineage. Add `"max_gas_price_multiplier": 10` to it during the merge.

Without it, `serde_json::from_str` on that resource fails with `missing field max_gas_price_multiplier`, and the first `VersionedConstants::latest_constants()` call panics, since `V0_14_5` is `LATEST` on main. That is a crash on the first block proposal or validation after startup.

Both `max_gas_price_multiplier_is_above_one_in_every_version` and `test_vc_diffs_regression` catch the omission in CI. Verified by negative control: stripping the key from `orchestrator_versioned_constants_0_14_2.json` fails the former with exactly that `missing field` panic.

## Not affected

The `consensus_l2_gas_price_above_maximum` alert reads the node's own clamp counter and carries no threshold of its own, so it follows the constant automatically. Only its doc comment mentioned the old identifier, and that is updated.

## Verification

`rust_fmt`, clippy on `apollo_versioned_constants` / `apollo_consensus_orchestrator` / `apollo_dashboard` with `-D warnings`, 262 tests across those three packages, and `apollo_versioned_constants` standalone (feature unification differs, and standalone is the shape of CI's per-package job).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2919,6 +2919,7 @@ version = "0.20.0-rc.0"
 dependencies = [
  "serde",
  "starknet_api",
+ "strum",
  "thiserror 1.0.69",
 ]
 
--- a/crates/apollo_consensus_orchestrator/src/fee_market/mod.rs
+++ b/crates/apollo_consensus_orchestrator/src/fee_market/mod.rs
@@ -22,12 +22,6 @@ mod test;
 // in approximately 10 minutes.
 const MIN_GAS_PRICE_INCREASE_DENOMINATOR: u128 = 333;
 
-// Ceiling on the L2 gas price, as a multiple of the minimum gas price in force for the height.
-// TODO(asaf-sw): Move to versioned constants in 0.14.4.
-const MAX_GAS_PRICE_MULTIPLIER: u128 = 10;
-// A multiplier of 0 or 1 pins the L2 gas price at 0 or at the minimum.
-const _: () = assert!(MAX_GAS_PRICE_MULTIPLIER > 1);
-
 /// Fee market information for the next block.
 #[cfg_attr(any(feature = "testing", test), derive(serde::Deserialize, PartialEq))]
 #[derive(Debug, Default, Serialize)]
@@ -59,11 +53,12 @@ pub fn get_min_gas_price_for_height(
         .unwrap_or(fallback_min_gas_price)
 }
 
-/// The ceiling on the L2 gas price: `MAX_GAS_PRICE_MULTIPLIER` times `min_gas_price`. Applies to
+/// The ceiling on the L2 gas price: `max_gas_price_multiplier` times `min_gas_price`. Applies to
 /// every block regardless of its `starknet_version`. Proposer and validator reach the ceiling
 /// through this function, so they cannot disagree on it.
 pub fn l2_gas_price_cap(min_gas_price: GasPrice) -> GasPrice {
-    GasPrice(min_gas_price.0.saturating_mul(MAX_GAS_PRICE_MULTIPLIER))
+    let max_gas_price_multiplier = VersionedConstants::latest_constants().max_gas_price_multiplier;
+    GasPrice(min_gas_price.0.saturating_mul(max_gas_price_multiplier))
 }
 
 /// The ceiling in force at `height`: the cap over the minimum configured for that height. The
--- a/crates/apollo_consensus_orchestrator/src/fee_market/test.rs
+++ b/crates/apollo_consensus_orchestrator/src/fee_market/test.rs
@@ -13,7 +13,6 @@ use crate::fee_market::{
     calculate_next_l2_gas_price_for_fin,
     get_min_gas_price_for_height,
     l2_gas_price_cap,
-    MAX_GAS_PRICE_MULTIPLIER,
     MIN_GAS_PRICE_INCREASE_DENOMINATOR,
 };
 use crate::metrics::{CONSENSUS_L2_GAS_PRICE_CLAMPED, LABEL_L2_GAS_PRICE_CLAMP_BOUND};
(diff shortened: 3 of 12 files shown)


matanl-starkware (changes requested):
> @matanl-starkware reviewed 9 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15051)* status: all files reviewed, 2 unresolved discussions (waiting on asaf-sw).

___
*[`crates/apollo_consensus_orchestrator/src/fee_market/mod.rs` line 68 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15051#-P01y_Yr33fAO3cr5EMr:-P01y_Yr33fAO3cr5EMs:b-e7tf86) ([raw file](https://github.com/starkware-libs/sequencer/blob/5dd384ef4901fb3f71eb71508f3425a5e32c7b45/crates/apollo_consensus_orchestrator/src/fee_market/mod.rs#L68)):*
> ```XML
>         max_gas_price_multiplier > 1,
>         "max_gas_price_multiplier is {max_gas_price_multiplier}, must be greater than one."
>     );
> ```

You have `max_gas_price_multiplier_is_above_one_in_every_version`, why do you also need a runtime check?



_Code quote:_
```XML
    // A multiplier of 0 or 1 pins the L2 gas price at 0 or at the minimum. Backstop for
    // `max_gas_price_multiplier_is_above_one_in_every_version`, which gates every shipped resource:
    // the value comes from a compiled-in resource or the field default, so a binary that passed
    // CI cannot trip this.
    assert!(
        max_gas_price_multiplier > 1,
        "max_gas_price_multiplier is {max_gas_price_multiplier}, must be greater than one."
    );
```
___
*[`crates/apollo_versioned_constants/src/lib.rs` line 25 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15051#-P01ys4q9ZKQz-FOp_fK:-P01ys4r0BOfUX586G8v:byyrhx4) ([raw file](https://github.com/starkware-libs/sequencer/blob/5dd384ef4901fb3f71eb71508f3425a5e32c7b45/crates/apollo_versioned_constants/src/lib.rs#L25)):*
> ```XML
>     /// that omits it inherits 10, the value that applied to every version before this constant
>     /// was versioned.
>     #[serde(default = "default_max_gas_price_multiplier")]
> ```

Why do we need this? (other members do not have it)



_Code quote:_
```XML
#[serde(default = "default_max_gas_price_multiplier")]
```

asaf-sw (commented):
> @asaf-sw+AGNT made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15051)* status: 6 of 13 files reviewed, 2 unresolved discussions (waiting on matanl-starkware).

___
*[`crates/apollo_consensus_orchestrator/src/fee_market/mod.rs` line 68 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15051#-P01y_Yr33fAO3cr5EMr:-P025u5O3kwqPGS9SiBO:b7ulfwe) ([raw file](https://github.com/starkware-libs/sequencer/blob/5dd384ef4901fb3f71eb71508f3425a5e32c7b45/crates/apollo_consensus_orchestrator/src/fee_market/mod.rs#L68)):*

You have `max_gas_price_multiplier_is_above_one_in_every_version`, why do you also need a runtime check?
</blockquote></details>

You're right, removed.
___
*[`crates/apollo_versioned_constants/src/lib.rs` line 25 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15051#-P01ys4q9ZKQz-FOp_fK:-P025x6E3-bifcOkulzf:bhsn6mx) ([raw file](https://github.com/starkware-libs/sequencer/blob/5dd384ef4901fb3f71eb71508f3425a5e32c7b45/crates/apollo_versioned_constants/src/lib.rs#L25)):*

Why do we need this? (other members do not have it)
</blockquote></details>

Removed the default and stated the key in all five resources. It would have been the only VC struct field with one.

matanl-starkware (approved):
> @matanl-starkware reviewed 7 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15051)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on asaf-sw).

gkaempfer:

---

PR #15058, opened 2026-08-27
starknet_os: add Rust mirror of the privacy proof-fact fold

einat-starkware:
Part 1 of 2 of the privacy proof-fact fold stack (OS side of "proving that privacy proofs were verified").

Adds `starknet_os::proof_fact_fold`, a Rust implementation of the digest chain the proving side builds over privacy-transaction leaf proofs (`stwo_run_and_prove_recursive_tree` + the multiverifier circuit):

- **Leaf digest**: `blake2s(encode_felt252s_to_u32s(proof_facts[2..]))` — drops the proof version and variant markers, keeping `[program_hash, ...virtual OS output]`, exactly the preimage the leaf simple bootloader hashes to its output (which the leaf cairo-verifier circuit emits verbatim as its public output).
- **Fold**: adjacent entries paired left-to-right; a parent's output digest is blake2s over the children's 32 raw u32 words (`circuit_hash ‖ output` per child, no felt encoding at fold levels); a trailing unpaired entry is carried unchanged, keeping its own circuit hash; a single-entry block self-folds.
- **Verifier output digest**: `blake2s(circuit_hash ‖ output_digest)` of the root entry — what the Cairo circuit verifier's `get_verification_output` returns.

Tests pin the implementation against golden values from real proven trees on the proving side: the leaf `H1` golden and the `four_leaves` fixture's root output digest.

The circuit hash constants are the `canonical_small` test-registry values (documented as such); they must be replaced with the production registry's values before production wiring.

Next in the stack: the Cairo0 implementation of the same fold, pinned bit-for-bit against this module.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XmPJM3Wph4QLmFmhcxVsh4

---
_Generated by [Claude Code](https://claude.ai/code/session_01XmPJM3Wph4QLmFmhcxVsh4)_

PR #15059, opened 2026-08-27
apollo_starknet_os_program,starknet_os: add Cairo0 privacy proof-fact fold

einat-starkware:
Part 2 of 2 of the privacy proof-fact fold stack, on top of #15058.

Adds the OS-side fact fold in Cairo0 (`starkware/starknet/core/os/proof_fact_fold.cairo`):

- Leaf output digest via `encode_felt252s_to_u32s` + `blake_with_opcode` over `proof_facts[2..]`.
- Pairwise fold over raw u32 words with the odd-entry carry rule, and the single-entry self-fold. Entries are contiguous 16-word `[circuit_hash, output_digest]` blocks, so two adjacent entries are literally the parent digest's 32-word blake preimage — no copying at fold levels.
- `compute_fold_digest` — the digest the Cairo circuit verifier outputs for the proof whose facts fold to a given entry.
- The leaf-verifier and multiverifier circuit hash constants (`canonical_small` registry values, documented as placeholders pending the production registry).

The module is **not yet reachable from the OS program**: it is compiled as a standalone test program under the `test_programs` feature, so the OS and aggregator program hashes and the bytecode-length pins are unchanged (verified by the existing pin tests). Wiring it into the OS output is deferred to the layering decision (where the block root continues — aggregator vs. a new program).

Tests pin the Cairo implementation bit-for-bit against the Rust mirror from #15058: the self-fold, the carry rule at one and at two layers, deeper trees, the felt-encoding boundaries at 2^63, the fold digest, and the circuit hash constants.

https://claude.ai/code/session_01XmPJM3Wph4QLmFmhcxVsh4

---
_Generated by [Claude Code](https://claude.ai/code/session_01XmPJM3Wph4QLmFmhcxVsh4)_

### DIFF
--- a/crates/apollo_starknet_os_program/build/compile_program.rs
+++ b/crates/apollo_starknet_os_program/build/compile_program.rs
@@ -52,6 +52,14 @@ pub async fn compile_test_contracts(out_dir: PathBuf) {
         "builtin_offset_increase_test",
         None,
     ));
+    // The proof-fact fold module is not yet reachable from the OS program, so it is
+    // compiled standalone for its Cairo-vs-Rust tests.
+    task_set.spawn(compile_and_output_program(
+        out_dir.clone(),
+        "starkware/starknet/core/os/proof_fact_fold.cairo",
+        "proof_fact_fold",
+        None,
+    ));
     task_set.join_all().await;
 }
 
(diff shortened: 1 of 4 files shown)


einat-starkware:
Closing: split into two smaller PRs per review feedback — #15092 (Cairo leaf digest) and #15093 (Cairo fold tree). No content is lost; #15064 is now retargeted onto #15093.

---

PR #15063, opened 2026-08-27
apollo_dashboard: add no-data fallbacks outside the observer guard for stuck alerts

matanl-starkware:
## Problem

Grafana rules are generated with `noDataState: NoData` (`deployments/monitoring/src/common/grafana10_objects.py:207`). That is **not silent** — per Grafana's docs an absent series raises a `DatasourceNoData` alert instance that routes to notification policies and contact points.

Eleven `LessThan` stuck-detectors could reach that state from a scrape gap alone, notifying on a perfectly healthy node.

## Evidence

Mainnet node-3, 2026-08-27. Scrapes for the whole `starknet-mainnet-apollo-3` cluster thinned from 6/min to 0 between 12:38:30 and 12:45:30 UTC — all 8 pods, 5 nodes, 3 node pools; the five sibling clusters were unaffected. The node itself was healthy throughout: 285 blocks added, 284 precommits broadcast, zero round changes fleet-wide.

`increase(apollo_state_sync_class_manager_marker[120s])` over that window:

```
12:41:00  50.107
   ...     (no data — fewer than 2 samples in every 120s window)
12:45:00  21.893
```

## Fix

`Alert::with_no_data_fallback` keeps the rule defined at a non-firing value. Two constraints make the placement non-obvious, and the helper enforces both:

**1. The fallback goes after the observer guard.** `Alert::new` wraps `NotApplicable` alerts as `(expr) and on() (is_observer == 0)`. A fallback appended by the caller lands inside that, and `is_observer` is scraped from the same pod as the metric it guards — so past the 5m staleness lookback the `and` has an empty right side and discards the fallback:

```promql
((sum(increase(…)) or vector(1)) and on() (absent == 0))   -- 0 series
((sum(increase(…))) and on() (absent == 0)) or vector(1)   -- 1
```

**2. The expression must aggregate away its labels.** `or` only suppresses a right-hand element whose label set already exists on the left, and `vector(N)` is labelless:

```promql
increase(m{ns}[2m]) or vector(1)        -- 2 series (real + constant 1, always)
sum(increase(m{ns}[2m])) or vector(1)   -- 1 series
```

Aggregation is safe because `alert_builder.py:168` substitutes an exact `namespace=`/`cluster=`, so it collapses pods within one namespace rather than across nodes. `sum` for the counters, matching the existing idiom; `min` for the peer gauges so a pod below the threshold still alerts.

Detection is unaffected: a real stall keeps the gauge present, so `increase` yields a present `0` and still fires. The fallback must stay `>=` the threshold, hence `vector(2)` for the two `peer_down` alerts.

## Scope

Six alerts that had no fallback (nine generated rules):

| Alert | Threshold | Fallback |
|---|---|---|
| `state_sync_stuck` (+`_long_time`) | < 1 | `vector(1)` |
| `preconfirmed_block_not_written` | < 1 | `vector(1)` |
| `batched_transactions_stuck` (+`_long_time`) | < 1 | `vector(1)` |
| `consensus_p2p_not_enough_peers_for_quorum` (+`_long_time`) | < 1 | `vector(1)` |
| `consensus_p2p_peer_down` | < 2 | `vector(2)` |
| `mempool_p2p_peer_down` | < 2 | `vector(2)` |

Plus the five existing fallbacks from #14705 and #14876, moved outside the guard — placed inside, those only ever worked for gaps shorter than the staleness lookback.

## Not in this PR

- Four alerts still pair `or vector(0)` with `LessThan`, which turns no-data into a hard `0` and a **false page**: `consensus_block_number_progress_is_slow`, `consensus_decisions_reached_by_consensus_ratio`, `consensus_votes_num_sent_messages`, `gateway_low_successful_transaction_rate`. During this same incident `consensus_block_number_progress_is_slow` held `0.000` for seven consecutive evaluations on the healthy node. Higher severity than this PR; worth a follow-up.
- Long-window blind alerts (15m–1h) are left alone — a gap of this size still leaves samples at the window edges.
- `pod_state_not_ready` is deliberately unchanged: the series vanishing *is* the signal, and a labelless fallback would not join with its `sum by (service_name)`.

## Testing

- `cargo test -p apollo_dashboard` — 22 passed
- `cargo clippy -p apollo_dashboard --all-targets -- -D warnings` — clean
- `cargo run --bin sequencer_dashboard_generator` re-run
- Each generated rule verified to end in a trailing fallback `>=` its threshold
- Behaviour checked against live GMP: data present → real value only; guard absent → fallback survives; metric absent → fallback; validator → real value, single series

No placeholders added or removed, so no per-env override changes are needed.

### DIFF
--- a/crates/apollo_dashboard/resources/dev_grafana_alerts.json
+++ b/crates/apollo_dashboard/resources/dev_grafana_alerts.json
@@ -18,7 +18,7 @@
       "name": "batched_transactions_stuck",
       "title": "Batched Transactions Stuck",
       "ruleGroup": "evaluation_rate_default",
-      "expr": "(changes(batcher_batched_transactions{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$$$batched_transactions_stuck-sampling_window_secs-expression$$$s])) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)",
+      "expr": "((sum(changes(batcher_batched_transactions{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$$$batched_transactions_stuck-sampling_window_secs-expression$$$s]))) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)) or vector(1)",
       "conditions": [
         {
           "evaluator": {
@@ -44,7 +44,7 @@
       "name": "batched_transactions_stuck_long_time",
       "title": "Batched Transactions Stuck Long Time",
       "ruleGroup": "evaluation_rate_default",
-      "expr": "(changes(batcher_batched_transactions{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$$$batched_transactions_stuck_long_time-sampling_window_secs-expression$$$s])) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)",
+      "expr": "((sum(changes(batcher_batched_transactions{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$$$batched_transactions_stuck_long_time-sampling_window_secs-expression$$$s]))) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)) or vector(1)",
       "conditions": [
         {
           "evaluator": {
@@ -408,7 +408,7 @@
       "name": "consensus_block_number_stuck",
       "title": "Consensus Block Number Stuck",
       "ruleGroup": "evaluation_rate_default",
-      "expr": "(sum(increase(consensus_block_number{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$$$consensus_block_number_stuck-sampling_window_secs-expression$$$s])) or vector(1)) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)",
+      "expr": "((sum(increase(consensus_block_number{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$$$consensus_block_number_stuck-sampling_window_secs-expression$$$s]))) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)) or vector(1)",
       "conditions": [
         {
           "evaluator": {
@@ -434,7 +434,7 @@
       "name": "consensus_block_number_stuck_long_time",
       "title": "Consensus Block Number Stuck Long Time",
       "ruleGroup": "evaluation_rate_default",
-      "expr": "(sum(increase(consensus_block_number{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$$$consensus_block_number_stuck_long_time-sampling_window_secs-expression$$$s])) or vector(1)) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)",
+      "expr": "((sum(increase(consensus_block_number{cluster=~\"$cluster\", namespace=~\"$namespace\"}[$$$consensus_block_number_stuck_long_time-sampling_window_secs-expression$$$s]))) and on() (is_observer{cluster=~\"$cluster\", namespace=~\"$namespace\"} == 0)) or vector(1)",
       "conditions": [
         {
           "evaluator": {
@@ -720,7 +720,7 @@
       "name": "consensus_p2p_not_enough_peers_for_quorum",
       "title": "Consensus P2P Not Enough Peers For Quorum",
       "ruleGroup": "evaluation_rate_default",
-      "expr": "max_over_time(apollo_consensus_num_connected_peers{cluster=~\"$cluster\", namespace=~\"$namespace\"}[120s])",
+      "expr": "(min(max_over_time(apollo_consensus_num_connected_peers{cluster=~\"$cluster\", namespace=~\"$namespace\"}[120s]))) or vector(1)",
       "conditions": [
         {
           "evaluator": {
@@ -746,7 +746,7 @@
       "name": "consensus_p2p_not_enough_peers_for_quorum_long_time",
       "title": "Consensus P2P Not Enough Peers For Quorum Long Time",
       "ruleGroup": "evaluation_rate_default",
-      "expr": "max_over_time(apollo_consensus_num_connected_peers{cluster=~\"$cluster\", namespace=~\"$namespace\"}[1800s])",
+      "expr": "(min(max_over_time(apollo_consensus_num_connected_peers{cluster=~\"$cluster\", namespace=~\"$namespace\"}[1800s]))) or vector(1)",
       "conditions": [
         {
           "evaluator": {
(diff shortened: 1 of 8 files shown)


ron-starkware (changes requested):
> @ron-starkware made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15063)* status: 0 of 6 files reviewed, 2 unresolved discussions (waiting on matanl-starkware).

___
*[`crates/apollo_dashboard/src/alert_scenarios/block_production_halt.rs` line 74 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15063#-P02dLqcC8yeiV_vevGB:-P02dLqcC8yeiV_vevGC:b-sbpkfj) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3e70c574d06ddf31e6290ea46a277adf7f4e92d/crates/apollo_dashboard/src/alert_scenarios/block_production_halt.rs#L74)):*
> ```XML
>     let name = title.to_lowercase().replace(' ', "_");
>     let expr_template_string =
>         format!("changes({}[{{}}s]) or vector(1)", BATCHED_TRANSACTIONS.get_name_with_filter());
> ```

This alert (and preconfirmed_block_not_written, mempool_p2p_peer_down) gets wrapped as (expr or vector(1)) and on() (is_observer == 0). The fallback is inside the parens, but in a full scrape gap is_observer stops being scraped too — it survives only ~5 more minutes (Prometheus staleness lookback), and then the and has an empty right side, which wipes out everything on the left, including the vector(1). Result: for gaps longer than ~5 minutes these rules are no-data blind again — and the incident that motivated this PR was a 7-minute gap.

Fix: put the fallback outside the guard instead: ((expr) and on() (is_observer == 0)) or vector(1). Then a gap of any length returns 1. On observer nodes it also returns 1, which is >= the threshold, so it still never fires there — the only change is they show Normal instead of NoData.



_Code quote:_
```XML
changes({}[{{}}s]) or vector(1)
```
___
*[`crates/apollo_dashboard/src/alert_scenarios/preconfirmed.rs` line 22 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15063#-P02fbuJ94wsRDVse_9j:-P02fbuJ94wsRDVse_9k:b-hop14g) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3e70c574d06ddf31e6290ea46a277adf7f4e92d/crates/apollo_dashboard/src/alert_scenarios/preconfirmed.rs#L22)):*
> ```XML
>         "Preconfirmed block not written",
>         EvaluationRate::Default,
>         format!("increase({}[2m]) or vector(1)", PRECONFIRMED_BLOCK_WRITTEN.get_name_with_filter()),
> ```

Small gotcha with or vector(1) here: vector(1) has no labels, while increase(metric{...}) keeps the metric's labels. PromQL's or only drops right-hand elements whose label set already exists on the left — and a labelless entry never matches a labeled one. So the constant 1 series is returned on every evaluation, not only during gaps. (The existing sum(increase(...)) or vector(1) alerts don't have this: sum strips the labels, so the fallback really is suppressed when data exists.)

It can't fire (1 is not < 1), but it means these rules can never show NoData again under any circumstance. If the metric is ever missing for a real, permanent reason — renamed metric, component not deployed in some env, wrong $cluster/$namespace regex — the rule used to sit in perma-NoData, which is visible and gets investigated. Now it will just show OK forever, and nobody will notice the alert is dead. If the metrics here are single-pod-per-namespace, wrapping in sum(...) like the existing idiom avoids this.



_Code quote:_
```XML
increase({}[2m]) or vector(1)
```

matanl-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15063)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on ron-starkware).

___
*[`crates/apollo_dashboard/src/alert_scenarios/block_production_halt.rs` line 74 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15063#-P02dLqcC8yeiV_vevGB:-P0EFK1GBqfksUi6HNTG:b-t4jypj) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3e70c574d06ddf31e6290ea46a277adf7f4e92d/crates/apollo_dashboard/src/alert_scenarios/block_production_halt.rs#L74)):*

This alert (and preconfirmed_block_not_written, mempool_p2p_peer_down) gets wrapped as (expr or vector(1)) and on() (is_observer == 0). The fallback is inside the parens, but in a full scrape gap is_observer stops being scraped too — it survives only ~5 more minutes (Prometheus staleness lookback), and then the and has an empty right side, which wipes out everything on the left, including the vector(1). Result: for gaps longer than ~5 minutes these rules are no-data blind again — and the incident that motivated this PR was a 7-minute gap.

Fix: put the fallback outside the guard instead: ((expr) and on() (is_observer == 0)) or vector(1). Then a gap of any length returns 1. On observer nodes it also returns 1, which is >= the threshold, so it still never fires there — the only change is they show Normal instead of NoData.
</blockquote></details>

Good catch! Thanks.
___
*[`crates/apollo_dashboard/src/alert_scenarios/preconfirmed.rs` line 22 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15063#-P02fbuJ94wsRDVse_9j:-P0EFZCd5NG5jeQ8kT6B:bnputky) ([raw file](https://github.com/starkware-libs/sequencer/blob/e3e70c574d06ddf31e6290ea46a277adf7f4e92d/crates/apollo_dashboard/src/alert_scenarios/preconfirmed.rs#L22)):*

Small gotcha with or vector(1) here: vector(1) has no labels, while increase(metric{...}) keeps the metric's labels. PromQL's or only drops right-hand elements whose label set already exists on the left — and a labelless entry never matches a labeled one. So the constant 1 series is returned on every evaluation, not only during gaps. (The existing sum(increase(...)) or vector(1) alerts don't have this: sum strips the labels, so the fallback really is suppressed when data exists.)

It can't fire (1 is not < 1), but it means these rules can never show NoData again under any circumstance. If the metric is ever missing for a real, permanent reason — renamed metric, component not deployed in some env, wrong $cluster/$namespace regex — the rule used to sit in perma-NoData, which is visible and gets investigated. Now it will just show OK forever, and nobody will notice the alert is dead. If the metrics here are single-pod-per-namespace, wrapping in sum(...) like the existing idiom avoids this.
</blockquote></details>

Done. Thanks

ron-starkware (approved):
I approved, but just in case mentioning that this PR has 3 commits (and not 1 that was amended) in case that wasn't your intention.

> @ron-starkware reviewed 8 files and all commit messages, made 1 comment, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15063#-:-P0EHIL3DoyiM0MN7zck:b-sqsn9y)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

matanl-starkware (commented):
> @matanl-starkware reviewed 8 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15063)* status: all files reviewed (commit messages unreviewed), all discussions resolved (waiting on matanl-starkware).

ron-starkware (approved):
> @ron-starkware reviewed all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15063)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

gkaempfer:

Reviewed diff: touches only Grafana alert-rule generation (PromQL expression building in `apollo_dashboard` + the generated `dev_grafana_alerts.json`). All inputs into the expressions (metric names, threshold constants) are compile-time internal values — no user input, auth, deserialization, or secrets involved.

---

PR #15089, opened 2026-09-02
apollo_dashboard: fix add_tx latency and mempool eviction alerts, remove empty-block alert

dafnamatsry:
Three of the fifteen rules that carried the $cluster substitution bug (#15047)
are also wrong on their own terms, so reviving them would not have made them
work. All three were unobservable until the substitution was fixed, which is why
the defects went unnoticed. Two are fixed here and one is removed.

http_server_min_add_tx_latency - two independent faults, either one fatal:
  - The selector pinned le="1.0", but the exporter writes bucket bounds through
    Display, so the 1-second boundary is le="1" and the selector matched no
    series. The bound now comes from a constant, and a test asserts it is a
    member of HISTOGRAM_BUCKETS.
  - A bare PromQL comparison yields the left operand rather than 1, so the
    product of the two conditions was count x 0 = 0 whenever the sub-second
    bucket was empty, and empty otherwise. Its value set was {0, no data}: it
    could not exceed its own "> 0" threshold under any traffic. Both comparisons
    now use bool.
  - Added a 50-call floor, since a window holding a handful of calls can have
    all of them slow by chance. The 2-minute window is unchanged: it already
    holds ~130 calls at mainnet rates, so widening it would only cost detection
    latency.

mempool_evictions_count - read the raw monotonic counter, so the first eviction
ever recorded would have pinned the alert on until the pod restarted. Now a
10-minute increase, summed across pods so a rollout does not emit one instance
per pod.

high_empty_blocks_ratio - removed. The ratio it measures is set almost entirely
by how many transactions arrived, so it cannot tell "nothing came in" from "we
failed to include what came in", and it fired on quiet traffic. Gating it on
mempool_priority_queue_size does not help either: the gauge is updated on every
get_txs call, and an active batcher keeps calling get_txs and draining the
queue, so the gate is closed exactly while the batcher runs, including while it
produces empty blocks. The failure this alert was meant to catch is covered by
batched_transactions_stuck. Its four override keys must be dropped from the
per-environment alert configs when this is deployed, since the monitoring
builder rejects override keys that no longer map to a placeholder.

Moves the frozen-rate alerts' observer rationale next to the applicability each
caller passes. The values themselves are unchanged and were verified against the
code: validate_proposal reaches the ETH/FRI oracle through
get_l1_prices_in_fri_and_wei, and an observer takes the validator branch every
round (state_machine.rs), so its eth_strk gauge moves; STRK/USD is reached only
through compute_proposer_fee_proposal on the build path, so an observer's
strk_usd gauge never moves.


dafnamatsry:
* **#15089** <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/15089?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/starkware-libs/sequencer/15089?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main-v0.14.4`


<h2></h2>

matanl-starkware (changes requested):
> @matanl-starkware reviewed 6 files and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15089)* status: all files reviewed, 3 unresolved discussions (waiting on dafnamatsry).

___
*[`crates/apollo_dashboard/src/alert_scenarios/l1_endpoints.rs` line 26 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15089#-P1AOVxr0feqF1-6cuPc:-P1AOVxs82rTXXrpc_Rv:b-w6sqt4) ([raw file](https://github.com/starkware-libs/sequencer/blob/c57b919741f8095a2ffb484c051dfe99b374f7f0/crates/apollo_dashboard/src/alert_scenarios/l1_endpoints.rs#L26)):*
> ```XML
>         ScraperLabel::L1GasPrice => 600.0,
>     }
> }
> ```

I don't see the benefit in this change



_Code quote:_
```XML
/// How long the primary may stay down before paging, per scraper.
fn max_primary_down_seconds(scraper: ScraperLabel) -> f64 {
    match scraper {
        ScraperLabel::L1Events => 600.0,
        ScraperLabel::L1GasPrice => 600.0,
    }
}
```
___
*[`crates/apollo_dashboard/src/alert_scenarios/l1_gas_prices.rs` line 280 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15089#-P1AOx116D27ULJwwGVt:-P1AOx116D27ULJwwGVu:b-6det6a) ([raw file](https://github.com/starkware-libs/sequencer/blob/c57b919741f8095a2ffb484c051dfe99b374f7f0/crates/apollo_dashboard/src/alert_scenarios/l1_gas_prices.rs#L280)):*
> ```XML
> /// "frozen"); only a present-but-flat gauge trips this.
> ///
> /// Observer applicability differs per pair, so the caller supplies it.
> ```

This is obvious



_Code quote:_
```XML
///
/// Observer applicability differs per pair, so the caller supplies it.
```
___
*[`crates/apollo_dashboard/src/alert_scenarios/transaction_delays.rs` line 147 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15089#-P1AM9DVDJTRBtRQiTn-:-P1AM9DVDJTRBtRQiTn0:b-3ekrw4) ([raw file](https://github.com/starkware-libs/sequencer/blob/c57b919741f8095a2ffb484c051dfe99b374f7f0/crates/apollo_dashboard/src/alert_scenarios/transaction_delays.rs#L147)):*
> ```XML
>     let expr_template_string = format!(
>         "(sum(increase({zero_bucket}[{{}}s])) / clamp_min(sum(increase({total_count}[{{}}s])), \
>          1)) * (max(min_over_time({ready_txs}[{READY_TXS_WINDOW}])) > bool {READY_TXS_THRESHOLD})"
> ```

The `MEMPOOL_PRIORITY_QUEUE_SIZE` metric is updated every time get_txs is called, so as long as the batcher is active, it will keep calling `get_txs` and empty the priority queue.
I tend to think we should delete this alert altogether and keep only the `batch_transactions_stuck` alert.
I'll take it with Ohad.



_Code quote:_
```XML
(max(min_over_time({ready_txs}[{READY_TXS_WINDOW}])) > bool {READY_TXS_THRESHOLD})
```

matanl-starkware (approved):
> @matanl-starkware reviewed 6 files and all commit messages, and resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15089)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on dafnamatsry).

PR #15106, opened 2026-09-10
apollo_consensus_orchestrator: retry transient feeder errors for the retrospective block hash

matanl-starkware:
## Problem

Consensus fetches the retrospective block hash (height − 10) through the state sync client. When state sync has not synced that block yet, `StateSync::get_block_hash` falls back to the feeder gateway (`apollo_state_sync/src/lib.rs`). A transient feeder failure (e.g. HTTP 500) surfaces as `StateSyncError::ReaderClientError`, which `wait_for_retrospective_block_hash` did not treat as retryable — only `BlockNotFound` was. The proposal failed on the first attempt.

Observed on mainnet 2026-09-09 ~01:22–01:30 UTC: a feeder-gateway degradation produced 16 `PROPOSAL_FAILED … RetrospectiveBlockHashError(StateSyncError(StateSyncError(ReaderClientError(… 500 …))))` across the validators with zero retry warnings, round churn up to round 4, and block cadence dropping from ~35/min to 2/min.

## Fix

Treat `ReaderClientError` as "state sync cannot answer yet", the same as `BlockNotFound`, so the existing deadline-bounded retry loop absorbs it. The loop already caps retries at `retrospective_block_hash_deadline`, so this cannot push a proposal past its deadline.

## Test

`wait_for_retrospective_block_hash_state_sync_transient_feeder_error`: state sync returns `ReaderClientError` on the first attempt and the hash on the second; expects `Ok(Some(hash))`. Fails before the fix (the loop returns the error after one attempt), passes after.

## Verification

- `SEED=0 cargo test -p apollo_consensus_orchestrator` — 243 passed, 0 failed
- `scripts/rust_fmt.sh` — clean
- `cargo clippy -p apollo_consensus_orchestrator --all-targets --all-features -- -D warnings` — clean

### DIFF
--- a/crates/apollo_consensus_orchestrator/src/utils.rs
+++ b/crates/apollo_consensus_orchestrator/src/utils.rs
@@ -438,12 +438,12 @@ pub(crate) async fn wait_for_retrospective_block_hash(
         )
         .await;
 
-        // If the block is not found, try again after the retry interval. In any other case, return
-        // the result.
+        // Retry while a client cannot answer yet: the block is not found, or state sync's feeder
+        // fallback failed transiently. In any other case, return the result.
         let state_sync_not_ready = matches!(
             result,
             Err(RetrospectiveBlockHashError::StateSyncError(StateSyncClientError::StateSyncError(
-                StateSyncError::BlockNotFound(_)
+                StateSyncError::BlockNotFound(_) | StateSyncError::ReaderClientError(_)
             )))
         );
         let batcher_not_ready = matches!(
--- a/crates/apollo_consensus_orchestrator/src/utils_test.rs
+++ b/crates/apollo_consensus_orchestrator/src/utils_test.rs
@@ -284,6 +284,61 @@ async fn wait_for_retrospective_block_hash_state_sync_ready_after_a_while() {
 }
 
 #[tokio::test]
+async fn wait_for_retrospective_block_hash_state_sync_transient_feeder_error() {
+    let (mut test_proposal_args, _proposal_receiver) = create_proposal_build_arguments();
+    test_proposal_args.build_param.height = CURRENT_BLOCK_NUMBER;
+    test_proposal_args
+        .deps
+        .batcher
+        .expect_get_block_hash()
+        .withf(|block_number| *block_number == MUST_HAVE_BLOCK_HASH_FOR)
+        .times(2)
+        .returning(|_| Ok(RETRO_BLOCK_HASH));
+    test_proposal_args
+        .deps
+        .batcher
+        .expect_get_block_hash()
+        .withf(|block_number| *block_number == RETRO_BLOCK_NUMBER)
+        .returning(|_| Ok(RETRO_BLOCK_HASH));
+    // State sync falls back to the feeder gateway for blocks it has not synced yet; the feeder
+    // fails transiently (e.g. a 500) in the first attempt.
+    test_proposal_args
+        .deps
+        .state_sync_client
+        .expect_get_block_hash()
+        .withf(|block_number| *block_number == RETRO_BLOCK_NUMBER)
+        .times(1)
+        .returning(|_| {
+            Err(StateSyncError::ReaderClientError("Internal server error".to_string()).into())
+        });
+    test_proposal_args
+        .deps
+        .state_sync_client
+        .expect_get_block_hash()
+        .withf(|block_number| *block_number == RETRO_BLOCK_NUMBER)
+        .times(1)
+        .returning(|_| Ok(RETRO_BLOCK_HASH));
+
+    let proposal_args: ProposalBuildArguments = test_proposal_args.into();
+    let init = get_proposal_init(&proposal_args).await;
+    let res = wait_for_retrospective_block_hash(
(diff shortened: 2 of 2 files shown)


ron-starkware (changes requested):
> @ron-starkware made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15106)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on matanl-starkware).

___
*[`crates/apollo_consensus_orchestrator/src/utils.rs` line 441 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15106#-P1Tf3Pv7UQnb4xjZl1a:-P1Tf3Pv7UQnb4xjZl1b:b91kges) ([raw file](https://github.com/starkware-libs/sequencer/blob/fe1620906ab634c6a2824b3f03bb0fdccd8c58f7/crates/apollo_consensus_orchestrator/src/utils.rs#L441)):*
> ```XML
>         .await;
> 
>         // Retry while a client cannot answer yet: the block is not found, or state sync's feeder
> ```

This deals with a feeder issue on the proposer side - do we want to deal with the same thing on the validator side? Since the same thing happens there, and can lead to a ValidateProposalError. Of course, this will only cause problems if >1/3 of validators have the same problem.
___
*[`crates/apollo_consensus_orchestrator/src/utils.rs` line 446 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15106#-P1TfZam00AiCUuQsjL0:-P1TfZam00AiCUuQsjL1:b52yq6) ([raw file](https://github.com/starkware-libs/sequencer/blob/fe1620906ab634c6a2824b3f03bb0fdccd8c58f7/crates/apollo_consensus_orchestrator/src/utils.rs#L446)):*
> ```XML
>             result,
>             Err(RetrospectiveBlockHashError::StateSyncError(StateSyncClientError::StateSyncError(
>                 StateSyncError::BlockNotFound(_) | StateSyncError::ReaderClientError(_)
> ```

Doesn't this mean that all errors, including ones that probably aren't transient, like AbortedBlock, SerdeError, etc. will be retried? 
Maybe it should be in `From<ReaderClientError>`? Or is it ok that sometimes we will be retying every single time, waiting for 500ms, if it's not a transient error?



_Code quote:_
```XML
StateSyncError::ReaderClientError(_)
```

matanl-starkware (commented):
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15106)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on ron-starkware).

___
*[`crates/apollo_consensus_orchestrator/src/utils.rs` line 441 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15106#-P1Tf3Pv7UQnb4xjZl1a:-P1V-157B5g0zf-ZAwCJ:bul32ri) ([raw file](https://github.com/starkware-libs/sequencer/blob/fe1620906ab634c6a2824b3f03bb0fdccd8c58f7/crates/apollo_consensus_orchestrator/src/utils.rs#L441)):*

This deals with a feeder issue on the proposer side - do we want to deal with the same thing on the validator side? Since the same thing happens there, and can lead to a ValidateProposalError. Of course, this will only cause problems if >1/3 of validators have the same problem.
</blockquote></details>

Agree it hits validators too. But a retry is wrong there - a lagging validator (due to missing this part of information) will cause it to be out of consensus in most cases.
___
*[`crates/apollo_consensus_orchestrator/src/utils.rs` line 446 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15106#-P1TfZam00AiCUuQsjL0:-P1V-UXn0GmJq7MiGevU:b-o1lbi2) ([raw file](https://github.com/starkware-libs/sequencer/blob/fe1620906ab634c6a2824b3f03bb0fdccd8c58f7/crates/apollo_consensus_orchestrator/src/utils.rs#L446)):*

Doesn't this mean that all errors, including ones that probably aren't transient, like AbortedBlock, SerdeError, etc. will be retried? 
Maybe it should be in `From<ReaderClientError>`? Or is it ok that sometimes we will be retying every single time, waiting for 500ms, if it's not a transient error?
</blockquote></details>

Yes, every `ReaderClientError` is retried. IMO that's acceptable: the loop is bounded by the retrospective deadline (~5.7s in prod, ≈11 attempts at 500ms), and validators run TimeoutPropose on their own clocks, so a proposer giving up at 0.1s vs 5.7s changes nothing for the round — cost is ~11 warn lines.

Classifying in `From<ReaderClientError>` would need an `UnknownErrorCode` heuristic anyway: the client parses 500 bodies as `StarknetError { UnknownErrorCode("500") }`
- exactly the incident's error - and the client's own retry doesn't cover it either.

ron-starkware (approved):
> @ron-starkware reviewed 2 files and all commit messages, and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15106)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

gkaempfer:

---

PR #15115, opened 2026-09-14
workspace,scripts: mark workspace-only crates publish = false and enforce it when publishing

avi-starkware:
Only the 16 crates with a `version` in the root workspace manifest are ever published; the other 99 members (97 path-only workspace deps plus `toml_test_utils` and `workspace_tests`) are referenced by path alone. This PR states that explicitly with `publish = false`, which makes `cargo publish` refuse them and lets cargo-deny's `allow-wildcard-paths` treat their versionless path dependencies as intended rather than as wildcards (next PR).

**`scripts/publish_crates.py`**
- Fails early if the publish order contains a `publish = false` crate (via `cargo metadata`; no new Python dependency).
- Follows only `normal,build` edges in `cargo tree`. Its default edges include the root crate's dev-dependencies, so publishing `apollo_config` would have attempted `apollo_test_utils`, and `blockifier` would have pulled in `blockifier_test_utils`.

**Test:** `workspace_tests::test_members_are_unpublishable_iff_they_are_not_for_publish` asserts `publish = false` ⟺ not in the version-derived publish set, so the flag and the `version` rule cannot drift apart (verified to fail when a flag is removed).

**Verified:** `cargo metadata` → 99 restricted / 16 unrestricted, and the 16 are exactly the versioned set; `taplo format --check` clean; `cargo check --workspace --all-targets` green; publish orders for `blockifier`, `starknet_api`, `apollo_starknet_os_program`, `blockifier_test_utils` are all-publishable with no dev-dep leakage.

**Deliberately not touched:** `starknet_proof_verifier` is versioned (publishable) but depends on the git dependency `privacy-circuit-verify-v2` with no crates.io version, so it cannot be published today. cargo-deny keeps reporting it as one wildcard finding; that is accurate and intentionally left visible.


dorimedini-starkware (commented):
> @dorimedini-starkware reviewed 1 file.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15115)* status: 1 of 101 files reviewed, all discussions resolved (waiting on asaf-sw).

einat-starkware (approved):
:lgtm: 

> @einat-starkware reviewed 101 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15115#-:-P1V-KKU1rLnZ4U4yygS:bg7igkv)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on asaf-sw).

PR #15120, opened 2026-09-14
workspace,apollo_starknet_os_program,blockifier,starknet_proof_verifier: accept V1 and V2 proofs

einat-starkware:
Makes the verifier, the OS and the blockifier accept proof version **V1 alongside V2**, so proofs from the previously deployed prover stay valid while V2 rolls out. Follow-up to #15014, which replaced V1 with V2.

The in-repo transaction prover is **unchanged** and keeps producing V2 only — this PR is about what we accept, not what we produce.

## Changes

- **Workspace**: restores `privacy-circuit-verify` from `proving-utils` as the `-v1` dependency, next to `-v2` from `proving`. The lock diff is purely additive — no shared dependency changes version, and `cairo-vm` / `starknet-types-core` stay single crates.io entries so `Felt` unifies across both stacks. `privacy-prove` is untouched (v2 only).
- **`starknet_proof_verifier`**: `verify_proof` dispatches V1 to the V1 circuit and V2 to the V2 circuit; V0 stays rejected (its circuit was removed). `try_into_proof_facts` still stamps V2 — the only version built here.
- **Cairo OS**: `check_proof_facts` accepts `PROOF_VERSION_V1` or `PROOF_VERSION_V2`. Only the `os` program hash changes (`0x3802ed…` → `0x1420d2…`); `virtual_os` is untouched, so `allowed_virtual_os_program_hashes` needs no change.
- **`starknet_os`**: restores `Const::ProofVersionV1`.
- **`blockifier`**: `allowed_proof_versions` for 0.14.4 becomes `[V1, V2]`.

## Testing

Tests cover the V2 flow only, keeping this PR minimal. The one test change is forced by the behavior change: `verify_proof_rejects_non_v2_proof_version` dropped its V1 case and was renamed, because a V1 marker no longer returns `InvalidProofVersion` — it now reaches the V1 circuit.

- `starknet_os_flow_tests::test_proof_facts_versions_and_program_hashes` already loops over `allowed_proof_versions`, so it now runs both versions through the real Cairo `check_proof_facts` without any test change.
- Full suites green: `blockifier` (983), `starknet_os_flow_tests` (115), `starknet_transaction_prover` (137), `starknet_api` (98), `starknet_os` (254), `starknet_proof_verifier` (10). fmt and clippy `-D warnings` clean.

## Notes

The node now links both verify circuits, which is required for the gateway to accept either at runtime.

**The V1 verification path has no Rust test coverage in this PR.** A 0.14.3 V1 regression fixture would exercise `privacy-circuit-verify-v1` end to end; it was left out deliberately to keep the diff small. Worth adding before V1 traffic is relied on in anger.

A follow-up PR adds the gateway flag for restricting which proof versions are accepted.


Yoni-Starkware (approved):
:lgtm:

> @Yoni-Starkware reviewed 14 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15120#-:-P1VKrYm3Ri7YlcRaLRZ:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on einat-starkware).

PR #15122, opened 2026-09-14
apollo_gateway,apollo_gateway_config: add per-version proof acceptance flags

einat-starkware:
Adds `allow_proof_version_v1` and `allow_proof_version_v2` to the stateless transaction validator config. **V1 defaults to `false`, V2 to `true`**, so the gateway accepts only the version the current prover produces — V1 is opt-in, even though #15120 makes the protocol accept it. Either flag can be flipped without a protocol change.

Stacked on #15120, which makes the verifier, the OS and the blockifier accept both versions. **Please review and merge that one first.**

## Changes

- **`apollo_gateway_config`**: two new `bool` fields next to the existing `allow_client_side_proving`, dumped as ordinary scalar params. The deployment app configs set `allow_proof_version_v1: false` to match the default.
- **`apollo_gateway`**: `validate_proof_version_allowed` runs in the stateless validator, right after `validate_client_side_proving_allowed`, so a rejected transaction costs nothing beyond parsing the proof facts. New `ProofVersionNotAllowed` error variant, surfaced as a `ValidationFailure` / `StarknetErrorCode.PROOF_VERSION_NOT_ALLOWED`.
- Regenerated `config_schema.json` and the gateway app configs.

## Design notes

- **Two booleans rather than a list of accepted versions.** The apollo config framework has no JSON-array params — there are zero in `config_schema.json` — so a `Vec<ProofVersion>` would have to be encoded as a delimited string, needing `FromStr` + `AsRef<str>` on `starknet_api::ProofVersion` and custom serde on the field. It would also make `StatelessTransactionValidatorConfig` non-`Copy`, forcing `.clone()` at ~29 `..*DEFAULT_VALIDATOR_CONFIG_FOR_TESTING` sites in the tests. Two scalars need none of that.
- **Proof facts that fail to parse are left alone.** They are rejected later on their own grounds, rather than being reported here as a version problem.
- **V0 has no flag.** The blockifier rejects it for this protocol version regardless, so a gateway-level switch would be dead config.

## Testing

- `test_allow_proof_version` covers the fixture's V2 proof facts accepted with the flag on and rejected with it off. The test config mirrors the production default (`v1: false`, `v2: true`).
- `apollo_gateway` (167), `apollo_gateway_config`, `apollo_node_config` (21), `apollo_deployments` (7) all green. fmt and clippy `-D warnings` clean.


Yoni-Starkware (approved):
:lgtm:

> @Yoni-Starkware reviewed 7 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15122#-:-P1VLfX76CYxjZs20Frx:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on einat-starkware).

gkaempfer:

Reviewed the diff (new `allow_proof_version_v1`/`allow_proof_version_v2` gateway config flags, `validate_proof_version_allowed` in `stateless_transaction_validator.rs`, the new `ProofVersionNotAllowed` error variant, and the regenerated config schema/app configs) for injection, auth-bypass, insecure deserialization, hardcoded secrets, unsafe crypto, and access-control issues. Findings:

- The new check runs after `validate_tx_size` (which bounds `calldata + proof_facts` length), so it can't be used for unbounded-work DoS.
- Unparseable/non-`Snos` proof facts are passed through by design (rejected later on their own grounds) — there's no alternate `ProofFactsVariant` that could smuggle a non-Snos "proof" past this gate.
- The gateway flag is a defense-in-depth/operational filter; the blockifier's `allowed_proof_versions` (from versioned constants) remains the sole authoritative protocol-level gate, so a misconfigured gateway flag can only be over- or under-restrictive at the gateway layer, not a consensus-safety issue.
- No secrets, unsafe arithmetic, or deserialization panics introduced; error messages don't leak sensitive data.

---

PR #15136, opened 2026-09-16
apollo_compile_to_casm,blockifier_test_utils: narrow the bundled libfunc list

Yoni-Starkware:
Stacked on #15133 — review that one first.

Removes a few entries from `allowed_libfuncs.json`, narrowing what can be declared under the `"bundled"` list. The other two list values are Cairo's own built-in lists and are unaffected.

`allowed_libfuncs_aligned_to_audited` asserts our list matches Cairo's audited list, so the omissions now live in a new `EXCLUDED_LIBFUNCS` const that filters the `missing` check. Two extra assertions keep that const honest:

- **present in the json** → the omission silently lapsed. Without this, re-adding an entry would pass every existing check: it is neither missing from the json nor extra to the audited list.
- **no longer in the audited list** → the entry is stale and should be dropped.

`allowed_libfuncs_legacy.json` is generated from the same source (pinned by `expect_file!`), so it is regenerated here too.

**No production impact today:** no environment selects `"bundled"` — mainnet/sepolia-alpha are `"audited"`, integration/testing are `"all"`.

## Testing

- `cargo test -p blockifier --lib` — 983 passed. The cache key hashes the libfunc list, so this recompiled every Cairo 1 feature contract under the new lists, including the two that pin older compiler versions and use the regenerated legacy list.
- `cargo test -p apollo_compile_to_casm` — 9 passed. Verified the new guard fires by temporarily re-adding an excluded entry.
- fmt + `clippy --all-targets -- -D warnings` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #15137, opened 2026-09-16
apollo_sierra_compilation_config,deployment: select the bundled libfunc list instead of audited

Yoni-Starkware:
Stacked on #15133 — review that one first.

Switches every place that selected `"audited"` over to our bundled list:

| | before | after |
|---|---|---|
| `DEFAULT_ALLOWED_LIBFUNCS_LIST` | `Audited` | `Bundled` |
| `mainnet` overlay | `"audited"` | `"bundled"` |
| `sepolia-alpha` overlay | `"audited"` | `"bundled"` |
| `sepolia-integration`, `testing` | `"all"` | unchanged |

`config_schema.json` is regenerated, so its default is now `"bundled"`.

The `Audited` variant stays selectable — nothing picks it now, but it still names Cairo's built-in list and costs one match arm. Say the word if you'd rather drop it.

Remaining `audited` references are all deliberate: the enum variant and its flag mapping, and `allowed_libfuncs_aligned_to_audited`, which is the test that compares our list *against* Cairo's audited one.

## What this changes operationally

Under `"bundled"` the node resolves `apollo_compile_to_casm/resources/allowed_libfuncs.json` from disk at startup and panics if it is absent — so the image must ship it. Both images that can construct the compiler already do, via #15133: `Dockerfile` (`apollo_node`) and `node_setup.Dockerfile` (`sequencer_node_setup`). `simulator` and the `dummy_*` images do not run the node.

As it stands the bundled list holds exactly the same entries as Cairo's audited list (258 each, both set differences empty), so switching these two overlays is a **no-op today** — it changes which flag is passed, not which classes are accepted. Sibling PR #15136 narrows the bundled list, so the effective change depends on merge order.

## Testing

- New `compile_with_the_default_config` — compiles with `SierraCompilationConfig::default()`, which is the path that now has to resolve the bundled list from disk. Without it, nothing in the suite exercised the new default.
- `cargo test -p apollo_compile_to_casm -p apollo_sierra_compilation_config -p apollo_node_config -p apollo_deployments` — 39 passed, including `deployment_files_are_up_to_date` and `all_service_configs_load` (which resolves the replacer placeholder to the new schema default, so `"bundled"` is loaded through the real config path).
- fmt + `clippy --all-targets -- -D warnings` clean.


avi-starkware (commented):
> @avi-starkware+AGNT made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15137)* status: 0 of 5 files reviewed, 1 unresolved discussion (waiting on avi-starkware, matanl-starkware, and Yoni-Starkware).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/15137#-P1dUvRE7KwJ1RlJH9Ln:-P1dUvRE7KwJ1RlJH9Lo:b-x9ali6) (no related file):*
Reading #15136 and this one together: each describes itself as having no production impact, each is right on its own, and the combination does change what mainnet accepts.

#15136 excludes `coupon_buy`, `coupon_call` and `coupon_refund` from the bundled list, and says "no production impact today: no environment selects `bundled`". True at its own head. This PR makes it false by switching mainnet and sepolia-alpha to `"bundled"`.

You do flag that the effective change depends on merge order. I would go further and state the combined outcome plainly, because it looks like the point of the stack rather than a side effect: once both land, mainnet stops accepting coupon libfuncs in newly declared classes.

Two things I would rather have confirmed than infer:

- Only *new* declarations are affected, since already-declared classes are already compiled. Is there any path that recompiles an existing class under the current list, in state sync or a cache rebuild, where a class that was accepted before would now be rejected?
- Is dropping coupon support on mainnet the intended product change, and does it need to be timed with anything outside this repo?

Not blocking, and the code itself reads fine. This is about the operational effect being written down where whoever runs the deploy will see it.

avi-starkware (commented):
> @avi-starkware+AGNT made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15137)* status: 0 of 5 files reviewed, 3 unresolved discussions (waiting on matanl-starkware and Yoni-Starkware).

___
*[a discussion](https://reviewable.io/reviews/starkware-libs/sequencer/15137#-P1dhLZU9MudjvQ_lYyN:-P1dhLZV1YyzfJX_pNb5:bm0aaip) (no related file):*
Correcting the Bugbot summary above, since it contradicts the PR description and could send someone to the wrong risk assessment.

Bugbot says "the bundled list is slightly wider than audited today". It is not wider. At the pinned `cairo-lang-starknet-classes` 2.19.4 the two lists hold the same 258 entries, with both set differences empty and no value mismatches:

```
$ python3 -c "
import json
D='.../cairo-lang-starknet-classes-2.19.4/src/allowed_libfuncs_lists/'
b=json.load(open('crates/apollo_compile_to_casm/resources/allowed_libfuncs.json'))['allowed_libfuncs']
a=json.load(open(D+'audited.json'))['allowed_libfuncs']
print(len(b), len(a), sorted(set(b)-set(a)), sorted(set(a)-set(b)),
      [k for k in set(b)&set(a) if b[k]!=a[k]])"
258 258 [] [] []
```

The three `sha512_*` libfuncs that `PENDING_LIBFUNCS` still names as pending were audited upstream some time ago, which is where the "slightly wider" idea comes from.

That check compares against the Rust library, so I also confirmed it against the binary that actually enforces the list. The installed `starknet-sierra-compile` 2.19.4 embeds `audited.json` byte for byte:

```
$ python3 -c "
bin=open('~/.cargo/tools/starknet-sierra-compile-2.19.4/bin/starknet-sierra-compile','rb').read()
aud=open('.../allowed_libfuncs_lists/audited.json','rb').read()
print(len(aud), aud in bin)"
8886 True
```

So the description is right and the bot is wrong: switching mainnet and sepolia-alpha to `"bundled"` accepts exactly the same classes it accepted before. What changes is where the list comes from, a file in this repo rather than a list compiled into the binary, and that is what makes the drift guard load-bearing from here on.

Purely informational, no action needed on this thread.
___
*[`crates/apollo_compile_to_casm/src/compile_test.rs` line 40 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15137#-P1dVHBV-5JSG_c-7qp_:-P1dVHBV-5JSG_c-7qpa:b-74ny8r) ([raw file](https://github.com/starkware-libs/sequencer/blob/79adf8241b3334158b473008954e55833ff10635/crates/apollo_compile_to_casm/src/compile_test.rs#L40)):*
> ```XML
> // Libfuncs in allowed_libfuncs.json but not yet in Cairo's audited list.
> // Remove entries once they're added to the audited list.
> const PENDING_LIBFUNCS: &[&str] =
> ```

`PENDING_LIBFUNCS` excuses nothing today, and this PR is what makes that cost something.

All three `sha512_*` names are already in Cairo's audited list at the pinned 2.19.4, so the bundled list and the audited list are the same 258 entries:

```
$ python3 -c "
import json
D='.../cairo-lang-starknet-classes-2.19.4/src/allowed_libfuncs_lists/'
b=json.load(open('crates/apollo_compile_to_casm/resources/allowed_libfuncs.json'))['allowed_libfuncs']
a=json.load(open(D+'audited.json'))['allowed_libfuncs']
print(len(b), len(a), sorted(set(b)-set(a)), sorted(set(a)-set(b)))"
258 258 [] []
```

The exemption still has an effect, though, and it is the wrong one. Any name listed here is filtered out of the `extra` check in `allowed_libfuncs_aligned_to_audited`, and after this PR that test is the only thing standing between a cairo-lang bump and what mainnet accepts. If a later version de-audits one of the `sha512_*` libfuncs, the bundled list keeps it, the test stays green, and mainnet goes on accepting a libfunc upstream just withdrew.

I ran both halves in a scratch worktree, adding a fake `zz_deaudited_probe` entry to the bundled JSON:

```
# listed in PENDING_LIBFUNCS
test compile_test::allowed_libfuncs_aligned_to_audited ... ok

# not listed (control)
Extra (in json but not in audited): ["zz_deaudited_probe"]
test result: FAILED. 0 passed; 1 failed
```

Emptying `PENDING_LIBFUNCS` entirely also leaves the test green, which confirms nothing currently relies on it.

So either empty it now, which is what its own comment asks for ("Remove entries once they're added to the audited list"), or keep the mechanism and make it self-expiring:

```rust
let stale: Vec<_> =
    PENDING_LIBFUNCS.iter().filter(|name| expected.contains_key(**name)).collect();
assert!(stale.is_empty(), "PENDING_LIBFUNCS entries are already audited: {stale:?}");
```

That fires immediately today and empties the list for you, and the next exemption cannot quietly outlive its reason.

The constant itself comes from #15133. I am raising it here because this is the PR that puts mainnet behind it. Not blocking.
___
*[`crates/apollo_compile_to_casm/src/compile_test.rs` line 178 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15137#-P1dVKe39uN28uTQjS90:-P1dVKe39uN28uTQjS91:bipvnok) ([raw file](https://github.com/starkware-libs/sequencer/blob/79adf8241b3334158b473008954e55833ff10635/crates/apollo_compile_to_casm/src/compile_test.rs#L178)):*
> ```XML
> /// without it would panic on startup rather than fail a compilation.
> #[test]
> fn compile_with_the_default_config() {
> ```

`compile_with_the_default_config` passes whichever list the default selects, so it does not guard the thing its comment describes.

`SierraCompilationConfig::default()` is field-for-field the config `compile_against_the_bundled_libfuncs_list` already compiles with, just above at line 165. `SIERRA_COMPILATION_CONFIG` sets the same three `DEFAULT_MAX_*` consts that `Default` uses, and the one field that used to differ, `allowed_libfuncs_list`, is `Bundled` on both sides once this PR lands. The new test also asserts less than its neighbour: a bare `.unwrap()` against that one's `assert_eq!`.

I flipped the default in a scratch worktree and ran it:

```
DEFAULT_ALLOWED_LIBFUNCS_LIST = Audited  ->  test compile_test::compile_with_the_default_config ... ok
DEFAULT_ALLOWED_LIBFUNCS_LIST = All      ->  test compile_test::compile_with_the_default_config ... ok
```

Both pass. I added a liveness control to rule out a stale build artifact, setting the default to `Audited` and asserting it was `Bundled`, and that failed as expected, so the mutation really was in the binary under test. All three selections compile this fixture, so the assertion cannot tell them apart.

The flip is not untested, to be clear. `default_config_file_is_up_to_date` already pins `DEFAULT_ALLOWED_LIBFUNCS_LIST` against the committed `config_schema.json`. And since #15133 added the `libfunc_list_arg` assertion on line 171, the neighbour now covers the argv shape too.

So I would drop this test. If you want the default pinned in code as well, one line in `config_test.rs` costs less than a second compiler spawn:

```rust
assert_eq!(SierraCompilationConfig::default().allowed_libfuncs_list, AllowedLibfuncsList::Bundled);
```

If you keep it, one note on the doc comment. Its second sentence, about a node panicking at startup rather than failing a compilation, is already stated twice in `compiler.rs`: on the `libfunc_list_arg` field at line 21 and on the function at line 66. That fact reads better where it is enforced, and the test comment can just say why the test exists.

Not blocking either way.

avi-starkware (approved):
> @avi-starkware reviewed 5 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15137)* status: all files reviewed, 3 unresolved discussions (waiting on matanl-starkware and Yoni-Starkware).

avi-starkware:
*[`crates/apollo_compile_to_casm/src/compile_test.rs` line 178 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15137#-P1dVKe39uN28uTQjS90:-P1dkIxb06QocjNl55RV:b-4nmsuk) ([raw file](https://github.com/starkware-libs/sequencer/blob/79adf8241b3334158b473008954e55833ff10635/crates/apollo_compile_to_casm/src/compile_test.rs#L178)):*
<details><summary><i>Previously, avi-starkware+AGNT (Agent AGNT for Avi Cohen) wrote…</i></summary><blockquote>

`compile_with_the_default_config` passes whichever list the default selects, so it does not guard the thing its comment describes.

`SierraCompilationConfig::default()` is field-for-field the config `compile_against_the_bundled_libfuncs_list` already compiles with, just above at line 165. `SIERRA_COMPILATION_CONFIG` sets the same three `DEFAULT_MAX_*` consts that `Default` uses, and the one field that used to differ, `allowed_libfuncs_list`, is `Bundled` on both sides once this PR lands. The new test also asserts less than its neighbour: a bare `.unwrap()` against that one's `assert_eq!`.

I flipped the default in a scratch worktree and ran it:

```
DEFAULT_ALLOWED_LIBFUNCS_LIST = Audited  ->  test compile_test::compile_with_the_default_config ... ok
DEFAULT_ALLOWED_LIBFUNCS_LIST = All      ->  test compile_test::compile_with_the_default_config ... ok
```

Both pass. I added a liveness control to rule out a stale build artifact, setting the default to `Audited` and asserting it was `Bundled`, and that failed as expected, so the mutation really was in the binary under test. All three selections compile this fixture, so the assertion cannot tell them apart.

The flip is not untested, to be clear. `default_config_file_is_up_to_date` already pins `DEFAULT_ALLOWED_LIBFUNCS_LIST` against the committed `config_schema.json`. And since #15133 added the `libfunc_list_arg` assertion on line 171, the neighbour now covers the argv shape too.

So I would drop this test. If you want the default pinned in code as well, one line in `config_test.rs` costs less than a second compiler spawn:

```rust
assert_eq!(SierraCompilationConfig::default().allowed_libfuncs_list, AllowedLibfuncsList::Bundled);
```

If you keep it, one note on the doc comment. Its second sentence, about a node panicking at startup rather than failing a compilation, is already stated twice in `compiler.rs`: on the `libfunc_list_arg` field at line 21 and on the function at line 66. That fact reads better where it is enforced, and the test comment can just say why the test exists.

Not blocking either way.
</blockquote></details>

Might be a good idea to indeed add
```rust
assert_eq!(SierraCompilationConfig::default().allowed_libfuncs_list, AllowedLibfuncsList::Bundled);
```
to make sure a node that ships without the bundled list fails at startup when running the default config.

Yoni-Starkware (commented):
> @Yoni-Starkware resolved 3 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15137)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on matanl-starkware).

avi-starkware (approved):
:lgtm:

> @avi-starkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15137#-:-P1dke8H3xdzHU6lIwfp:bnfp4nl)* status: all files reviewed, 3 unresolved discussions (waiting on avi-starkware+AGNT, matanl-starkware, and Yoni-Starkware).

gkaempfer:

---

PR #15141, opened 2026-09-16
apollo_compile_to_casm: test that the bundled libfunc list rejects what it excludes

Yoni-Starkware:
Follow-ups deferred from the #15133 review.

## The test

Nothing so far distinguished `Bundled` from the built-in lists: the list selection gates acceptance but does not change the emitted CASM, so comparing outputs can't tell them apart. A class using an excluded libfunc can.

`resources/coupon_contract.sierra.json` is such a class, with its Cairo source committed beside it. Run against the three selections directly:

| List | Result |
|---|---|
| `all` | compiles |
| `audited` | compiles |
| `bundled` | **rejected** — `Libfunc coupon_buy is not allowed` |

Both built-in lists permit coupons, so rejecting this class is the only behaviour that separates our list from either of them. `bundled_libfuncs_list_rejects_an_excluded_libfunc` asserts all three.

This only became possible once #15136 landed — on #15133 alone the bundled list held exactly the same entries as `audited`.

## Other review follow-ups

- **`PENDING_LIBFUNCS` had no staleness guard.** It is only used to *permit* extras, so an entry that has since been audited is neither missing nor extra and sits there indefinitely. Added the mirror of the `EXCLUDED_LIBFUNCS` check; it flagged all three `sha512_*` entries, which are now in the audited list at the pinned 2.19.4, so the const is emptied.
- **The libfunc argument assertions moved to `compiler_test.rs`.** They sat inside a test that constructs `SierraToCasmCompiler`, so they only ran where the pinned `starknet-sierra-compile` is installed, though the function needs nothing but a resolvable path. They now also pin the bundled *path*, not just the flag — pointing `Bundled` at the wrong file previously still passed.
- **`test_max_memory_usage` positive flow** overrode `max_memory_usage` with the value `SIERRA_COMPILATION_CONFIG` already sets, so it rebuilt the same compiler and compared a value with itself. Now uses a limit that differs from the default.

## Testing

`cargo test -p apollo_compile_to_casm` — 13 passed. `apollo_sierra_compilation_config`, `apollo_node_config`, `apollo_deployments` — 30 passed. fmt + `clippy --all-targets -- -D warnings` clean.


avi-starkware (commented):
> @avi-starkware+AGNT made 6 comments.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15141)* status: 0 of 5 files reviewed, 4 unresolved discussions (waiting on avi-starkware and Yoni-Starkware).

___
*[`crates/apollo_compile_to_casm/src/compiler_test.rs` line 6 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15141#-P1epR9G7v9gOMMBHaXt:-P1epR9G7v9gOMMBHaXu:bhz4cg4) ([raw file](https://github.com/starkware-libs/sequencer/blob/e74a7c146bc153fd1eb9dde0a26a5ce5464df0c7/crates/apollo_compile_to_casm/src/compiler_test.rs#L6)):*
> ```XML
> use crate::constants::BUNDLED_ALLOWED_LIBFUNCS_PATH;
> 
> #[test]
> ```

`built_in_libfunc_lists_are_selected_by_name` asserts the same string literals that `libfunc_list_arg` is written with, so it pins the code against itself rather than against cairo-lang.

The match arms in `compiler.rs` spell `"audited"` and `"all"` on lines 75-76, and the test spells them again. If cairo-lang renamed a built-in list, the compiler invocation would break and this test would stay green.

`cairo_lang_starknet_classes::allowed_libfuncs::{BUILTIN_AUDITED_LIBFUNCS_LIST, BUILTIN_ALL_LIBFUNCS_LIST}` are both `pub const` in the pinned 2.19.4 and would pin the actual contract. Neither is imported anywhere in the repo today, so this needs a new `use` here.

Dropping the test is a fair answer too, given what it covers as written.

Nonblocking.
___
*[`crates/apollo_compile_to_casm/src/compile_test.rs` line 38 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15141#-P1epSVMDjUCqDDiSbx1:-P1epSVMDjUCqDDiSbx2:bj3l2ee) ([raw file](https://github.com/starkware-libs/sequencer/blob/e74a7c146bc153fd1eb9dde0a26a5ce5464df0c7/crates/apollo_compile_to_casm/src/compile_test.rs#L38)):*
> ```XML
> };
> 
> // Libfuncs in allowed_libfuncs.json but not yet in Cairo's audited list.
> ```

`PENDING_LIBFUNCS` is empty now, so the `pending_set` filter and the new staleness assert are both no-ops. Keep the empty hook, or delete the three pieces and re-add them the next time `allowed_libfuncs.json` runs ahead of the audited list?

Asking which you intended, not reporting a bug. The guard works. Populating it fires the assert as designed:

```
PENDING_LIBFUNCS = &["array_append"]
-> panics at compile_test.rs:188
```

And there is a real argument for keeping it: it is the only assert that catches a pending entry that has since been audited. Such an entry sits in both maps, so the earlier missing/extra check stops seeing it, and this one takes over. If that is the intent, saying so in the comment above the constant would make the empty slice read as deliberate rather than leftover.

Nonblocking.
___
*[`crates/apollo_compile_to_casm/src/compile_test.rs` line 49 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15141#-P1epMwADIEJixylE9Ez:-P1epMwADIEJixylE9F-:bu0sprg) ([raw file](https://github.com/starkware-libs/sequencer/blob/e74a7c146bc153fd1eb9dde0a26a5ce5464df0c7/crates/apollo_compile_to_casm/src/compile_test.rs#L49)):*
> ```XML
> const EXCLUDED_LIBFUNCS: &[&str] = &["coupon_buy", "coupon_call", "coupon_refund"];
> 
> // A class using an excluded libfunc. Regenerate from the sibling .cairo with:
> ```

The regeneration recipe on `EXCLUDED_LIBFUNC_CLASS_PATH` does not regenerate `coupon_contract.sierra.json`.

`coupon_buy` sits behind an experimental Cairo feature that has to be enabled in a crate config, so the fixture cannot be produced from a single `.cairo` file at all. I ran the documented command four ways against the cairo-package `starknet-compile`:

```
# as written, pointed at the sibling .cairo
Error: The given path is a file, but --single-file was not supplied.

# adding -s
error[E2167]: Coupons are disabled in the current crate. You can enable them by
              enabling the coupons experimental feature in the crate config.
error[E2030]: Wrong number of arguments. Expected 1, found: 2

# pointed at resources/ as a crate
Failed to load project config: No such file or directory

# a real crate dir whose cairo_project.toml carries
#   [config.global] edition = "2024_07"
#   [config.global.experimental_features] coupons = true
exit 0
```

Only the last one produces the fixture. Worth noting `-s` alone looks like the fix and is not.

`scripts/install_compiler_binaries.sh` also never installs `starknet-compile`. It installs `starknet-sierra-compile` and `starknet-native-compile`, so whoever picks this up next has to find the binary under `target/bin/cairo_package__<CAIRO1_COMPILER_VERSION>/cairo/bin/`. Worth naming in the comment.

Committing the crate layout the fixture was built from is the durable fix. Spelling the full recipe out in the comment works too.

I'd fix it rather than leave it, because nothing in CI compiles the `.cairo`, so drift between it and the committed `.sierra.json` stays invisible.

On the good news side, the committed fixture is genuine: regenerating it that last way reproduced the file byte for byte except `sierra_program` felts 2 and 5, which carry the sierra and compiler versions.

Caveat on all of the above: measured with `starknet-compile` 2.19.0-rc.3. No 2.19.4 build exists on my machine, though the repo pins 2.19.4.

Nonblocking.
___
*[`crates/apollo_compile_to_casm/src/compile_test.rs` line 213 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15141#-P1epKHz2uGl_ZsQ5gNb:-P1epKHz2uGl_ZsQ5gNc:b-33127t) ([raw file](https://github.com/starkware-libs/sequencer/blob/e74a7c146bc153fd1eb9dde0a26a5ce5464df0c7/crates/apollo_compile_to_casm/src/compile_test.rs#L213)):*
> ```XML
> }
> 
> /// The default selects the bundled list, which has to be resolved from disk; a node that ships
> ```

The doc comment on `bundled_libfuncs_list_rejects_an_excluded_libfunc` opens with two lines written for a different test.

Lines 213-214 documented `compile_with_the_default_config` before this change. The new test landed between them and that function, and Rust attaches a contiguous `///` block downward, so all four lines now document the new test:

```rust
/// The default selects the bundled list, which has to be resolved from disk; a node that ships
/// without it would panic on startup rather than fail a compilation.
/// The built-in lists both permit [`EXCLUDED_LIBFUNCS`], so rejecting this class is the only
/// behaviour that tells the bundled list apart from `Audited` and `All`.
#[test]
fn bundled_libfuncs_list_rejects_an_excluded_libfunc() {
```

The new test never builds `SierraCompilationConfig::default()`, so the first sentence misdescribes it. And `compile_with_the_default_config` on line 243 is left with no doc at all.

Moving lines 213-214 back down onto `compile_with_the_default_config` fixes it.

One thing that does not fix it, in case it looks like it would: putting a blank line between the two halves. rustc reads a blank line inside a doc block as a paragraph break and still attaches both halves to the item below. I checked that before writing this.

Nonblocking.
___
*[`crates/apollo_compile_to_casm/src/compile_test.rs` line 226 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15141#-P1epORYE4N_0nmUKpH7:-P1epORYE4N_0nmUKpH8:b-ih2gpr) ([raw file](https://github.com/starkware-libs/sequencer/blob/e74a7c146bc153fd1eb9dde0a26a5ce5464df0c7/crates/apollo_compile_to_casm/src/compile_test.rs#L226)):*
> ```XML
>         ..SIERRA_COMPILATION_CONFIG
>     });
>     let bundled_list_compiler = SierraToCasmCompiler::new(SierraCompilationConfig {
> ```

The bundled-compiler construction is written three times in this file now. Lines 226-229 are byte-identical to lines 204-207 in `compile_against_the_bundled_libfuncs_list`, and 222-225 differ only in the variant:

```
md5(204-207) = 78ed5c9284336cc83a30167fc24499c3
md5(226-229) = 78ed5c9284336cc83a30167fc24499c3
```

A helper next to `compiler()` collapses each site to one line:

```rust
fn compiler_with_libfuncs_list(allowed_libfuncs_list: AllowedLibfuncsList) -> SierraToCasmCompiler {
    SierraToCasmCompiler::new(SierraCompilationConfig {
        allowed_libfuncs_list,
        ..SIERRA_COMPILATION_CONFIG
    })
}
```

`compiler()` can then delegate to it with `AllowedLibfuncsList::All`, since `SIERRA_COMPILATION_CONFIG` already sets that on line 35.

Nonblocking, and easy to drop if you'd rather keep each construction spelled out.
___
*[`crates/apollo_compile_to_casm/src/compile_test.rs` line 231 at r1](https://reviewable.io/reviews/starkware-libs/sequencer/15141#-P1epPrX1g3VxB_3B29F:-P1epPrX1g3VxB_3B29G:b-xka1xr) ([raw file](https://github.com/starkware-libs/sequencer/blob/e74a7c146bc153fd1eb9dde0a26a5ce5464df0c7/crates/apollo_compile_to_casm/src/compile_test.rs#L231)):*
> ```XML
>     });
> 
>     compiler().compile(excluded_libfunc_class.clone()).expect("`All` must accept the class.");
> ```

The `All` compile on this line cannot fail while the `Audited` compile just below it passes.

`audited` is a subset of `all` in cairo-lang-starknet-classes 2.19.4, which is what `Cargo.lock` resolves at this commit, and upstream has a test enforcing the relation, so it holds by invariant rather than by luck:

```
audited_not_in_all == []
```

So the line spawns a compiler subprocess that carries no signal the `Audited` line does not already carry.

Before you act on that, there is a reason to keep it. The doc comment above the test claims "The built-in lists both permit [`EXCLUDED_LIBFUNCS`]", and this line is what demonstrates the "both" half. If you want that claim in the doc, keeping the line that backs it is a fair trade. The leaner version drops the line and states the subset relation in the comment instead.

Your call, and I'm happy either way. Nonblocking.

Yoni-Starkware (commented):
> @Yoni-Starkware reviewed 1 file and resolved 4 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15141)* status: 0 of 6 files reviewed, all discussions resolved (waiting on avi-starkware+AGNT).

avi-starkware:
*[`crates/apollo_compile_to_casm/src/compile_test.rs` line 41 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/15141#-P283tXa5qAxnY30A8mX:-P283tXbCPLPvY1hecYE:b45h) ([raw file](https://github.com/starkware-libs/sequencer/blob/d46a670c6f0c54a5295332c456e7977d42ad6629/crates/apollo_compile_to_casm/src/compile_test.rs#L41)):*
> ```XML
> // state: entries go in when the json runs ahead, and come out when the audited list catches up.
> // The staleness assert in `allowed_libfuncs_aligned_to_audited` is what catches the latter, since
> // a since-audited entry sits in both maps and the missing/extra checks stop seeing it.
> ```





_Suggestion:_
```XML
// Libfuncs in allowed_libfuncs.json but not yet in Cairo's audited list. This list should normally
// be empty. `allowed_libfuncs_aligned_to_audited` fails if an audited entry remains here.
```

avi-starkware:
*[`crates/apollo_compile_to_casm/src/compile_test.rs` line 59 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/15141#-P284Q8CA89zhuXIMHsF:-P284Q8CA89zhuXIMHsG:b45h) ([raw file](https://github.com/starkware-libs/sequencer/blob/d46a670c6f0c54a5295332c456e7977d42ad6629/crates/apollo_compile_to_casm/src/compile_test.rs#L59)):*
> ```XML
> //   starknet-compile crates/apollo_compile_to_casm/resources/coupon_contract \
> //       --allowed-libfuncs-list-name all \
> //       crates/apollo_compile_to_casm/resources/coupon_contract.sierra.json
> ```





_Suggestion:_
```XML
// A class using an excluded libfunc. `coupon_buy` requires the crate's `coupons` feature.
// Regenerate from the repo root:
//
// target/bin/cairo_package__<CAIRO1_COMPILER_VERSION>/cairo/bin/starknet-compile \
//   crates/apollo_compile_to_casm/resources/coupon_contract \
//   --allowed-libfuncs-list-name all \
//   crates/apollo_compile_to_casm/resources/coupon_contract.sierra.json
```

avi-starkware (approved):
> @avi-starkware partially reviewed 6 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15141)* status: all files reviewed, 2 unresolved discussions (waiting on Yoni-Starkware).

avi-starkware (approved):
:lgtm:

> @avi-starkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15141#-:-P287XspDTK1ODs8e3ou:bnfp4nl)* status: all files reviewed, 2 unresolved discussions (waiting on Yoni-Starkware).

avi-starkware:
*[`crates/apollo_compile_to_casm/src/compile_test.rs` line 59 at r2](https://reviewable.io/reviews/starkware-libs/sequencer/15141#-P284Q8CA89zhuXIMHsF:-P287amABuQxnI7d78ze:b-j2ysgd) ([raw file](https://github.com/starkware-libs/sequencer/blob/d46a670c6f0c54a5295332c456e7977d42ad6629/crates/apollo_compile_to_casm/src/compile_test.rs#L59)):*
> ```XML
> //   starknet-compile crates/apollo_compile_to_casm/resources/coupon_contract \
> //       --allowed-libfuncs-list-name all \
> //       crates/apollo_compile_to_casm/resources/coupon_contract.sierra.json
> ```

Sorry can keep `starknet-compile` command without the full path

Yoni-Starkware (commented):
> @Yoni-Starkware reviewed 1 file and resolved 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15141)* status: 5 of 6 files reviewed, all discussions resolved (waiting on avi-starkware).

avi-starkware (approved):
> @avi-starkware reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15141)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Yoni-Starkware).

avi-starkware (approved):
> @avi-starkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15141)* status: all files reviewed, 1 unresolved discussion (waiting on Yoni-Starkware).

___
*[`crates/apollo_compile_to_casm/src/compile_test.rs` line 53 at r3](https://reviewable.io/reviews/starkware-libs/sequencer/15141#-P2D0-tg1PGhS3lbmJ3D:-P2D0-tg1PGhS3lbmJ3E:b-z57egv) ([raw file](https://github.com/starkware-libs/sequencer/blob/97d72f9aa98c5121f7690dbdb79f2524301322f4/crates/apollo_compile_to_casm/src/compile_test.rs#L53)):*
> ```XML
> //
> // starknet-compile \
> //   crates/apollo_compile_to_casm/resources/coupon_contract \
> ```

No need for this line-break



_Suggestion:_
```XML
// starknet-compile crates/apollo_compile_to_casm/resources/coupon_contract \
```

avi-starkware (approved):
:lgtm:

> @avi-starkware made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15141#-:-P2D09AyED0WoeWSSK8x:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on Yoni-Starkware).

Yoni-Starkware (commented):
> @Yoni-Starkware resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/starkware-libs/sequencer/15141)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on Yoni-Starkware).

